PR #31, opened 2015-01-05
Fix 'pip' install problem

lungothrin:
install 'llvmlite' with pip will cause this error
  error: option --single-version-externally-managed not recognized

this patch fix this problem


seibert:
Thanks!

KennethNielsen:
I was wondering when the next release of llvmlite is expected, so llvmlite can installed with pip?

Issue #35, opened 2015-01-25
"Extra" dependency


danilobellini:
Calling `python setup.py install` on the v0.2.1 tag version (9f60cd1), I've got the message (and some "cascading" error messages afterwards):

```
/usr/bin/ld: cannot find -ledit
```

That appears because the `llvm-config --system-libs --libs all` included that `-ledit`. That might be an Ubuntu 14.10 (utopic) LLVM  packaging issue, but I'm not sure whether `llvm-3.5-dev` should have `libedit` as dependency. Anyway, including in the README that [some] people should install `libedit-dev` would help at least as a workaround.

eliben:
I encountered this problem as well and mentioned it in the short blog post I wrote about installing and using llvmlite [http://eli.thegreenplace.net/2015/building-and-using-llvmlite-a-basic-example/]

LLVM gained the dependency on libedit just last year [with https://github.com/llvm-mirror/llvm/commit/cb6684b63b3c4c5a90e194c5719bc82690180f30] for the sake of the clang-query tool. It was placed in the core of LLVM because it can have additional users like LLDB.

pitrou:
A mention was added in the README in PR #37.

PR #37, opened 2015-01-26
Mention libedit-dev in the README

eliben:
Issue #35


pitrou:
Thanks. Do you have an idea if it's possible to entirely avoid the dependency on libedit?

eliben:
I wrote some details about this in #35 - this may be possible when building a custom LLVM, but probably not with the default one. The default LLVM binaries is what tools like clang-query and lldb expect to find and they need the LineEditor library.

I suppose this can be circumvented by moving the loading of libedit to runtime within LLVM. Perhaps a bug request can be opened for that in the LLVM tracker.

pitrou:
Ok, thank you.

Issue #39, opened 2015-01-28
Query ir.Module for function or global


eliben:
LLVM's Module class has `getFunction()` and `getGlobalVariable()` for name lookup.

I was looking for something similar in `ir.Module` in llvmlite. I see the `globals` attribute, but I'm not sure whether it's officially part of the public API of the class. Would it be worthwhile to expose accessor methods? Or can I assume that any non `_` prepended attribute in ir classes is public?

pitrou:
I think it's better to expose accessor methods. We might want to make those attribute privates :)

eliben:
I'll see if I can come up with a simple patch

PR #42, opened 2015-01-29
Add .get_global() accessor to ir.Module

eliben:
The accessor allows querying the globals mapping without directly accessing the
class's attributes. Discussed in issue #39 

Note: the C++ API has getFunction and getGlobalVariable, which do the type checking in addition to querying the symbol table. It's possible to redo the API that way, if this is deemed preferable.

### DIFF
--- a/llvmlite/tests/test_ir.py
+++ b/llvmlite/tests/test_ir.py
@@ -111,6 +111,15 @@ class TestFunction(TestBase):
 
 class TestIR(TestBase):
 
+    def test_globals_access(self):
+        mod = self.module()
+        foo = ir.Function(mod, ir.FunctionType(ir.VoidType(), []), 'foo')
+        ir.Function(mod, ir.FunctionType(ir.VoidType(), []), 'bar')
+        globdouble = ir.GlobalVariable(mod, ir.DoubleType(), 'globdouble')
+        self.assertEqual(mod.get_global('foo'), foo)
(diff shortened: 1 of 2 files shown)


eliben:
Friendly ping :)

pitrou:
Sorry, I hadn't noticed this PR.

llvmlite/tests/test_ir.py
```diff
@@ -111,6 +111,15 @@ def test_define(self):
 
 class TestIR(TestBase):
 
+    def test_globals_access(self):
+        mod = self.module()
+        ir.Function(mod, ir.FunctionType(ir.VoidType(), []), 'foo')
+        ir.Function(mod, ir.FunctionType(ir.VoidType(), []), 'bar')
+        ir.GlobalVariable(mod, ir.DoubleType(), 'globdouble')
+        self.assertIsInstance(mod.get_global('foo'), ir.Function)
```
pitrou: Can't you just test that the object created above is returned? Or am I missing something?

eliben: Sorry, just to clarify what you're asking. You suggest to replace the assertIsInstance by identity comparison with the `Value`\s created with ir.Function and ir.GlobalVariable?

pitrou: Yes :)

eliben: Makes sense. Redid the pull request

PR #47, opened 2015-02-09
Add support for loading dynamic libraries

leonardt:


pitrou:
Thanks for the patch. Can you explain why this is necessary? (rather than e.g. using ctypes)

leonardt:
Some of our JITted code will make calls into run times such as OpenCL and OpenMP so we need the libraries to be linked into the engine

pitrou:
Ok. The patch should have proper error management, though. See other exported functions to see how this is done. Also, adding a minimal test would be good.

leonardt:
I just added error handling which will raise a RuntimError with the message passed from the llvm call (how others are implemented).

It tests a failed library load, I can also add a successful one but I will have to stub out some llvm to make sure that calling a library function works properly.

leonardt:
Okay I've added a test.  A couple notes.  It calls into OpenCL, so if an OpenCL runtime is not present the test will fail.  I've added a conditional that checks whether or not it is present before running the test (should save failures for devs who don't have it installed).  I added it to the .travis.yml (opencl-headers + fglrx) to get a runtime.  This unfortunately extends the install step of the travis build, not sure how you feel about this.  Alternative would be to write a simpler test that calls a library already present or one that is faster to install.  Let me know (p.s. is there a library that can be linked and called that should be readily available on most machines?)

leonardt:
As a note the sudo apt-get install step goes from 14s to 22s with the extra dependencies

leonardt:
Another note is that currently if it fails to find a symbol (library not present), LLVM throws an assertion that will kill Python.  I could't find any other examples (please point me to them if they exist) but perhaps we need to catch the exception somehow down in C and pass the string back to Python.

pitrou:
We shouldn't depend on OpenCL for the tests. Can't you use something simpler? (for example libm)
Also, I don't think we need to go as far as calling a function from the library.

leonardt:
I removed the dependency, one issue with libm is that it's linked by default on OS X (i.e. there are no issues running it when I don't link it), not sure about the same with linux I will test it right now on a box.  By the way I can squash the commits before this eventually merge to keep the history clean.

leonardt:
@pitrou Okay I squashed the commits.  I tested loading libm which shouldn't throw a runtime error, I removed calling a function because it really doesn't test anything since a function call to libm works even without linking it.  I think this is as well as we can test it unless we decide to include a non system library for the test.

llvmlite/tests/test_dylib.py
```diff
@@ -0,0 +1,35 @@
+from . import TestCase
+
+from llvmlite import binding as llvm
+from llvmlite.binding import dylib
+import platform
+from ctypes.util import find_library
+import unittest
+
+@unittest.skipUnless(platform.system() in ["Linux", "Darwin"], "Unsupport test for current OS")
```
pitrou: Actually, test_bad_library would probably work under Windows.

llvmlite/tests/test_dylib.py
```diff
@@ -0,0 +1,35 @@
...
+                            in str(context.exception))
+
+    def test_libm(self):
+        try:
+            if self.system  == "Linux":
+                libm = find_library("m")
+            elif self.system == "Darwin":
+                libm = find_library("libm")
+            dylib.load_library_permanently(libm)
+        except Exception:
```
pitrou: No need to catch the exception.

llvmlite/tests/test_dylib.py
```diff
@@ -0,0 +1,35 @@
+from . import TestCase
```
pitrou: This doesn't deserve to be in its separate test file. You can put in test_binding.py.

llvmlite/tests/test_dylib.py
leonardt: not sure what the error message should look like, okay to just assert that an exception is raised?

pitrou: Yes, that's ok.

pitrou:
Thank you!

PR #53, opened 2015-04-14
Use latin1 encoding instead of UTF8

sklam:
With certain constant i8 values, LLVM will generate LLVM IR that contain invalid characters for UTF8.  LLVM is probably using latin1.  Although, I can't find any documentation about character encoding for LLVM IR.

This PR contains a test that fails when using UTF8 encoding.  I have verified that this patch works with Numba on OSX.

### DIFF
--- a/llvmlite/tests/test_ir.py
+++ b/llvmlite/tests/test_ir.py
@@ -862,8 +862,13 @@ class TestConstant(TestBase):
         self.assertEqual(str(c), '{float, i1} undef')
 
     def test_encoding_problem(self):
+        try:
+            get_buffer = buffer
+        except:
+            get_buffer = lambda x: x   # Insert stub
+
         celems = [ir.Constant(ir.ArrayType(ir.IntType(8), 8),
-                              bytearray(array("d", [i + 0.5])))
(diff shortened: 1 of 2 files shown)


llvmlite/tests/test_ir.py
```diff
@@ -860,6 +861,22 @@ def test_structs(self):
         c = ir.Constant(ir.LiteralStructType((flt, int1)), ir.Undefined)
         self.assertEqual(str(c), '{float, i1} undef')
 
+    def test_encoding_problem(self):
+        celems = [ir.Constant(ir.ArrayType(ir.IntType(8), 8),
```
pitrou: Does it have to be this complicated? Why not a single constant with e.g. `bytearray(b"\x00\xff")` ?

pitrou: Or perhaps `bytearray(range(0, 256))` to exercise all possible bytes...

sklam: I was duplicating the failing encountered in numba.

I will try your suggestion

gmarkall:
This passes all llvmlite tests with Python 3 on Windows 7 64 bit, and Numba master passes all tests using this branch of llvmlite.

PR #56, opened 2015-04-28
Try to use Travis' container infrastructure

pitrou:
I'm just curious to see if if it makes builds genuinely faster. For now it seems to make them slower :-)


pitrou:
Since the non-container infrastructure seems chronically saturated nowadays (see https://www.traviscistatus.com/), I'm merging this PR to reduce wait times.

PR #62, opened 2015-05-17
Add support for memset, memcpy & memmove in Module.declare_intrinsic.

isbadawi:
(I also removed the redundant else to save a level of indentation).


pitrou:
Thanks for the PR. Could you add some tests? (probably in test_ir.py)

isbadawi:
OK, I added a couple of tests.

I meant to earlier, but I had LLVM 3.6 set up instead of 3.5 and couldn't run the test suite. It might be nice if you could easily run tests for the pure python parts without needing LLVM, but it's not a big deal I guess.

pitrou:
Great, thank you!

PR #64, opened 2015-05-25
Add binding to view control flow graph

sklam:
This adds the following to the ``llvmlite.binding.ValueRef`` for viewing the CFG of a function:
* ``.view_cfg()``: Generate a DOT file of the function CFG with all instructions included
* ``.view_cfg_only()``: Generate a DOT file of the function CFG without any instruction

This also add a helper function exported to ``llvmlite.ir`` for using the above functions on unmaterialized ``llvmlite.ir.Function`` or materialized ``llvmlite.ir.ValueRef``:
* ``view_function_cfg()``
* ``view_function_cfg_only()``

Note: 
* these functions depend on ``dot`` to be available in the PATH.  
* they cannot be unit-tested because they open a modal window to view the generated CFG.


seibert:
I think this just needs to be added to the docs, and then we can merge.

seibert:
@sklam: I think this also just needs some quick documentation of the new functions, and this can be merged as well.

sklam:
All the functions should be documented already in:
* https://github.com/numba/llvmlite/pull/64/files#diff-9264949b9c05b7cb425240c651945072R8
* https://github.com/numba/llvmlite/pull/64/files#diff-4e5facc3f5a02ea6579493aeb67416f8R69

pitrou:
The PR needs to be merged again.
Also, the docs mention a DOT file is produced, but they do not say where or under which name.
Bonus points if you can add a simple test :-)

sklam:
You don't have control on where that DOT file will go.  LLVM just print out its path on stdout.  I will add a note about that behavior.  Then, merge master.

sklam:
I discovered an alternative function that will allow the DOT graph to be returned as a string.  I will be replacing this PR with a new version.

PR #67, opened 2015-05-27
[WIP] Debug Symbol

sklam:
* Add minimal support for generating debug info for functions
* Support GDB version >7
* Require LLVM build with debug symbol
     * This is due to how GDB find out about JIT-ed function.  It needs to recognize a function call by looking at debug info.
* Require MCJIT with jit_debug=True


sklam:
**Note**
Need to build LLVM with the flags in http://llvm.org/docs/Packaging.html#compile-flags

pitrou:
Can you give pointers to the docs you used? That would be useful :)

llvmlite/ir/debuginfo.py
```diff
@@ -0,0 +1,356 @@
...
+    DW_TAG_base_type = 36
+    DW_TAG_array_type = 1
+    DW_TAG_enumeration_type = 4
+    DW_TAG_structure_type = 19
+    DW_TAG_union_type = 23
+    DW_TAG_inheritance = 28
+
+
+class DI_Attribute(IntEnum):
+    DW_AT_sibling = 0x01,
```
pitrou: Are those deliberately tuples? Just asking.

sklam: oops, no. good catch

sklam:
I am using documentation from: http://llvm.org/docs/SourceLevelDebugging.html
But that is not enough information.  A lot is from trial and error and comparing against what clang is doing.

sklam:
I think this PR is mostly a demonstration that debug info does work on some platform.

Let's move to 3.6 without merging this and I will update the PR afterwards.  Does 3.6 has the new metadata where they look like namedtuples?  That makes debugging the debuginfo a lot easier.

pitrou:
Hmm, actually it seems the new metadata is only available on 3.7: http://llvm.org/docs/LangRef.html#specialized-metadata-nodes

whitequark:
This can be closed, now that #162 is fixed.

pitrou:
Right, closing.

PR #69, opened 2015-06-11
Update build.py

t3h0th3r:
Add FreeBSD as supported platform.

### DIFF
--- a/ffi/Makefile.freebsd
+++ b/ffi/Makefile.freebsd
@@ -1,0 +1,19 @@
+LLVM_CONFIG ?= llvm-config
+
+CXX = clang++ -std=c++11 -stdlib=libc++
+CXXFLAGS = `$(LLVM_CONFIG) --cxxflags`
+LDFLAGS = `$(LLVM_CONFIG) --ldflags`
+LIBS = `$(LLVM_CONFIG) --system-libs --libs all`
+SRC = assembly.cpp bitcode.cpp core.cpp initfini.cpp module.cpp value.cpp \
+	executionengine.cpp transforms.cpp passmanagers.cpp targets.cpp dylib.cpp \
+	linker.cpp
+OUTPUT = libllvmlite.so
+
+all: $(OUTPUT)
+
+$(OUTPUT): $(SRC)
+	$(CXX) -shared $(CXXFLAGS) $(SRC) -o $(OUTPUT) $(LDFLAGS) $(LIBS)
+
+
+clean:
+	rm -rf test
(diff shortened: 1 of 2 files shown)


pitrou:
Hi,
The makefile you posted in the other PR is the same as the Linux, except for the different CXX definition, or are there other differences that I missed?

t3h0th3r:
It's using Clang like the OSX Build but creates an .so like the linux build, so it's actually a new file that differs from both the linux and OSX version. neither of those would compile on FreeBSD.

pitrou:
Yes, but I'm wondering if it wouldn't be better to have a "Makefile.generic" for non-OS X platforms and parameterize it by passing "CXX" as an environment variable. Hence my question :)

t3h0th3r:
Since FreeBSD also uses clang, the file is actually closer to the OSX version than the linux one:
~/llvmlite-0.5.0]$ diff -u ffi/Makefile.osx ffi/Makefile.freebsd |wc -l
      18
~/llvmlite-0.5.0]$ diff -u ffi/Makefile.linux ffi/Makefile.freebsd |wc -l
      27

pitrou:
Can you posts the diff? So as to see if the FreeBSD and OS X ones can be merged...

t3h0th3r:
```
--- ffi/Makefile.osx    2015-06-05 20:50:20.000000000 +0200
+++ ffi/Makefile.freebsd        2015-06-11 11:35:20.937569277 +0200
@@ -7,10 +7,12 @@
 SRC = assembly.cpp bitcode.cpp core.cpp initfini.cpp module.cpp value.cpp \
          executionengine.cpp transforms.cpp passmanagers.cpp targets.cpp dylib.cpp \
          linker.cpp
-OUTPUT = libllvmlite.dylib
+OUTPUT = libllvmlite.so

-all: $(SRC)
-       MACOSX_DEPLOYMENT_TARGET=10.7 $(CXX) -dynamiclib $(CXXFLAGS) $(SRC) -o $(OUTPUT) $(LDFLAGS) $(LIBS)
+all: $(OUTPUT)
+
+$(OUTPUT): $(SRC)
+       $(CXX) -shared $(CXXFLAGS) $(SRC) -o $(OUTPUT) $(LDFLAGS) $(LIBS)


 clean:
```

t3h0th3r:
eeeh i didn't mean to close it -.-'

pitrou:
Thanks. In the end I guess a separate Makefile is better. Could you just coalesce both pull requests? (i.e. add the Makefile to this very PR)

t3h0th3r:
like so?

pitrou:
Yes, exactly. I'm merging the PR, thank you!

PR #71, opened 2015-06-12
ConstOpMixin.inttoptr's str(self.get_reference()) already includes type

squisher:
This is a relatively simple fix: without it, IR like this gets created:
    inttoptr (i64 i64 39281280 to %sometype*)

### DIFF
--- a/llvmlite/ir/values.py
+++ b/llvmlite/ir/values.py
@@ -61,9 +61,8 @@ class ConstOpMixin(object):
     def inttoptr(self, typ):
         assert isinstance(self.type, types.IntType)
         assert isinstance(typ, types.PointerType)
-        op = "inttoptr ({0} {1} to {2})".format(self.type,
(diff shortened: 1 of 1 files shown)


pitrou:
Thanks for the patch. Could you also add a test? It should probably be added to test_cast_ops() in llvmlite/tests/test_ir.py

pitrou:
Actually, test_cast_ops() may not be the right place, since this is not about the IRBuilder API. Perhaps in the TestConstant class?

abebeos:
Is this one still relevant?

gmarkall:
This is superseded by #735, which implemented/fixed various operations on consts and added a test checking that `inttoptr` casts of consts produces correct IR:

https://github.com/numba/llvmlite/blob/00320f6fdf35bfb080717c2977edf2c38e6341fe/llvmlite/tests/test_ir.py#L2367-L2374

PR #72, opened 2015-06-15
Update Makefile.freebsd

floatwork:
llvm-config doesn't exist, was a symlink i made. llvm-config36 is the correct one.


floatwork:
nevermind, the build.py also points at the generic one, so easiest to ad that symlink.

PR #73, opened 2015-06-18
Add get_process_triple() and get_host_cpu_features()

sklam:
~~Note: get_host_cpu_features() is not implemented by LLVM 3.6 for x86.~~ (edited)

Updated for LLVM 3.7


llvmlite/binding/targets.py
```diff
@@ -6,10 +6,28 @@
...
+        return str(out)
+
+def get_host_cpu_features():
+    """
+    Returns a string (e.g. "+sse,-avx) indicating the available CPU features.
+    May return an empty string if LLVM cannot detect the CPU features.
+    """
+    with ffi.OutputString() as out:
+        ffi.lib.LLVMPY_GetHostCPUFeatures(out)
+        return str(out)
```
pitrou: Why not recreate a dict of str to bool here?

pitrou:
Hi Siu, since we have documentation now, those new functions should be added to the docs :-) (probably in source/binding/targets)

llvmlite/binding/targets.py
sklam: I am lazy and the output can go directly into the feature attributes of `create_target_machine` https://github.com/numba/llvmlite/blob/93a41b1243b6f3fe5b751ba842ca01794f7be6da/llvmlite/binding/targets.py#L130
=)

Perhaps, wrapping the `features` string in an object for providing the dictionary lookup lazily would be nice.

pitrou: Ah, I had forgotten about create_target_machine... The thing is, if you want to change stuff in that string, it's not very pretty: you have to resort to string or regex matching.
You mean a string that would also act as a dict? That's ugly.

sklam: I did not think about modifying it. hmmm....

seibert:
Do we still need this PR?  If so, then I think we just need some documentation on the added functions.

pitrou:
I think the API should be changed to return something more palatable (a dict or list of features).

pitrou:
I also don't need know there is necessary, although inspecting the possible features for a platform can always be insightful.

sklam:
Sorry for the delay.  Now that we have moved to LLVM 3.7 which implemented get_host_cpu_features() for x86, I have updated the PR, have applied pitrou's suggestion (https://github.com/numba/llvmlite/pull/73#discussion_r32848776), added docs and a numba PR to use the new features.

pitrou:
By the way, can you explain what this changes concretely? I have a hard time thinking of a situation where the "default triple" and the "process triple" would differ.

docs/source/binding/targets.rst
```diff
@@ -35,6 +42,15 @@ Functions
    Get the name of the host's CPU as a string.  You can use the
    return value with :meth:`Target.create_target_machine()`.
 
+.. function:: get_host_cpu_features()
+
+   Returns a dictionary-like object indicating the CPU features for current
+   architecture and whether they are enabled for this CPU.  The key-value pairs
+   are the feature name as string and a boolean indicating whether the feature
+   is available.  The returned value is an instance of ``FeatureMap`` class,
+   which adds a new method ``.flatten()`` for returning a string stuiable for
```
stuartarchibald: `s/stuiable/suitable/`

llvmlite/binding/targets.py
```diff
@@ -6,10 +6,63 @@
...
+                        for k, v in iterator)
+
+
+def get_host_cpu_features():
+    """
+    Returns a dictionary-like object indicating the CPU features for current
+    architecture and whether they are enabled for this CPU.  The key-value pairs
+    are the feature name as string and a boolean indicating whether the feature
+    is available.  The returned value is an instance of ``FeatureMap`` class,
+    which adds a new method ``.flatten()`` for returning a string stuiable for
```
stuartarchibald: `s/stuiable/suitable/`

seibert:
I think this is ready to merge after the spelling fixes.

sklam:
(replying to @pitrou https://github.com/numba/llvmlite/pull/73#issuecomment-187942547)

On my OSX, i will see the "default triple" showing 'x86_64-apple-darwin14.5.0' where as "process triple" shows 'x86_64-apple-darwin15.0.0'.  I am curious to see the behavior on windows with cygwin and mingw32.  

According to <http://lists.llvm.org/pipermail/llvm-commits/Week-of-Mon-20120102/134343.html>, the "default triple" may not match the host when LLVM is configured to cross-compile.  This is mainly for those who compile their own llvm.

Also, using the "process triple" matches the behavior of ``EngineBuilder.selectTarget`` (<http://llvm.org/docs/doxygen/html/TargetSelect_8cpp_source.html#l00047>).

pitrou:
Oh, I see, thank you.

PR #79, opened 2015-07-20
Expand _VALID_CHARS to match all LLVM IR identifiers

whitequark:
See http://llvm.org/docs/LangRef.html#identifiers.


seibert:
Looks like the patched expression is missing a closing parenthesis.

whitequark:
Fixed.

seibert:
Thanks!

PR #81, opened 2015-07-20
Always change to C locale before printing a module

### LINKED ISSUE #80: LLVM prints invalid IR with certain locale
Due to a LLVM bug https://llvm.org/bugs/show_bug.cgi?id=12906, certain locale (e.g. 'de_DE') causes LLVM to comma for decimal point when printing the IR.  The result is an invalid IR.  In fact, it may even trigger an assertion during printing.

sklam:
... and reverting afterwards.

* Fix #80
* Without the fix, LLVM will fail with an assertion.

### DIFF
--- a/ffi/module.cpp
+++ b/ffi/module.cpp
@@ -1,4 +1,5 @@
 #include <string>
+#include <clocale>
 #include "llvm-c/Core.h"
 #include "llvm-c/Analysis.h"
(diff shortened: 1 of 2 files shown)


seibert:
Just to verify: this solves the bug on your dev system?  We only can't test it in Travis-CI?

PR #82, opened 2015-07-21
Numerous improvements

whitequark:
Please see the individual commits. Sending them as separate PRs would be too noisy, so I group them in one.


whitequark:
Ping?

whitequark:
Hm, I seem to have deleted the repository. Let me resend the PR from the new one.

PR #83, opened 2015-07-22
Replace deprecated .get_pointer_to_global() 

sklam:
... with .get_function_address() and .get_global_value_address()

### DIFF
--- a/llvmlite/binding/executionengine.py
+++ b/llvmlite/binding/executionengine.py
@@ -1,6 +1,8 @@
 from __future__ import print_function, absolute_import
 
-from ctypes import byref, POINTER, c_char_p, c_bool, c_uint, c_void_p, c_int
+from ctypes import (byref, POINTER, c_char_p, c_bool, c_uint, c_void_p, c_int,
+                    c_uint64)
+import warnings
 
 from . import ffi, targets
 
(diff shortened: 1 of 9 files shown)


seibert:
This looks good to me.  @pitrou?

llvmlite/binding/executionengine.py
```diff
@@ -53,15 +55,25 @@ def __init__(self, ptr, module):
...
-        # getGlobalValueAddress should be used instead.
+    def get_pointer_to_function(self, gv):
+        warnings.warn(".get_pointer_to_function() is deprecated.  Use "
+                      ".get_function_address() instead.", DeprecationWarning)
         ptr = ffi.lib.LLVMPY_GetPointerToGlobal(self, gv)
         if ptr is None:
             raise ValueError("Cannot find given global value %r" % (gv.name))
         return ptr
 
-    get_pointer_to_function = get_pointer_to_global
```
pitrou: It looks a bit weird to deprecate one function but remove the other.

sklam: The name ``get_pointer_to_global()``  implies it works on any global values but it does not.  It gives incorrect result for non-function global values.  With ``get_pointer_to_function()``, the name implies that it is expecting a function value and it works correctly.  I intentionally break the API to help user avoid getting a segfault.

sklam: Also note that ``get_pointer_to_global()`` on global values returns a pointer within the address range of the emitted object module.  One can write to the returned value but who knows what memory is being corrupted.

pitrou: Thanks for the explanation. Could you update the docs? `get_pointer_to_global` is mentioned there.

PR #84, opened 2015-07-24
Eliminate dependencies on makefiles

achernet:
* It makes more sense to handle Darwin and FreeBSD within build.py - Windows is already being handled there anyway
* The static version of libstdc++ is added if it exists - if it doesn't, it will fall back to the dynamic version, which works fine on my system at least (and I'm running Fedora Rawhide, so breaking API changes would have shown up by now).
** Also, the behavior should be the same for all the libs that come back in <code>llvm-config --system-libs --libs all</code>, i.e. libpthread, libz, libffi, libedit, libtinfo, libdl, and libm.
* Other platforms will have "preliminary" support rather than just raising an exception


pitrou:
Is there a reason to do that, apart from stylistic preferences? I'd rather not end up reimplementing a autoconf-like system in Python...

achernet:
It improves maintainability, makes it easier to support additional platforms, some of which might not have the static library available. Examples might include building within a Cygwin environment on Windows or with Android NDK or other such nonstandard configurations. Also, having all this in Python makes it easier to write tests against it.

pitrou:
Ok, so sorry for the delay again... We have discussed this a bit and we aren't fond of that approach. If we ditch makefiles, it should be in favour of another well-known build system (e.g. cmake), not hand-written logic.

Besides removing the makefiles, is this PR fixing a specific build issue for you?

PR #87, opened 2015-08-10
Numerous improvements

whitequark:
Please see the individual commits. Sending them as separate PRs would be too noisy, so I group them in one.

Ex #82.

### DIFF
--- a/docs/source/ir/values.rst
+++ b/docs/source/ir/values.rst
@@ -151,6 +151,12 @@ Global values are values accessible using a module-wide name.
       If true, the variable is declared a constant, i.e. its contents
       cannot be ever modified.  Default is False.
 
+   .. attribute:: unnamed_addr
+
+      If true, the address of the variable is deemed insignificant, i.e.
+      it will be merged with other variables which have the same initializer.
+      Default is False.
+
    .. attribute:: initializer
 
       The variable's initialization value (probably a :class:`Constant`
--- a/docs/source/ir/values.rst
+++ b/docs/source/ir/values.rst
@@ -111,6 +111,20 @@ Values are what a :term:`module` mostly consists of.
 
       The block's :term:`terminator instruction`, if any.  Otherwise None.
 
+.. class:: BlockAddress
+
+   A constant representing an address of a basic block.
+
+   Block address constants have the following attributes:
+
+   .. attribute:: function
+
+      The function the basic block is defined in.
+
+   .. attribute:: basic_block
+
+      The basic block. Must be a part of :attr:`function`.
+
 
 Global values
 -------------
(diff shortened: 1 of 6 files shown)


seibert:
@pitrou: Can you review this?

pitrou:
Sorry for missing the previous PR. Here are some comments.

docs/source/ir/values.rst
```diff
@@ -151,6 +151,12 @@ Global values are values accessible using a module-wide name.
       If true, the variable is declared a constant, i.e. its contents
```
pitrou: You should also document the new Builder methods in ir/builder.rst.

llvmlite/ir/instructions.py
```diff
@@ -95,18 +98,40 @@ def called_function(self):
...
+    def __init__(self, parent, func, args, normal_to, unwind_to, name='', cconv=None):
+        assert isinstance(normal_to, Block)
+        assert isinstance(unwind_to, Block)
+        super(InvokeInstr, self).__init__(parent, func, args, name, cconv)
+        self.opname = "invoke"
+        self.normal_to = normal_to
+        self.unwind_to = unwind_to
+
+    def descr(self, buf):
+        super(InvokeInstr, self)._descr(buf, add_metadata=False)
```
pitrou: This simply ignores any metadata that's attached to the instruction? Why not let the metadata through, if any? If "invoke" doesn't accept any metadata, then attaching metadata to it would be a programming error.

whitequark: Oh. The next line was supposed to add missing metadata. Not sure where it went.

whitequark:
Fixed.

pitrou:
Thanks! Apparently there was a warning when building the docs:
https://travis-ci.org/numba/llvmlite/jobs/75518427#L424

whitequark:
@pitrou I don't understand why that happens. The method definitely exists and the warning does not appear when I build docs locally.

pitrou:
I'll take a look and see if I can reproduce.

pitrou:
Ok, the method exists, but it's not documented, which is why Sphinx can't find it:

```
$ grep -r add_destination source/
source/ir/builder.rst:454:   the :meth:`~IndirectBranch.add_destination` method on the return
```

whitequark:
Fixed.

pitrou:
There are still some errors. I would recommend you use the same command-line as on Travis to build locally:
```make SPHINXOPTS=-Wn clean html```

whitequark:
All errors have been fixed.

pitrou:
Thank you! I'm now merging the PR.

PR #88, opened 2015-08-12
[WIP] Fix #60: provide hooks in to the MCJIT object cache

### LINKED ISSUE #60: Provide hook into the MCJIT object cache
Ideally we would allow people to register per-module callables. See C++ example:
http://blog.llvm.org/2013/08/object-caching-with-kaleidoscope.html

pitrou:
Todo:
* Validate API with Numba
* Add documentation

### DIFF
--- a/docs/source/binding/engine.rst
+++ b/docs/source/binding/engine.rst
@@ -75,13 +75,30 @@ The ExecutionEngine class
 
    .. method:: get_global_value_address(name)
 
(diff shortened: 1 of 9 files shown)


pitrou:
The API is used successfully in https://github.com/numba/numba/pull/1360.
The callback-based design is not great to work with (it would be easier for Numba to be able to just say "give me the module's object code" or "load this object code into the module"), but LLVM doesn't provide any obvious facilities for that...

PR #90, opened 2015-08-19
Add IRBuilder.debug_metadata attribute

whitequark:
This attribute may be set by the frontend once prior to emitting
many LLVM IR instructions corresponding to a single source-level
entity (such as an AST node).

### DIFF
--- a/docs/source/ir/builder.rst
+++ b/docs/source/ir/builder.rst
@@ -19,6 +19,10 @@ pointer inside the block's list of instructions.  When adding a new
 instruction, it is inserted at that point and the pointer is then advanced
 after the new instruction.
 
+A :class:`IRBuilder` also maintains a reference to metadata describing
+the current source location, which will be attached to all inserted
+instructions.
+
 Instantiation
 -------------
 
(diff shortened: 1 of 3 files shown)


pitrou:
Hi! That's interesting because there's another PR about debugging information: PR #67. Is there an intersection with what you are doing?
Long-term, we want to support helper APIs on llvmlite to inject debug info, but LLVM doesn't expose that in a very user-friendly way. Help welcome :)

sklam:
That does not intersect with my PR #67, which focused on adding debug info on function.  This PR is for attaching source info each instruction of the body.  The approach (setting the attribute on the IRBuilder) is interesting.

whitequark:
Yes. This is the approach taken by the C++ IRBuilder and other bindings, I believe.

whitequark:
I have this for generating debug metadata: https://github.com/m-labs/artiq/blob/new-py2llvm/artiq/compiler/transforms/llvm_ir_generator.py#L21-L158

However, that code is already obsolete. We're staying on LLVM 3.5 because of a custom backend, however LLVM 3.7 and newer includes specialized metadata nodes (see http://llvm.org/docs/SourceLevelDebugging.html#id3). So it makes no sense to contribute a version of that to llvmlite.

pitrou:
Ok, this looks good for merging. Thank you for contributing!

PR #96, opened 2015-09-11
try to find .dll in local directory to fix support for frozen apps

gdementen:
Fixes (AFAIK) numba/numba#1177

With this PR, we simply need to copy llvmlite.dll to the frozen executable directory and it works.


pitrou:
Ok, this looks simple enough. I'm merging this PR, thank you for your contribution!

Issue #10, opened 2014-10-30
Add debugging information


pitrou:
Apparently, MCJIT would allow us to expose debug information so that JITted functions aren't seen as opaque under supported debuggers (such as gdb).
Unfortunately, since we aren't using the C++ API for building module code, it seems we'll have to manually emit metadata in some complicated format. But at least it should be possible.

Reference:
- http://llvm.org/docs/DebuggingJITedCode.html
- http://llvm.org/docs/SourceLevelDebugging.html
- http://llvm.org/docs/doxygen/html/classllvm_1_1DIBuilder.html

sklam:
**Update**
Interesting info from llvm ml http://lists.cs.uiuc.edu/pipermail/llvmdev/2014-March/071443.html
Summary: need debug symbol in LLVM binaries

whitequark:
Also fixed by #162.

pitrou:
Here is the updated link to the llvm-dev message mentioned above: http://lists.llvm.org/pipermail/llvm-dev/2014-March/071405.html

pitrou:
Ok, our compiled `lli` (in the `llvmdev` package) works fine with the example at http://lists.llvm.org/pipermail/llvm-dev/2014-March/071417.html, and our `libllvmlite.so` seems to have the two required symbols as well:
```
$ nm --defined-only llvmlite/binding/libllvmlite.so | \grep jit_debug
00000000016af7c0 d __jit_debug_descriptor
00000000005d6480 t __jit_debug_register_code
```
(about those symbols: https://sourceware.org/gdb/onlinedocs/gdb/Declarations.html#Declarations, https://www.sourceware.org/ml/gdb-patches/2009-07/msg00571.html)

Of course, a real experiment will be better, and then we can close this issue.

pitrou:
See PR #200.

PR #98, opened 2015-09-28
Support profiing with OProfile

markdewing:
Add the OProfile JIT event listener to allow the JIT'ed code
to be analyzed with a hardware profiler on Linux.

To enable this code:
 - enable LLVM for OProfile JIT events (-DLLVM_USE_OPROFILE=ON)
 - define USE_OPROFILE_JITEVENTS when compiling llvmlite
 - define environment variable ENABLE_JITPROFILING
 - LD_LIBRARY_PATH may need to point to the directory containing
        libopagent.so (/usr/lib/oprofile)

Tested with LLVM 3.6

Annotating assembly works (with opannotate), but I think debug information will be needed to annotate source (see issue #10 )

### DIFF
--- a/ffi/executionengine.cpp
+++ b/ffi/executionengine.cpp
@@ -4,9 +4,7 @@
 
 #include "llvm/IR/Module.h"
 #include "llvm/ExecutionEngine/ExecutionEngine.h"
-#ifdef USE_OPROFILE_JITEVENTS
 #include "llvm/ExecutionEngine/JITEventListener.h"
-#endif
 #include "llvm/ExecutionEngine/ObjectCache.h"
 #include "llvm/Support/Memory.h"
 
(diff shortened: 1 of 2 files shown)


pitrou:
Oh, thank you for doing this. Do you think it's possible to have a Python API to enable profiling, rather than an environment variable? Also, does it also work with Linux `perf`?

markdewing:
I will look into making a Python API instead.

The `perf` tool does have a simple JIT interface, but I think it would require some support in LLVM to write the proper file.

markdewing:
Updated with a potential Python API.

pitrou:
Thanks. The API looks good enough for now. For Linux perf it may be possible to write a specific event listener class to write the specific file (Mesa does this which looks rather hackish: http://lists.freedesktop.org/archives/mesa-commit/2013-April/042805.html).

PR #104, opened 2015-10-16
Add binding to get DOT string of control flow graph 

sklam:
Alternative for https://github.com/numba/llvmlite/pull/64 because the `viewCFG` api from llvm dumps the DOT file to random temp file and tries to open it with `gv`.  Instead, this PR gets the DOT string of the CFG and returns it.


sklam:
This addition will be useful for debugging in numba.

llvmlite/tests/test_binding.py
```diff
@@ -15,6 +15,7 @@
 from llvmlite import six, ir
 from llvmlite import binding as llvm
 from llvmlite.binding import ffi
+from llvmlite import ir
```
pitrou: The import is already done three lines above.

pitrou:
Ok, I merged despite the tiny nit mentioned above. This can be fixed in a later changeset if we care enough.

sklam:
Oops, thanks

sklam:
@pitrou, I cleaned up some dead import at https://github.com/numba/llvmlite/pull/204/files

PR #106, opened 2015-10-20
[WIP] expose more controlflow analysis

sklam:
This PR adds more controlflow analysis from llvm.  This will expose analysis for:
* Regions
* Post Dominators
* Dominators
* Dominance Frontiers
* Loop 

Depends on https://github.com/numba/llvmlite/pull/104


sklam:
This is probably not useful right now.

PR #110, opened 2015-11-25
Adapt Travis config for Trusty beta

pitrou:
This avoids having to install and activate a gcc backport.

See https://docs.travis-ci.com/user/trusty-ci-environment/


pitrou:
Ok, it seems the builds are currently *very* slow (insufficient allocated infrastructure?).

PR #112, opened 2015-12-18
Add support for non-alphanumeric characters in type names

whitequark:


seibert:
Seems reasonable to me.  @sklam: Any objections?

sklam:
I think we don't need to replace space once we have the doublequote

whitequark:
Hm right, I copied this from Value. It shouldn't be needed there either, let me update.

whitequark:
Fixed.

llvmlite/ir/types.py
```diff
@@ -17,6 +17,10 @@
 TYPE_METADATA = _type_enum()
 
 
+def _wrapname(x):
+    return '"{0}"'.format(x.replace('\\', '\\5c').replace('"', '\\22'))
```
sklam: can you add tests for '\\' and '"' ?

whitequark: Done

whitequark:
Ping

sklam:
Thanks!

PR #113, opened 2016-01-02
Let Type.get_abi_{size,alignment} not choke on identified types.

whitequark:
If an identified struct type from a context that is not global
is used anywhere in the type that is passed to
get_abi_{size,alignment}, the generated LLVM IR is invalid.

This pull request adds a `context` argument, which allows
to specify a non-global context explicitly.

### DIFF
--- a/llvmlite/ir/types.py
+++ b/llvmlite/ir/types.py
@@ -42,31 +42,34 @@ class Type(object):
     def __ne__(self, other):
         return not (self == other)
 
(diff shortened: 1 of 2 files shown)


pitrou:
That's a nice fix. Perhaps you could add some tests for this?

whitequark:
Done

pitrou:
Thank you!

PR #114, opened 2016-01-04
Use conda for readthedocs

pitrou:


abebeos:
@esc, to use or not to use?

esc:
Hard to say, not sure what the current state is. But I think @sklam knows?

sklam:
It's okay for us to switch. We switched to conda in numba already. The docs is not build with llvmdev, nor does it need installing llvmlite. The main advantage to use conda is consistency with local development given that we are using conda primarily.

sklam:
Intentionally closed and reopened to retrigger checks.

sklam:
It's building fine. Perhaps we should bump the python version.

esc:
Closin in favour of #711

PR #132, opened 2016-03-05
Correct constant in llvmir.py example

ian-bertolacci:
using `builder.constant` instead of `ll.constant` resulting in 'IRBuilder' object has no attribute 'Constant' error.


pitrou:
Nice, thank you!

PR #134, opened 2016-03-08
Fix #131: allow easier creation of constants

### LINKED ISSUE #131: Allow easier creation of constants
Right now to e.g. increment an integer you have to:
```
c = ir.Constant(op.type, 1)
res = builder.add(op, c)
```
It would be easier to write:
```
res = builder.add(op, op.type(1))
```
... that is, allow creating a constant directly by calling a type instance with the intended value as argument.

pitrou:
Allow creating constants by calling a type instance (e.g. IntType(32)(42)).
Also, allow recursive creation of aggregate constants
(e.g. ArrayType(int32, 3)([4, 5, 6])).


pitrou:
@stuartarchibald

llvmlite/tests/test_ir.py
```diff
@@ -276,11 +287,15 @@ def test_simple(self):
...
-        builder.add(a, b, 'res')
+        inst = builder.add(a, b, 'res')
         self.check_block(block, """\
             my_block:
                 %"res" = add i32 %".1", %".2"
             """)
+        # Instructions should have a useful repr()
+        self.assertEqual(repr(inst),
+                         "<ir.Instruction 'res' of type 'i32', opname 'add', "
+                         "operands (<ir.Argument '.1' of type i32>, <ir.Argument '.2' of type i32>)>")
```
stuartarchibald: thanks :)

stuartarchibald:
This seems to cover the intended. Thanks for the patch, it'll make using llvmlite easier and the use of constants cleaner.

PR #141, opened 2016-03-09
MAINT: ensure no CMakeCache.txt file is included in sdist.

### LINKED ISSUE #140: pypi tarball contains stray cmake build files
I had some issues with building llvmlite 0.9.0 on windows, because the pypi tarball contains `CMakeCache.txt` files in `ffi/build`. I think those should be blacklisted in your MANIFEST.in

cournape:
Having stray `CMakeCache.txt` files cause trouble when building on windows. Removing them is easier than trying to understand cmake :)

Fix #140


pitrou:
Thank you!

PR #144, opened 2016-03-11
Allow overriding flto in ffi builds on Linux.

eliben:
Background: Clang & LLVM's binary releases are now built with Clang itself,
which means we have to compile llvmlite with Clang when compiling vs. a binary
release taken from http://llvm.org/releases/

However, in that case Clang looks for the Gold plugin for LTO, which is not
built into the binaries by default. So this makes building llvmlite on your own
very cumbersome.

Alternatively we can just not use LTO. This change permits overriding the LTO
setting from the environment when needed.


pitrou:
Good, thank you!

PR #146, opened 2016-03-18
Clean command added to setup.py

ml31415:
It's a bit annoying to remove all temporary files before building from scratch. This PR adds a simple clean command to fix this.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,10 +1,10 @@
 try:
-    from setuptools import setup, Extension
+    from setuptools import setup, Extension, Command
     from setuptools.command.build_py import build_py as build
     from setuptools.command.build_ext import build_ext
     from setuptools.command.install import install
 except ImportError:
-    from distutils.core import setup, Extension
+    from distutils.core import setup, Extension, Command
     from distutils.command.build import build
     from distutils.command.build_ext import build_ext
     from distutils.command.install import install
@@ -74,9 +74,22 @@ class LlvmliteInstall(install):
         install.run(self)
 
 
+class LlvmliteClean(Command):
+    """Custom clean command to tidy up the project root."""
+    user_options = []
+    def initialize_options(self):
+        pass
+    def finalize_options(self):
+        pass
+    def run(self):
+        os.system('rm -vrf ./build ./dist ./ffi/*.so ./*.egg-info')
+        os.system('find . -name "*.pyc" -delete')
+
+
 cmdclass.update({'build': LlvmliteBuild,
                  'build_ext': LlvmliteBuildExt,
                  'install': LlvmliteInstall,
+                 'clean': LlvmliteClean,
                  })
 
 packages = ['llvmlite',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -74,9 +74,22 @@ def run(self):
...
 
+class LlvmliteClean(Command):
+    """Custom clean command to tidy up the project root."""
+    user_options = []
+    def initialize_options(self):
+        pass
+    def finalize_options(self):
+        pass
+    def run(self):
+        os.system('rm -vrf ./build ./dist ./ffi/*.so ./*.egg-info')
```
pitrou: This is really system-specific. If you want to trigger a rebuild on Unix, `touch ffi/*.cpp` should be enough. If we want a `setup.py clean` command, it should also work on Windows...
(also, distutils / setuptools has its own `clean` command, so we should probably derive from it)

ml31415: The setuptools clean seemd to miss the library. About windows, yeah, didn't had that in mind. If the patch is welcome in general, I'd fix that.

pitrou: On the principle I think the proposal is sound, yes.

ml31415:
I don't have any windows machine around, so this is completely untested there.

setup.py
```diff
@@ -74,9 +74,45 @@ def run(self):
         install.run(self)
 
 
+class LlvmliteClean(Command):
```
pitrou: Can't you inherit from setuptools' or distutils' own `clean` command class? I don't think we want to reinvent the removal of `egg-info` or `pyc` files (not to mention any other things you may have overlooked here).

ml31415: I didn't see a clean command in setuptools, only in distutils. Looks like this is also not too concise https://bitbucket.org/carljm/python-distutils/src/48c42eeaee4410d76675b637bcd401b8919ff19a/command/clean.py?at=default&fileviewer=file-view-default But I agree, it would make sense to derive from it anyways. I basically had taken the next best receipe and adapted it to some extended needs.

ml31415:
Rebased on master and subclassed from distutils now. Please have another look.

setup.py
```diff
@@ -3,14 +3,20 @@
     from setuptools.command.build_py import build_py as build
     from setuptools.command.build_ext import build_ext
     from setuptools.command.install import install
+    from setuptools.command.clean import clean
```
pitrou: If it never exists then let's never try importing it?

ml31415: I didn't see it here: https://github.com/pypa/setuptools/tree/master/setuptools/command but I don't know if this is a never never for any version out there.

ml31415: Another question, if it depends on `distutils` then anyways, shouln't the attempt to import from `setuptools` be thrown out completely?

pitrou: That's a good question. `setuptools` does some amount of monkeypatching, so perhaps `from setuptools import setup, Extension` is enough.

ml31415:
Can this build be retried? I'm not sure if this crash is related to the changes.

pitrou:
That seems to have been a transient Travis-CI issue, it's fine now.

pitrou:
Ok, I've merged. Thank you!

PR #149, opened 2016-03-24
Implement LLVM scoping correctly

whitequark:
LLVM has three scopes: a context-level scope that contains named
types (which have a %name); a module-level scope that contains named
globals (which have a @name); and a function-level scope that
contains instructions (which have a %name). Names in all three scopes
cannot conflict: either the sigil is different, or the sigil is
the same but the names appear in disjoint lexical positions.

So, remove all the logic responsible for handling nested scopes,
since it is pointless.

### DIFF
--- a/llvmlite/ir/_utils.py
+++ b/llvmlite/ir/_utils.py
@@ -8,18 +8,12 @@ class DuplicatedNameError(NameError):
 
 
 class NameScope(object):
-    def __init__(self, parent=None):
-        self.parent = parent
+    def __init__(self):
         self._useset = set([''])
         self._basenamemap = defaultdict(int)
 
     def is_used(self, name):
-        scope = self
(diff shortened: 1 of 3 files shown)


whitequark:
Looks like the decrease in coverage is spurious.

pitrou:
Oh, great. I've always wondered if this was useful.

sklam:
This looks good.

I didn't realize the following IR is valid:

```llvm
; ModuleID = '<stdin>'

%apple = type opaque

define %apple @apple(%apple %apple) {
.3:
  ret %apple %apple
}
```

whitequark:
https://en.wikipedia.org/wiki/Buffalo_buffalo_Buffalo_buffalo_buffalo_buffalo_Buffalo_buffalo :)

PR #151, opened 2016-03-26
Always copy TargetData when adding to a pass manager

whitequark:
There was no check of self._owned, which could lead to a segfault
during the straightforward use of the API. It is essentially never
useful to have a reference to TargetData owned by the pass manager,
so just always make a copy instead.


pitrou:
Thank you!

PR #152, opened 2016-03-31
Expose initialize_all_{targets,asmprinters}

whitequark:
Without this, it's not possible to emit assembly or code directly using llvmlite, which severely limits its usefulness; one would have to resort to executing llc, which means search path problems, Windows compatibility problems, etc. So, the resulting enlargement of the package is worth it.

### DIFF
--- a/docs/source/binding/initfini.rst
+++ b/docs/source/binding/initfini.rst
@@ -11,6 +11,16 @@ These functions need only be called once per process invocation.
 
    Initialize the LLVM core.
 
+.. function:: initialize_all_targets()
+
+   Initialize all targets. Necessary before targets can be looked up
+   via the :class:`Target` class.
+
+.. function:: initialize_all_asmprinters()
+
+   Initialize all code generators. Necessary before generating
+   any assembly or machine code via the :meth:`TargetMachine.emit_object`
+   and :meth:`TargetMachine.emit_assembly` methods.
 
 .. function:: initialize_native_target()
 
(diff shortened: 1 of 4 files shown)


whitequark:
The `initialize_all_targets` function is grouped like that to match the C API, which is probably what most users expect.

pitrou:
> Without this, it's not possible to emit assembly or code directly using llvmlite

What do you mean by that?

> So, the resulting enlargement of the package is worth it.

How much is it exactly?

whitequark:
> Without this, it's not possible to emit assembly or code directly using llvmlite

You need asmprinters to emit assembly or machine code (yes, asmprinters are responsible for machine code too) via TargetMachine. `llvm.initialize_native_target()` or creating a JIT implicitly do that for the backend that matches the current architecture, but if you need to emit code for a different architecture, that doesn't work.

Our use case is that we use LLVM and llvmlite to emit code for an OpenRISC1000 softcore running on an FPGA.

> How much is it exactly?

That's impossible to say because it varies by architecture and with LLVM build options. Generally, a small target is a few hundred kB and a large target is a few MB (when built without debug information, of course). X86 and ARM are probably the only large ones. I haven't built llvmlite with upstream LLVM but I'd expect the package to grow to around 20MB.

pitrou:
Ok, I've tried on our own LLVM builds, and it keeps the size almost unchanged, which is good :-)
Can you add some documentation for the new functions?

whitequark:
You probably have only the X86 backend then.

I will add docs.

whitequark:
Docs added.

whitequark:
Don't you find it laborous to manually duplicate all the docstrings in the Sphinx documentation?

pitrou:
Not that much. It is common to write more detailed explanations and/or example in the Sphinx documentation.

pitrou:
Ok, thank you!

PR #153, opened 2016-03-31
Remove deprecated uses of TargetMachine::getDataLayout()

whitequark:
`TargetMachine::getDataLayout()` is going away in LLVM 3.8. `TargetMachine::createDataLayout()` returns `DataLayout` by value, so I'm wrapping it in `new`.

### DIFF
--- a/ffi/targets.cpp
+++ b/ffi/targets.cpp
@@ -29,6 +29,10 @@ inline Target *unwrap(LLVMTargetRef T) {
     return reinterpret_cast<Target*>(T);
 }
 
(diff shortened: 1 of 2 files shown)


pitrou:
Does it have any semantic consequences, or is it just an implementation detail?

whitequark:
No semantic consequences, now or in the future. The reason the change is made upstream is because the R600 backend has data layout varying by subtarget, and having just one data layout owned by the target machine doesn't allow that.

pitrou:
Ok, I'm merging then. Thanks!

PR #154, opened 2016-03-31
Expose functions to create individual passes

whitequark:
This is just a bunch of passes our compiler needs, since exposing them all is a lot of work for unclear benefit (indeed, exposing passes no one needs in practice just adds maintenance burden, as they are removed, renamed, etc in newer LLVM versions). However, this list should nevertheless be generally useful.

### DIFF
--- a/docs/source/binding/passmanager.rst
+++ b/docs/source/binding/passmanager.rst
@@ -56,13 +56,79 @@ configure a :class:`PassManagerBuilder`.
 
 .. class:: PassManager
 
-   The base class for pass managers.
+   The base class for pass managers. Use individual ``add_*`` methods
+   or :meth:`PassManagerBuilder.populate` to add optimization passes.
 
+
+   .. function:: add_constant_merge_pass()
+
+      See `pass documentation <http://llvm.org/docs/Passes.html#constmerge-merge-duplicate-global-constants>`_.
+
+   .. function:: add_dead_arg_elimination_pass()
+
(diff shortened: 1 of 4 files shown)


pitrou:
Interesting. Are these not enabled by default with `-O3`? What is the benefit of enabling them specifically? This might benefit us if we know how to select them...

By the way, this would deserve docstrings and perhaps documentation entries.

whitequark:
> Are these not enabled by default with -O3? What is the benefit of enabling them specifically?

The default pipeline (the one used by PassManagerBuilder) is tailored to the sort of IR emitted by clang. We do not emit code that looks like it, e.g. there are basically no available optimizations before SROA and GlobalOpt. Thus, we have a completely custom pipeline; e.g. we run SROA and GlobalOpt right away. Removing the passes that produce no useful results on the code emitted by our frontend also helps reducing compile time, which is a major concern for us.

> By the way, this would deserve docstrings and perhaps documentation entries.

I can add links to http://llvm.org/docs/Passes.html, yes.

whitequark:
Documentation added. Note that the SROA pass, ScalarRepl pass, and ScalarRepl-SSA passes are all distinct, with the latter two being historical aberrations that are going to be removed once someone remembers to do it.

The links are to the ToT LLVM documentation, which can change underneath llvmlite. I'm not sure if we should link to 3.7 directly instead, or something.

The TBAA pass appears to be completely undocumented. Frankly, if you want TBAA, you'll have to become accustomed to reading the LLVM source anyway, so I don't see this as a major omission. :)

PR #155, opened 2016-03-31
Properly emit IR for metadata null

whitequark:
Before, it was emitted as `!0 = { zeroinitializer }`.

### DIFF
--- a/llvmlite/tests/test_ir.py
+++ b/llvmlite/tests/test_ir.py
@@ -224,6 +224,12 @@ class TestIR(TestBase):
         self.assertInText("!bar = !{ !0, !0 }", str(mod))
         self.assert_valid_ir(mod)
 
+    def test_metadata_null(self):
+        mod = self.module()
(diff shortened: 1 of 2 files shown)


llvmlite/tests/test_ir.py
```diff
@@ -224,6 +224,12 @@ def test_named_metadata_2(self):
         self.assertInText("!bar = !{ !0, !0 }", str(mod))
         self.assert_valid_ir(mod)
 
+    def test_metadata_null(self):
+        mod = self.module()
+        md = mod.add_metadata([ir.Constant(ir.PointerType(ir.IntType(32)), None)])
```
pitrou: Note this can be more concisely written as `ir.IntType(32).as_pointer()(None)`.

whitequark: Do you want me to change it?

pitrou: No, it's ok. Just pointing it out :-)

pitrou:
Thanks again :-)

PR #158, opened 2016-04-01
[WIP] Update Linker for changes in 3.8

whitequark:
This PR is a bit drastic, so let me support the logic behind my changes with a timeline.
  * LLVM adds LinkModules. LinkModules accepts a PreserveSource argument, which was added because someone thought it was useful.
  * PreserveSource argument is not useful. No one uses it.
  * Bit rot sets in.
  * llvmlite implements link_modules, exposing the PreserveSource argument, because whoever wrote that binding decided it was useful.
  * *years pass*
  * PreserveSource in LLVM is discovered to be a buggy feature that no one noticed was buggy because no one used it. It was removed in r220741.

Thus, I think it should just be removed. It's not really the job of llvmlite to emulate obsolete LLVM features that were ripped out from LLVM because no one wanted them. Even if someone does want it, the changes to the frontend are trivial, and they properly highlight the cost of the operation (copying).

As for the diagnostic change: LLVM now reports all errors via the diagnostic handler set via `LLVMContextSetDiagnosticHandler`: linker errors, inline assembly errors, possibly others I don't know of... I feel it is the function that should be exposed, although I'm not sure just quite how. In the meantime, returning a generic error might just be fine, although I would consider defining a separate exception class here.

### DIFF
--- a/ffi/linker.cpp
+++ b/ffi/linker.cpp
@@ -1,43 +1,12 @@
 #include "core.h"
-#include "llvm/IR/Module.h"
-#include "llvm/IR/DiagnosticPrinter.h"
-#include "llvm/Linker/Linker.h"
-#include "llvm/Support/raw_ostream.h"
+#include "llvm-c/Linker.h"
 
 extern "C" {
 
 API_EXPORT(int)
-LLVMPY_LinkModules(LLVMModuleRef Dest, LLVMModuleRef Src, int Preserve,
-                   const char **Err)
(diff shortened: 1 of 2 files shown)


pitrou:
If this is not compatible with 3.7, this should be part of a larger 3.8 migration PR.
Also, we should really preserve the original LLVM error message, instead of raising a generic exception. Can't we override the diagnostic handler for the specific linking operation?

pitrou:
This should be superseded by PR #176, which as a proper solution to fetch the linking error message.

PR #159, opened 2016-04-01
Allow emitting verbose assembly

whitequark:
See the docstring for explanation.

### DIFF
--- a/docs/source/binding/targets.rst
+++ b/docs/source/binding/targets.rst
@@ -154,6 +154,11 @@ Classes
       a code object, suitable for use with the platform's linker.
       A bytestring is returned.
 
+   .. method:: set_asm_verbosity(is_verbose)
+
+      Set whether this target machine will emit assembly with human-readable
+      comments describing control flow, debug information, and so on.
+
    .. method:: emit_assembly(module)
 
       Return the compiled *module*'s native assembler, as a string.
--- a/llvmlite/binding/targets.py
+++ b/llvmlite/binding/targets.py
@@ -246,6 +246,13 @@ class TargetMachine(ffi.ObjectRef):
         """
         ffi.lib.LLVMPY_AddAnalysisPasses(self, pm)
 
+    def set_asm_verbosity(self, verbose):
+        """
+        Set whether this target machine will emit assembly with human-readable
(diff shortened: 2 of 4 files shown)


llvmlite/binding/targets.py
```diff
@@ -246,6 +246,13 @@ def add_analysis_passes(self, pm):
         """
         ffi.lib.LLVMPY_AddAnalysisPasses(self, pm)
 
+    def set_verbose(self, verbose):
```
pitrou: How about a more descriptive name? `set_asm_verbosity` perhaps?

llvmlite/tests/test_binding.py
```diff
@@ -632,6 +632,7 @@ def test_emit_assembly(self):
         target_machine = self.target_machine()
         mod = self.module()
         ee = self.jit(mod, target_machine)
+        target_machine.set_verbose(True)
```
pitrou: I think we should test both the non-verbose and the verbose case (perhaps with different checks?).

whitequark: The exact structure of the assembly comments, including the character that starts the comment, is dependent on the selected target, so I'm not sure how. Besides there isn't much to break, except the function not being called correctly...

llvmlite/binding/targets.py
whitequark: Ah, yes, will do.

whitequark:
Updated

pitrou:
Thank you! Can you add documentation for the new method in `docs/source/binding/targets.rst`?

whitequark:
Done

docs/source/binding/targets.rst
```diff
@@ -154,6 +154,11 @@ Classes
       a code object, suitable for use with the platform's linker.
       A bytestring is returned.
 
+   .. method:: set_asm_verbosity(module, is_verbose)
```
pitrou: Just a nit: the method doesn't take a *module* argument.

whitequark: Ah right, fixed.

pitrou:
Thanks! AppVeyor is currently broken because of a `conda` issue. I'm merging this PR.

PR #160, opened 2016-04-02
Escape the contents of metadata strings

whitequark:
Also, allow all printable ASCII characters in IR string literals,
including metadata strings and character array initializers.

### DIFF
--- a/llvmlite/ir/values.py
+++ b/llvmlite/ir/values.py
@@ -7,20 +7,22 @@ from __future__ import print_function, absolute_import
 
 import string
 
-from ..six import StringIO
+from ..six import StringIO, text_type
 from . import types, _utils
 from ._utils import _StrCaching, _StringReferenceCaching, _HasMetadata
 
 
 _VALID_CHARS = (frozenset(map(ord, string.ascii_letters)) |
                 frozenset(map(ord, string.digits)) |
-                frozenset('._-$'))
+                frozenset(map(ord, ' !#$%&\'()*+,-./:;<=>?@[]^_`{|}~')))
 
 
 def _escape_string(text, _map={}):
     """
     Escape the given bytestring for safe use as a LLVM array constant.
     """
+    if isinstance(text, str):
+        text = text.encode('ascii')
     assert isinstance(text, (bytes, bytearray))
 
     if not _map:
@@ -258,7 +260,7 @@ class MetaDataString(NamedValue):
         buf += (self.get_reference(), "\n")
 
     def _get_reference(self):
-        return '!"{0}"'.format(self.string)
+        return '!"{0}"'.format(_escape_string(self.string))
 
     _to_string = _get_reference
 
--- a/llvmlite/ir/values.py
+++ b/llvmlite/ir/values.py
@@ -7,7 +7,8 @@ from __future__ import print_function, absolute_import
 
 import string
 
-from ..six import StringIO, text_type
+from .. import six
+from ..six import StringIO
 from . import types, _utils
 from ._utils import _StrCaching, _StringReferenceCaching, _HasMetadata
 
@@ -31,6 +32,8 @@ def _escape_string(text, _map={}):
                 _map[ch] = chr(ch)
             else:
                 _map[ch] = '\\%02x' % ch
+            if six.PY2:
+                _map[chr(ch)] = _map[ch]
 
     buf = [_map[ch] for ch in text]
     return ''.join(buf)
--- a/llvmlite/tests/test_ir.py
+++ b/llvmlite/tests/test_ir.py
@@ -202,6 +202,12 @@ class TestIR(TestBase):
         self.assertInText(pat1, str(mod))
         self.assertInText(pat2, str(mod))
 
+    def test_metadata_string(self):
+        mod = self.module()
+        mod.add_metadata([ir.MetaDataString(mod, "\"\\$")])
+        pat = '!0 = !{ !"\\22\\5c$" }'
+        self.assertInText(pat, str(mod))
+
     def test_named_metadata(self):
         mod = self.module()
         md = mod.add_metadata([ir.Constant(ir.IntType(32), 123)])
@@ -1357,7 +1363,7 @@ class TestConstant(TestBase):
         self.assertEqual(str(c), '[2 x i32] zeroinitializer')
         # Raw array syntax
         c = ir.Constant(ir.ArrayType(int8, 11), bytearray(b"foobar_123\x80"))
-        self.assertEqual(str(c), r'[11 x i8] c"foobar\5f123\80"')
+        self.assertEqual(str(c), r'[11 x i8] c"foobar_123\80"')
         c = ir.Constant(ir.ArrayType(int8, 4), bytearray(b"\x00\x01\x04\xff"))
         self.assertEqual(str(c), r'[4 x i8] c"\00\01\04\ff"')
         # Recursive instantiation of inner constants


llvmlite/ir/values.py
```diff
@@ -14,13 +14,15 @@
 
 _VALID_CHARS = (frozenset(map(ord, string.ascii_letters)) |
                 frozenset(map(ord, string.digits)) |
-                frozenset('._-$'))
+                frozenset(map(ord, ' !#$%&\'()*+,-./:;<=>?@[]^_`{|}~')))
```
whitequark: This was actually broken before, since it didn't have `map(ord, ...)`.

whitequark:
I strongly suspect based on the nature of the fix I had to do, but haven't bothered to prove, that `_escape_string` has never properly worked on Python 2.

whitequark:
Yeah, it would have broke if someone passed it a `bytes` constant. Have you yet considered doing the appropriate actions to handle Python 2, i.e. killing it with fire?

whitequark:
The Appveyor error is bogus.

llvmlite/ir/values.py
pitrou: How about the reverse, i.e. only use `ord` on Python 3?

llvmlite/ir/values.py
```diff
@@ -14,13 +15,15 @@
...
 
 
 def _escape_string(text, _map={}):
     """
     Escape the given bytestring for safe use as a LLVM array constant.
     """
+    if isinstance(text, six.text_type):
```
pitrou: I'm not sure why this is necessary? You should never pass a unicode string here...

pitrou:
> Have you yet considered doing the appropriate actions to handle Python 2, i.e. killing it with fire?

I would like that, but many of our users (including professional customers of Continuum products) still use Python 2...

llvmlite/ir/values.py
whitequark: Initially I did `isinstance(text, str)` since otherwise the assert below would fail on Python 3, but that failed on Python 2 because of the superfluous `encode` call...

pitrou: But why did you need that at all? You should never pass a text string there, only a bytes or bytearray object.

pitrou: In other words, if you need an encode() call, it should be in `metadata._get_reference()`, not here.

whitequark: I don't see why. This code is common for all double-quoted strings in the IR, whether originally strings (and thus user-facing UTF-8 identifiers) or byte arrays (probably emitted by the compiler).

pitrou: Passing an unicode string as an array constant feels like a programmer error to me, so I'd rather raise an exception. I don't know about metadata strings.

llvmlite/ir/values.py
whitequark: Well, that adds a py2/py3 branching, which means more opportunity for bugs. I don't see this as necessary since this code is executed exactly once.

whitequark: Well, I guess there's this branch below...
```

    if six.PY2 and isinstance(text, bytes):
        buf = [_map[ord(ch)] for ch in text]
    else:
        buf = [_map[ch] for ch in text]
```
so using `ord()` would pessimize py3. This is a rather hot function since IR contains many strings, and I don't think that should be done.

llvmlite/ir/values.py
whitequark: Removed the `encode` call.

whitequark:
Updated.

llvmlite/ir/instructions.py
```diff
@@ -162,7 +162,7 @@ def descr(self, buf):
 class PredictableInstr(Instruction):
 
     def set_weights(self, weights):
-        operands = [MetaDataString(self.module, "branch_weights")]
+        operands = [MetaDataString(self.module, b"branch_weights")]
```
pitrou: For the required, is this change required by the rest of the patch?

llvmlite/ir/values.py
pitrou: Hmm, I understand. Perhaps `_map` can contain both cases on py2 instead? That way, the branch would happen at construction time. Something like:
```
if six.PY2:
    _map[chr(ch)] = _map[ch]
```

llvmlite/ir/instructions.py
whitequark: Yes, since you've requested the encode call to be removed. This worked before by accident.

llvmlite/tests/test_ir.py
```diff
@@ -202,12 +202,18 @@ def test_metadata_2(self):
         mod = self.module()
         mod.add_metadata([ir.Constant(ir.IntType(32), 123)])
         mod.add_metadata([ir.Constant(ir.IntType(32), 321),
-                          ir.MetaDataString(mod, "kernel")])
+                          ir.MetaDataString(mod, b"kernel")])
```
pitrou: If `MetaDataString` starts accepting both bytes and unicode, we should probably test both here.

whitequark: It doesn't, that's why it's changed. It only accepts bytes and bytearray.

llvmlite/ir/values.py
whitequark: I did that (well, something similar).

llvmlite/ir/instructions.py
pitrou: Hmm, I see. It seems there are two conflicting constraints here:
- `_escape_string`, given its current docstring ("Escape the given bytestring for safe use as a LLVM array constant"), should not accept unicode strings
- metadata strings definitely feel like "text" and it would be awkward to refuse Python 3 string literals for them (Numba has a couple examples of them and I'm sure other users would, too)

So perhaps `_escape_string` should be renamed to e.g. `_escape_bytes` (for use by array constants), and a wrapper function `_escape_string` created (for use by metadata strings). What do you think?

whitequark: No, this is completely wrong. What it "feels" is irrelevant. What matters is exactly one thing: how we consider `.ll` files to be encoded. There are two (reasonable) options:
 1. `.ll` files are ASCII, and UTF-8 cannot be put into them.
 2. `.ll` files are UTF-8-encoded.
`master` has a random mixture of these two. This patchset currently fixes it to represent option 1. If that's undesirable then it should be fixed to use option 2 (which I prefer anyway, since the instruction names are human-readable and LLVM doesn't care what's in there, and putting a `unicode` into a character array has a perfectly valid use, namely emitting UTF-8 string literals in a frontend).

whitequark: Correction: the previous comment should say "the strings in the [abstract] LLVM IR", not "`.ll` files".

pitrou: Well, the LLVM Language Reference doesn't say anything about the character encoding in which the IR is to be interpreted, so we have to find some compromise.

Arrays contain binary data, this is certain. Accepting text for them is at odds with the Python 3 model of separating text and bytes: if a frontend knows a string must be utf-8 encoded, therefore, it should encode it itself (after all, the string could be utf-16 encoded instead, e.g. for use with the Windows Unicode filesystem APIs).

I would lean towards considering metadata strings are human-readable text, but I suppose there may be cases where they are not. My concern here is that most existing code probably passes a `str` value (i.e. unicode on Python 3), not a `bytes`.

whitequark: There is no real difference between allowing to pass `unicode` to metadata strings and to arrays. The latter is a way to communicate with your compiled program, the former is a way to communicate with various framework around your program. In one case you are assuming that your program should see human-readable text as UTF-8, in another you are assuming that your e.g. debugger or symbolizer should see human-readable text as UTF-8.

Metadata strings contain such things as function names and paths, neither of which are *required* to be UTF-8, or human-readable text for that matter. Indeed, on Linux, and depending on how you write your code, on Windows, there's no guarantee that paths are valid UTF-8.

And I think in both cases it's a reasonable assumption that any `unicode` passed is a human-readable text that should be converted to UTF-8, because if you're not representing human-readable text as UTF-8 then something is wrong with you.

pitrou: > Metadata strings contain such things as function names and paths, neither of which are required to be UTF-8, or human-readable text for that matter.

That's true. However, those (function names, file paths) are precisely considered unicode strings on Python 3, so it makes sense that MetaDataString respects this. Or, at least, MetaDataString could encode them using the ASCII codec, which is safer (people would have to encode non-ASCII text manually, which prevents llvmlite from having to guess).

whitequark: If so then we can simply restore the old commit but with a `.encode('ascii')` call rather than `.encode('utf-8')`. That would preserve the option 1 being the default behavior but also allow putting strings in various places.

sklam:
To make it easier for users and to not breaking most of the existing code, I would prefer @pitrou's suggestion of:

>  Or, at least, MetaDataString could encode them using the ASCII codec, which is safer (people would have to encode non-ASCII text manually, which prevents llvmlite from having to guess).

pitrou:
As a side note, the tests currently fail on 2.7:
https://travis-ci.org/numba/llvmlite/jobs/138080937#L624-L640

whitequark:
So, any decision on this?

pitrou:
I would selfishly defend the proposal that @sklam supported above :-) Basically, "practicality beats purity".

whitequark:
Ok, that actually doesn't sound too bad to me.

whitequark:
Updated, py2 issues fixed.

pitrou:
Great, thanks!

PR #161, opened 2016-04-02
Allow adding metadata to functions

whitequark:
This was brought in around LLVM 3.7 and is only currently used for debug information.

I feel like the `leading_comma` parameter, when it exists, should default to `False`, hence a bit of churn below.

### DIFF
--- a/llvmlite/ir/_utils.py
+++ b/llvmlite/ir/_utils.py
@@ -53,3 +53,23 @@ class _StringReferenceCaching(object):
         except AttributeError:
             s = self.__cached_refstr = self._get_reference()
             return s
+
+
+class _HasMetadata(object):
+
+    def set_metadata(self, name, node):
+        """
+        Attach unnamed metadata *node* to the metadata slot *name* of this value.
+        """
+        self.metadata[name] = node
+
+    def _stringify_metadata(self, leading_comma=False):
+        if self.metadata:
+            buf = []
+            if leading_comma:
+                buf.append("")
+            buf += ["!{0} {1}".format(k, v.get_reference())
+                    for k, v in self.metadata.items()]
+            return ', '.join(buf)
+        else:
+            return ''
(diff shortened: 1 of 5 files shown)


llvmlite/ir/_utils.py
```diff
@@ -53,3 +53,20 @@ def get_reference(self):
         except AttributeError:
             s = self.__cached_refstr = self._get_reference()
             return s
+
+
+class _HasMetadata(object):
```
pitrou: Can you add a docstring here and on `.set_metadata()`?

llvmlite/tests/test_ir.py
```diff
@@ -127,6 +127,16 @@ def test_function_attributes(self):
             """declare noalias i32 @"my_func"(i32 zeroext %".1", i32 %".2", double %".3", i32* nonnull %".4")"""
             )
 
+    def test_function_metadata(self):
+        # Now with parameter attributes
```
pitrou: This comments looks out of place.

llvmlite/ir/_utils.py
whitequark: What's the point of a docstring on a private class? It's not visible unless you're looking at the code and there isn't anything to add to the code anyway.

pitrou: As a documentation to ourselves, to know why this exists ;-)
As for the `set_metadata` method, I'm assuming it is a public API?

llvmlite/tests/test_ir.py
whitequark: Fixed the comment

llvmlite/ir/_utils.py
whitequark: Added docstring for `set_metadata`.

pitrou:
Thank you! Can you document `Function.set_metadata()` in `docs/source/ir/values.rst`?

whitequark:
Done

PR #162, opened 2016-04-02
Add support for debug information

whitequark:
This is pretty much enough to do emission of debug info of any complexity, e.g. see https://github.com/m-labs/artiq/blob/ce30045dd4b85d20bd9ac2aa72ca8f7481cc6a43/artiq/compiler/transforms/llvm_ir_generator.py#L42-L127. It is in principle possible to model individual DINodes within llvmlite, but doesn't really make a lot of sense and is prone to breakage after upgrading LLVM.

### DIFF
--- a/llvmlite/ir/module.py
+++ b/llvmlite/ir/module.py
@@ -31,6 +31,23 @@ class Module(object):
             md = self._metadatacache[key]
         return md
 
+    def add_debug_info(self, kind, operands, is_distinct=False):
+        """
+        Add debug information metadata to the module with the given
+        *operands* (a dict of values with string keys) or return
+        a previous equivalent metadata. A DIValue instance is returned,
+        it can then be associated to e.g. an instruction.
+        """
+        n = len(self.metadata)
+        operands = tuple(sorted(operands.items()))
+        key = (kind, operands, is_distinct)
+        if key not in self._metadatacache:
+            di = values.DIValue(self, is_distinct, kind, operands, name=str(n))
+            self._metadatacache[key] = di
+        else:
+            di = self._metadatacache[key]
+        return di
+
     def add_named_metadata(self, name):
         nmd = values.NamedMetaData(self)
         self.namedmetadata[name] = nmd
--- a/llvmlite/ir/values.py
+++ b/llvmlite/ir/values.py
@@ -298,14 +298,9 @@ class MDValue(NamedValue):
         self.operands = tuple(values)
         parent.metadata.append(self)
 
-    @property
-    def operand_count(self):
-        return len(self.operands)
-
     def descr(self, buf):
         operands = []
         for op in self.operands:
-            typestr = str(op.type)
             if isinstance(op.type, types.MetaData):
                 if isinstance(op, Constant) and op.constant == None:
                     operands.append("null")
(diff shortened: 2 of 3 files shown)


whitequark:
Tests will pass once #160 is merged.

llvmlite/tests/test_ir.py
```diff
@@ -234,6 +235,34 @@ def test_metadata_null(self):
         self.assertInText("!{ i32* null }", str(mod))
         self.assert_valid_ir(mod)
 
+    def test_debug_info(self):
+        def order(d):
+            pairs = [(k, d[k]) for k in d]
+            pairs.sort()
+            return OrderedDict(pairs)
```
whitequark: The most ergononic API for `add_debug_info` is accepting a dict, yet the hash randomization screws up tests, so this workaround is used.

llvmlite/ir/values.py
```diff
@@ -312,6 +311,64 @@ def __ne__(self, other):
     def __hash__(self):
         return hash(self.operands)
 
+class DIToken:
+    def __init__(self, value):
+        self.value = value
```
whitequark: The purpose of this class is to represent DWARF enumeration values that should appear bare in the debug information metadata.

llvmlite/ir/module.py
```diff
@@ -31,6 +31,17 @@ def add_metadata(self, operands):
             md = self._metadatacache[key]
         return md
 
+    def add_debug_info(self, kind, operands, is_distinct=False):
+        n = len(self.metadata)
+        operands = tuple((key, operands[key]) for key in operands)
```
pitrou: How about `sorted(operands.items())`? This will give a you a deterministic order (nicer for testing).

llvmlite/ir/module.py
```diff
@@ -31,6 +31,17 @@ def add_metadata(self, operands):
             md = self._metadatacache[key]
         return md
 
+    def add_debug_info(self, kind, operands, is_distinct=False):
```
pitrou: Can you add a docstring?

llvmlite/ir/values.py
pitrou: That would be nicer as a docstring than as a review comment ;-)

llvmlite/ir/values.py
```diff
@@ -312,6 +311,64 @@ def __ne__(self, other):
...
+
+class DIValue(NamedValue):
```
pitrou: Docstring here as well?

llvmlite/ir/values.py
```diff
@@ -312,6 +311,64 @@ def __ne__(self, other):
...
+
+    def __init__(self, parent, is_distinct, kind, operands, name):
+        super(DIValue, self).__init__(parent, types.MetaData(), name=name)
+        self.is_distinct = is_distinct
+        self.kind = kind
+        self.operands = tuple(operands)
+        parent.metadata.append(self)
+
+    @property
+    def operand_count(self):
```
pitrou: If `.operands` is a public property, then this doesn't really look necessary.

pitrou:
There is also an obsolete PR in #67. Is anything worse reusing there?

llvmlite/ir/module.py
whitequark: llvmlite is already slow enough to do unnecessary sorting for silly reasons.

llvmlite/ir/values.py
whitequark: OK. I copied this from `MDValue`, should it be removed from there too?

pitrou: Oh, yes, we can.

whitequark:
I'll add docstrings. Wasn't bothering with them before in case you rejected the entire design.

As for #67 I think it can be just closed. There will be a near complete rework of debug info metadata in 3.9 (mainly to make it less DWARF-specific) and spending any effort on doing any sort of higher level representation right now is wasteful.

pitrou:
Fair enough. We can close #67 once this PR is merged.

llvmlite/ir/module.py
pitrou: It depends how often this is called. Determinism is desireable in its own right...

whitequark: There is *a lot* of debug metadata emitted, especially if your frontend, like mine, discerns column numbers. One DILocation per a few instructions.

In any case llvmlite is already nondeterministic, since it uses dicts and hashing in many places (like for metadata itself), and there's something going on with types too that makes the `.number` suffixes change between runs. Another PR may address this, maybe even mine if that type naming issue annoys me enough.

pitrou: Ok, fair enough ;-)

llvmlite/tests/test_ir.py
pitrou: So you can really spell this `OrderedDict(sorted(pairs.items()))`. It's more concise without losing readability :-)

whitequark: Ah yeah. Didn't know about `items()`.

whitequark:
All comments addressed

llvmlite/ir/module.py
```diff
@@ -31,6 +31,23 @@ def add_metadata(self, operands):
...
+        """
+        Add debug information metadata to the module with the given
+        *operands* (a dict of values with string keys) or return
+        a previous equivalent metadata. A DIValue instance is returned,
+        it can then be associated to e.g. an instruction.
+        """
+        n = len(self.metadata)
+        operands = tuple((key, operands[key]) for key in operands)
```
pitrou: Since a dict in unordered, you may want to sort the operands if you want to make the cache more efficient (e.g. `operands = tuple(sorted(operands.items()))`.

pitrou:
Note the Travis-CI build failed for some reason: https://travis-ci.org/numba/llvmlite/jobs/138088255#L659-L686

We'll also want to add some new documentation for this. Whether that should in the Values page or in a new page remains to be decided. I think a new page would be better.

whitequark:
> Note the Travis-CI build failed for some reason: https://travis-ci.org/numba/llvmlite/jobs/138088255#L659-L686

As expected. This changeset requires #160.

> We'll also want to add some new documentation for this. Whether that should in the Values page or in a new page remains to be decided. I think a new page would be better.

I guess. I don't have time to write those docs.

pitrou:
> As expected. This changeset requires #160.

Sorry, I had completely forgotten this. Let's finish #160 first then :-)

whitequark:
Since #160 was merged, this can be revisited.

pitrou:
Right. Perhaps you want to address https://github.com/numba/llvmlite/pull/162/files#r67347695 ?

whitequark:
@pitrou Oops, I missed that. Fixed now.

llvmlite/tests/test_ir.py
```diff
@@ -234,6 +235,34 @@ def test_metadata_null(self):
         self.assertInText("!{ i32* null }", str(mod))
         self.assert_valid_ir(mod)
 
+    def test_debug_info(self):
+        def order(d):
```
pitrou: This shouldn't be necessary anymore, now that the operands are sorted in add_debug_info(), right?

whitequark: Fixed

pitrou:
Thank you!

PR #163, opened 2016-04-03
Allow emission of fast-math fcmp instructions

whitequark:
Currently the way to add fast-math flags to instructions in
llvmlite is to mutate Instruction.opname for every individual one
to append flags, e.g. 'add nsw nuw' or 'frem fast'. However,
before this commit CompareInstr used the class attribute
self.OPNAME, defeating this method.

### DIFF
--- a/docs/source/ir/builder.rst
+++ b/docs/source/ir/builder.rst
@@ -351,15 +351,19 @@ Comparisons
    Unsigned integer compare *lhs* with *rhs*.  *cmpop*, a string, can be one
    of ``<``, ``<=``, ``==``, ``!=``, ``>=``, ``>``.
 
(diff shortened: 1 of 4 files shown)


pitrou:
Do you want to add a test for the bug that's being fixed?

pitrou:
Hi @whitequark, are you still working on the set of PRs? It would be nice to get those improvements in :-)

whitequark:
I'll do that at some point.

whitequark:
I've reworked this properly.

pitrou:
Thank you! Can you update the docs in `docs/source/ir/builder.rst`?

whitequark:
Done

PR #169, opened 2016-04-29
Default initializer for global variable with non external linkage

sklam:
IR spec requires GV with non-external linkage (linkage other than external and extern_weak) to have a initializer.  Otherwise, `parse_assembly` will reject the IR as invalid.  This patch makes `undef` the default initializer for these cases.

In details: if the the `.initializer` is `None` and linkage is non-external, emit `undef` to the IR.

### DIFF
--- a/llvmlite/tests/test_binding.py
+++ b/llvmlite/tests/test_binding.py
@@ -12,7 +12,7 @@ import subprocess
 import sys
 import unittest
 
-from llvmlite import six
+from llvmlite import six, ir
 from llvmlite import binding as llvm
 from llvmlite.binding import ffi
 from . import TestCase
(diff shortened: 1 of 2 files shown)


stuartarchibald:
I've built this patch and tested it with the IR from the HSA toolchain that was failing, I can confirm this fixes it and produces the correct initializer (`undef`) in the case that identified the issue:
```
RuntimeError: Failed at nopython (nopython mode backend)
LLVM IR parsing error
<string>:1477:1: error: global variable reference must have pointer type
@"_hsapy_smem.1" = internal   addrspace(3) global [400 x float]
```
This patch now correctly produces:
```
@"_hsapy_smem.1" = internal   addrspace(3) global [400 x float] undef
```

With reference to the language reference, the rest of the patch seems sound on inspection. Thanks for your effort.

llvmlite/tests/test_binding.py
```diff
@@ -982,5 +982,47 @@ def test_libm(self):
         llvm.load_library_permanently(libm)
 
 
+class TestGlobalVariables(BaseTest):
+    def check_global_variable_linkage(self, linkage):
+        mod = ir.Module()
```
pitrou: Why are those tests going into the bindings test? IMHO it would be better to have them in the IR tests (and comparing the actual IR output as we do for other IR constructs).

sklam: The main reason is that it is not clear in the llvm langref spec which linkage requires an initializer.  I am relying on parse_assembly() to validate the IR.

pitrou: Thanks. Can you add a comment explaining the issue ?

PR #172, opened 2016-05-02
Fix codecov.yml

pitrou:


stevepeak:
The `changes` status is under

```
coverage:
  status:
    changes: false
```

Thanks!

PR #176, opened 2016-06-15
Port llvmlite to LLVM 3.8

pitrou:
With these changes, llvmlite builds fine and the test suite passes (on Linux).

The whole Numba CPU test suite passes (not the CUDA suite). I spotted one significant performance regression: `bench_centdiff.CenteredDifference.time_centered_difference_1d` becomes 2x slower.

I've uploaded llvmdev packages for linux64 and win64 to the anaconda.org `numba` channel. Other packages will have to be built by the Jenkins builders.


llvmlite/tests/test_binding.py
```diff
@@ -269,7 +269,7 @@ def test_set_option(self):
 
     def test_version(self):
         major, minor, patch = llvm.llvm_version_info
-        self.assertIn((major, minor), [(3, 6), (3, 7)])
+        self.assertIn((major, minor), [(3, 7), (3, 8)])
```
pitrou: Le 12/07/2016 19:50, Stan Seibert a écrit :
> 
> Shouldn't this always be 3.8?

Yes... originally I thought llvmlite could be compatible with
consecutive LLVM versions (3.6 and 3.7, at the time), but I gave up on
the idea.

PR #188, opened 2016-07-09
added isinstance str check in _get_reference()

BennetLeff:
I had some errors by accidentally passing a non string in to `_get_reference()`. This change just makes sure a string is being iterated on in the condition.


pitrou:
Thank you for proposing this patch. I think the check should be done in `_set_name()`, and a TypeError raised if the name isn't a string.

PR #190, opened 2016-07-13
Suppress optimization remarks

pitrou:
The default LLVM diagnostic handler always prints some optimization remarks
(see http://lists.llvm.org/pipermail/llvm-dev/2016-July/102252.html).
Work around it by setting our own diagnostic handler during optimization.

### DIFF
--- a/ffi/passmanagers.cpp
+++ b/ffi/passmanagers.cpp
@@ -15,7 +15,6 @@
 
 using namespace llvm;
 
-extern "C" {
 
 /* Helpers to override the diagnostic handler when running
  * optimizations.
@@ -56,8 +55,10 @@ UnsetOptimizationDiagnosticHandler(LLVMContext &Ctx, diag_handler_t OldHandler)
 
 
 /*
- * Exposed optimization API
+ * Exposed API
  */
+
+extern "C" {
 
 API_EXPORT(LLVMPassManagerRef)
 LLVMPY_CreatePassManager()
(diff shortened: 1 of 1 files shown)


ffi/passmanagers.cpp
```diff
@@ -7,6 +15,49 @@
...
+        // If an error, print the message and bail out (as the default
+        // handler does).
+        DiagnosticSeverity DS = DI.getSeverity();
+        if (DS == DS_Error) {
+            raw_ostream &out = errs();
+            DiagnosticPrinterRawOStream DP(out);
+            out << "LLVM error: ";
+            DI.print(DP);
+            out << "\n";
+            exit(1);
```
sklam: Can we signal the python side instead of killing the process here?  `LLVMPY_RunFunctionPassManager` returns something to indicate success/failure already.  I am just afraid that calling `exit()` here makes it harder to debug if things go wrong.

pitrou: `exit(1)` is what LLVM does by default. Note that `LLVMPY_RunFunctionPassManager` returns 1 if something was modified, 0 otherwise, but does not return any error status.

I'm not aware of optimizations that might produce errors, do we have examples of how to trigger such an error?

sklam: Oh... in that case, it is fine

sklam:
I also wonder if we should put the set-diagnostic-handler in LLVMPY_GetGlobalContext and any Context creation API (IIRC, we don't have any, yet).  So far, llvmlite always use the global-context.  We just have to set the handlers once.

PR #199, opened 2016-08-15
Update for changes in LLVM 3.9

whitequark:
RelocMode is now Optional<RelocMode>, where None has the same meaning
as Default used to.

Data layout is no longer ever optional, so there is no need in
explicitly adding it anywhere; those APIs were removed in LLVM.

### DIFF
--- a/ffi/build.py
+++ b/ffi/build.py
@@ -109,7 +109,7 @@ def main_posix(kind, library_ext):
 
     out = out.decode('latin1')
     print(out)
-    if not out.startswith('3.8.'):
+    if not out.startswith('3.9.'):
         msg = (
             "Building llvmlite requires LLVM 3.8.x. Be sure to "
             "set LLVM_CONFIG to the right executable path.\n"
--- a/ffi/targets.cpp
+++ b/ffi/targets.cpp
@@ -88,13 +88,6 @@ LLVMPY_CreateTargetData(const char *StringRep)
     return LLVMCreateTargetData(StringRep);
 }
 
-API_EXPORT(void)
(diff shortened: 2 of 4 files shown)


pitrou:
Interesting, are these the only changes you had to make, or is it a work in progress?

whitequark:
The only ones.

pitrou:
I see, that's surprisingly small. Normally we would wait for 3.9.1 before making the switch, perhaps we'll try to test earlier this time.

whitequark:
@pitrou 3.9.1 is released...

whitequark:
Nevermind the previous comment, I misread some things.

sklam:
I have just tested with 3.9.1.  The following test now fails:


```
ERROR: test_debug_info (llvmlite.tests.test_ir.TestIR)
  File "/home/sklam/dev/llvmlite/llvmlite/tests/test_ir.py", line 315, in test_debug_info
    self.assert_valid_ir(mod)
  File "/home/sklam/dev/llvmlite/llvmlite/tests/test_ir.py", line 52, in assert_valid_ir
    llvm.parse_assembly(str(mod))
  File "/home/sklam/dev/llvmlite/llvmlite/binding/module.py", line 22, in parse_assembly
<string>:10:123: error: invalid field 'subprograms'
!5 = distinct !DICompileUnit(file: !0, isOptimized: true, language: DW_LANG_Python, producer: "ARTIQ", runtimeVersion: 0, subprograms: !4)
                                                                                                                          ^
```

Very likely due to https://llvm.org/bugs/show_bug.cgi?id=27284

PR #205, opened 2016-09-13
binary intrinsic declarations

sn6uv:
Fixes binary intrinsic (e.g. `llvm.maxnum`) IR generation.

**Note**: I haven't tested e.g. `llvm.convert.from.fp16`.

### DIFF
--- a/llvmlite/ir/module.py
+++ b/llvmlite/ir/module.py
@@ -155,7 +155,13 @@ class Module(object):
             raise NotImplementedError("unknown intrinsic %r with %d types"
                                       % (intrinsic, len(tys)))
 
-        name = '.'.join([intrinsic] + [t.intrinsic_name for t in tys])
+        name = [intrinsic]
+        if intrinsic in ('llvm.memcpy', 'llvm.memset'):
+            name.extend((t.intrinsic_name for t in tys))
+        elif tys:
+            name.append(tys[0].intrinsic_name)
+        name = '.'.join(name)
+
         if name in self.globals:
             return self.globals[name]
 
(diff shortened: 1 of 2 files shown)


sn6uv:
Any word on this? Is anyone aware of a workaround to use in the meantime?

pitrou:
Oops, sorry. I hadn't noticed this PR.
The workaround is simply to construct the `Function` object yourself.

sn6uv:
I understand better what's going on now and my patch isn't quite right. The solution for me is just to avoid the `declare_intrinsic` function.

However, there's still something to fix here I feel:
```
fnty = ir.FunctionType(double_type, [double_type, double_type])
mod.declare_intrinsic('llvm.pow', [double_type, double_type], fnty)
```
crashes because the name mangling isn't correct:
```
    mod = compile_ir(engine, llvm_ir)
  File "/home/angus/Mathics/mathics/builtin/compile/compile.py", line 38, in compile_ir
    mod.verify()
  File "/home/angus/llvmlite/llvmlite/binding/module.py", line 95, in verify
    raise RuntimeError(str(outmsg))
RuntimeError: Intrinsic name not mangled correctly for type arguments! Should be: llvm.pow.f64
double (double, double)* @llvm.pow.f64.f64
```

Or is the intended way to call function just exactly what's needed to match the llvm intrinsic full name? I.e.:
 ```
mod.declare_intrinsic('llvm.pow', [double_type], fnty)
```
In this case I can just update the docstring for `declare_intrinsic`.

pitrou:
> is the intended way to call function just exactly what's needed to match the llvm intrinsic full name?

Right. For `llvm.pow`, it seems `fnty` needn't even be passed.

sn6uv:
Okay, thanks @pitrou. I understand the behaviour of `declare_intrinsic` but I still find it slightly counter-intuitive. Regardless, I'm closing this PR.

PR #206, opened 2016-09-13
Update index.rst

kerrywatson1:
Adding project descriptors to all docs.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -2,7 +2,8 @@
 llvmlite
 ========
 
-**A lightweight LLVM python binding for writing JIT compilers**
+A lightweight LLVM python binding for writing JIT compilers
+-----------------------------------------------------------
 
 .. toctree::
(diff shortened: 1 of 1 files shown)


kerrywatson1:
@sklam can you review and merge this one too?

docs/source/index.rst
```diff
@@ -1,6 +1,8 @@
 
-llvmlite -- a light-weight Python binding to LLVM
-=================================================
+llvmlite
+========
+
+**A lightweight LLVM python binding for writing JIT compilers**
```
sklam: It is not limited to writing JIT compilers.  I would just keep it as "A lightweight LLVM python binding".

sklam: Nvm, just realize that description is already used in the README.  So, it's fine.

Perhaps, this https://github.com/numba/numba/pull/2123#issuecomment-246792878 applies here as well.

Issue #208, opened 2016-09-23
test_binding.py regular expression overly restrictive


toddrme2178:
I am trying to package llvmlite, and was running into a unit test error:

```
======================================================================
FAIL: test_linux (llvmlite.tests.test_binding.TestDependencies)
----------------------------------------------------------------------
Traceback (most recent call last):
    File "/home/abuild/rpmbuild/BUILDROOT/python3-llvmlite-0.13.0-0.x86_64/usr/lib64/python3.5/site-packages/llvmlite/tests/test_binding.py", line 171, in test_linux
        self.fail("invalid NEEDED line: %r" % (line,))
AssertionError: invalid NEEDED line: '  NEEDED               libLLVMObjCARCOpts.so.3.8'

----------------------------------------------------------------------
```

I tracked the problem down to [this regular expression](https://github.com/numba/llvmlite/blob/master/llvmlite/tests/test_binding.py#L163).  It matches libraries of the form `libname.so` or `libname.so.X` where `X` is an integer.  However, it is common on Linux to have `libname.so.X.Y` and `libname.so.X.Y.Z`, where `Y` and `Z` are also integer.  

So in this specific case, the test failed for `libLLVMObjCARCOpts.so.3.8`.  It would only accept something like `libLLVMObjCARCOpts.so.3`.  `libLLVMObjCARCOpts.so.3.8.1` would also be a valid library under Linux, but also would fail the test.

The fix is just to have the regular expression match multiple repetitions:  `lib_pat = re.compile(r'^([-_a-zA-Z0-9]+)\.so(?:\.\d+){0,3}$')`, or simply `lib_pat = re.compile(r'^([-_a-zA-Z0-9]+)\.so(?:\.\d+)*$')`

pitrou:
Well, `libLLVMObjCARCOpts` would still fail the test, no? It's not in the set of allowed dependencies.

PR #209, opened 2016-09-23
Allow TestDependencies to match other valid library file names

toddrme2178:
As mentioned in [issue #208](https://github.com/numba/llvmlite/issues/208), this unit tests forbids valid library file names, like `libname.so.3.8` and `libname.so.3.8.1`.  This modifies the regular expression to allowed those sorts of file names.


pitrou:
Sorry for the delay in reviewing this. I'm gonna merge your PR. Thank you for the fix!

PR #212, opened 2016-10-12
Add initializer for asm parser

sklam:
needed at https://github.com/numba/numba/pull/2149/files#diff-0faf9dcbe850c373d6769c3b1519a29aR194


pitrou:
The test failed on Travis-CI.

sklam:
ah, the asm format is different.   I should just test for the `nop`.

pitrou:
Should we add a doc for this? See http://llvmlite.pydata.org/en/latest/binding/initfini.html

sklam:
Yes, I will add doc

PR #213, opened 2016-10-14
Add partial LLVM bindings for object::ObjectFile

aguinet:
It misses the symbols API, but that's still a start!

### DIFF
--- a/llvmlite/binding/object_file.py
+++ b/llvmlite/binding/object_file.py
@@ -50,7 +50,7 @@ ffi.lib.LLVMPY_GetSectionName.argtypes = [ffi.LLVMSectionIteratorRef]
 ffi.lib.LLVMPY_GetSectionName.restype  = c_char_p
 
 ffi.lib.LLVMPY_GetSectionSize.argtypes = [ffi.LLVMSectionIteratorRef]
-ffi.lib.LLVMPY_GetSectionSize.restype  = c_ulonglong
+ffi.lib.LLVMPY_GetSectionSize.restype  = c_uint64
 
 ffi.lib.LLVMPY_GetSectionContents.argtypes = [ffi.LLVMSectionIteratorRef]
 ffi.lib.LLVMPY_GetSectionContents.restype  = c_char_p
--- a/llvmlite/binding/object_file.py
+++ b/llvmlite/binding/object_file.py
@@ -1,0 +1,59 @@
+from . import ffi
+from ctypes import c_bool, c_char_p, c_size_t, c_ulonglong, string_at
+
+class SectionIteratorRef(ffi.ObjectRef):
+    def name(self):
+        return ffi.lib.LLVMPY_GetSectionName(self)
+    def is_text(self):
+        return ffi.lib.LLVMPY_IsSectionText(self)
+    def size(self):
+        return ffi.lib.LLVMPY_GetSectionSize(self)
+    def data(self):
+        return string_at(ffi.lib.LLVMPY_GetSectionContents(self), self.size())
+    def is_end(self, object_file):
+        return ffi.lib.LLVMPY_IsSectionIteratorAtEnd(object_file, self)
+    def next(self):
+        ffi.lib.LLVMPY_MoveToNextSection(self)
+    def _dispose(self):
+        ffi.lib.LLVMPY_DisposeSectionIterator(self)
+
+class ObjectFileRef(ffi.ObjectRef):
+    @classmethod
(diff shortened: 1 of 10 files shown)


aguinet:
Currently adding test cases... Sorry about that!

llvmlite/binding/object_file.py
```diff
@@ -0,0 +1,59 @@
...
+        return ffi.lib.LLVMPY_GetSectionName(self)
+    def is_text(self):
+        return ffi.lib.LLVMPY_IsSectionText(self)
+    def size(self):
+        return ffi.lib.LLVMPY_GetSectionSize(self)
+    def data(self):
+        return string_at(ffi.lib.LLVMPY_GetSectionContents(self), self.size())
+    def is_end(self, object_file):
+        return ffi.lib.LLVMPY_IsSectionIteratorAtEnd(object_file, self)
+    def next(self):
```
sklam: This should use the iterator protocol (`__next__()`) and raise StopIteration when `is_end()` returns True.

aguinet: that's what is done below in the ObjectFileRef object. I did it this way to keep SectionIteratorRef close to the LLVM C API, and use ObjectFileRef as a wrapper for this.
Note that one issue with this original API is that previous iterators are invalid at each iterator, which is something that someone in python wouldn't expect.
For instance, if someone do `sections = list(ObjectFileRef().sections())`, it will get a list with only the last section.

llvmlite/binding/object_file.py
```diff
@@ -0,0 +1,59 @@
...
+ffi.lib.LLVMPY_MoveToNextSection.argtypes = [ffi.LLVMSectionIteratorRef]
+
+ffi.lib.LLVMPY_IsSectionIteratorAtEnd.argtypes = [ffi.LLVMObjectFileRef, ffi.LLVMSectionIteratorRef]
+ffi.lib.LLVMPY_IsSectionIteratorAtEnd.restype = c_bool
+
+ffi.lib.LLVMPY_GetSectionName.argtypes = [ffi.LLVMSectionIteratorRef]
+ffi.lib.LLVMPY_GetSectionName.restype  = c_char_p
+
+ffi.lib.LLVMPY_GetSectionSize.argtypes = [ffi.LLVMSectionIteratorRef]
+ffi.lib.LLVMPY_GetSectionSize.restype  = c_ulonglong
```
sklam: perhaps, this should use `c_uint64` to match the signature of LLVMPY_GetSectionSize even though `c_ulonglong` is equivalent in most platform.

sklam:
I am not familiar with the ObjectFile API.  What are the usecases enabled by this API?

llvmlite/binding/object_file.py
aguinet: Yes you're right ! I'll change this

aguinet:
I'm using it here : https://github.com/quarkslab/arybo/blob/feature/identasm/arybo/lib/exprs_asm.py#L184 (WIP code).
The overall idea is to get the assembly of a naked function described in LLVM IR. To do this, we product an object thanks to the LLVM Jitter (using llvmlite), and then we parse this object to extract the .text section and thus the wanted assembly.

llvmlite/binding/object_file.py
```diff
@@ -0,0 +1,59 @@
...
+ffi.lib.LLVMPY_MoveToNextSection.argtypes = [ffi.LLVMSectionIteratorRef]
+
+ffi.lib.LLVMPY_IsSectionIteratorAtEnd.argtypes = [ffi.LLVMObjectFileRef, ffi.LLVMSectionIteratorRef]
+ffi.lib.LLVMPY_IsSectionIteratorAtEnd.restype = c_bool
+
+ffi.lib.LLVMPY_GetSectionName.argtypes = [ffi.LLVMSectionIteratorRef]
+ffi.lib.LLVMPY_GetSectionName.restype  = c_char_p
+
+ffi.lib.LLVMPY_GetSectionSize.argtypes = [ffi.LLVMSectionIteratorRef]
+ffi.lib.LLVMPY_GetSectionSize.restype  = c_uint64
```
sklam: Test failed due to missing import for `c_uint64`

aguinet: ... sorry about that. fixed!

aguinet:
Is that okay to squash and merge or do I need to modify other things ?

sklam:
Thanks for the PR.

aguinet:
Thanks :)

PR #215, opened 2016-10-14
Add InlineAsm helpers in the builder

aguinet:
Add three helper functions in the builder:
- asm: create a call to an inline assembler object
- store_reg: store an LLVM value into a specific register
- load_reg: set an LLVM value as the value inside a register

The previous pull request was non-sense since there already is an InlineAsm instruction...

### DIFF
--- a/docs/source/ir/builder.rst
+++ b/docs/source/ir/builder.rst
@@ -550,6 +550,24 @@ Inline assembler
    Add an inline assembler call instruction. This is used for instance in
    :meth:`load_reg` and :meth:`store_reg`.
 
+   Arguments:
+     * *asm* is the inline assembler, i.e.: `̀ "mov $2, $0\nadd $1, $0"`` (x86
+       inline ASM uses the AT&T syntax)
+     * *constraint* defines the input/output constraints, i.e: ``=r,r,r``
+     * *args* is the list of arguments, as IR values, according to the contraits
+     * *side_effect* (boolean), whether this instruction has side effects not
+       visible in the constraint list
+     * *name* optional name of a possible LLVM value
+
+   For more information about these parameters, see the official LLVM
+   documentation here:
+   http://llvm.org/docs/LangRef.html#inline-asm-constraint-string.
+
+   Example that adds two 64-bit values on x86 would be::
+
+      fty = FunctionType(IntType(64), [IntType(64),IntType(64)])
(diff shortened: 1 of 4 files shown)


aguinet:
Any chance to have this merged ? I'm waiting to have this in master to make a new release of https://github.com/quarkslab/arybo/ . If you think the usage is too "specific" to land into master, I don't care adding the wrapper inside the project.
Thanks :)

seibert:
@pitrou: Can you take a look?

pitrou:
This looks fine on the principle, but it would be nice to add the new methods in the IRBuilder documentation (see http://llvmlite.pydata.org/en/latest/ir/builder.html).

Also, why are `load_reg` and `store_reg` limited to integer registers?

aguinet:
> This looks fine on the principle, but it would be nice to add the new methods in the IRBuilder documentation (see http://llvmlite.pydata.org/en/latest/ir/builder.html).

Okay, I'll add this today!

> Also, why are load_reg and store_reg limited to integer registers?

We can see registers as fixed-number of bits, so an integer type is the most basic representation of them. But indeed we can see them as "specific registers", like registers that are used only for floating point operations. So if that sounds fine for you we can change the API and replace "reg_size" by "reg_type" ?

Thanks!

pitrou:
According to http://llvm.org/releases/3.8.1/docs/LangRef.html#supported-constraint-code-list, some constraints can represent floating-point registers, so I assume a non-integer type is supported by the IR. So, yes, replacing reg_size by reg_type sounds more general.

aguinet:
I fixed these two issues in the new commit pushed (squashed the modification to get a clean history). Let me know if that's okay for you :) Thx!

docs/source/ir/builder.rst
```diff
@@ -542,6 +542,29 @@ Exception handling
    that the landing pad did not catch the exception after all (perhaps
    because it only performed cleanup).
 
+Inline assembler
+''''''''''''''''
+
+.. method:: IRBuilder.asm(ftype, asm, constraint, args, side_effect, name='')
+
+   Add an inline assembler call instruction. This is used for instance in
```
pitrou: Thanks for adding this! Can you also describe the arguments a bit? Or instead add an illustrative example.

aguinet: Fixed in the last commit! Sorry for the delay...

aguinet:
Is this possible to have this merged before the next release ? Let me know if some works still need to be done! Thanks!

pitrou:
Sorry for the delay! I'm gonna merge now.

PR #219, opened 2016-10-31
Remove unnecessary verify in module linkage

sklam:
We should let user run the verification instead of doing it in linkage.


pitrou:
Sounds fine to me.

PR #221, opened 2016-11-08
Add all function/return value attributes from LLVM 3.9

whitequark:
... except the Swift ones since those seem useless.

### DIFF
--- a/llvmlite/ir/values.py
+++ b/llvmlite/ir/values.py
@@ -666,6 +666,36 @@ class ArgumentAttributes(AttributeSet):
                         'nocapture', 'nonnull', 'returned', 'signext',
                         'sret', 'zeroext'])
 
+    def __init__(self):
+        self._dereferenceable = 0
+        self._dereferenceable_or_null = 0
+
+    @property
+    def dereferenceable(self):
+        return self._dereferenceable
+
+    @dereferenceable.setter
+    def dereferenceable(self, val):
+        assert isinstance(val, six.integer_types) and val >= 0
+        self._dereferenceable = val
+
+    @property
+    def dereferenceable_or_null(self):
+        return self._dereferenceable_or_null
+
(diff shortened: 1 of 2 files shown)


llvmlite/ir/values.py
```diff
@@ -649,6 +649,48 @@ class ArgumentAttributes(AttributeSet):
...
+        self._dereferenceable = 0
+        self._dereferenceable_or_null = 0
+
+    @property
+    def align(self):
+        return self._align
+
+    @align.setter
+    def align(self, val):
+        assert isinstance(val, int) and val >= 0
```
sklam: Can you replace `int` to `six.integer_types` so that it works with python2 long?  (See https://github.com/numba/llvmlite/pull/236/files#diff-daa10273fc8d59dae221475ac5306b12R409 for reference)

llvmlite/ir/values.py
```diff
@@ -649,6 +649,48 @@ class ArgumentAttributes(AttributeSet):
...
+        self._align = val
+
+    @property
+    def dereferenceable(self):
+        return self._dereferenceable
+
+    @dereferenceable.setter
+    def dereferenceable(self, val):
+        assert isinstance(val, int) and val >= 0
```
sklam: Same as above.

llvmlite/ir/values.py
```diff
@@ -649,6 +649,48 @@ class ArgumentAttributes(AttributeSet):
...
+        self._dereferenceable = val
+
+    @property
+    def dereferenceable_or_null(self):
+        return self._dereferenceable_or_null
+
+    @dereferenceable_or_null.setter
+    def dereferenceable_or_null(self, val):
+        assert isinstance(val, int) and val >= 0
```
sklam: Same as above.

whitequark:
Updated.

PR #226, opened 2016-12-28
Fix get_host_cpu_features() so that it fails properly

sklam:


seibert:
If we raise an exception, won't this cause problems using Numba with targets that do not have CPU features?

seibert:
Can we return an empty FeatureMap instead?

sklam:
@seibert 

> Can we return an empty FeatureMap instead?

With that, one can't tell if that is an error or there are actually no cpu features.

sklam:
> If we raise an exception, won't this cause problems using Numba with targets that do not have CPU features?

We can easily change numba to handle that case. 

Looking at the [target feature matrix](http://llvm.org/docs/CodeGenerator.html#target-feature-matrix), I am afraid that numba jit will not work on many platforms.

seibert:
OK, makes sense.  There are a few of the failing platforms where JIT should work, so this change will help a little.

Issue #194, opened 2016-07-27
Python 2.7 enum34 dependancy


davidkallie:
Creating a port for FreeBSD, it was brought up to my attention that for python 2.7, enum34 seems to be a necessary dependency for installing llvmlite, not running post installation. It seems that llvmlite.__init__.py tries to `import enum`, however this is happening during the install. 

It seems that it should be mention that setup.py requires enum34 during the install. I mention this because in the ports system, `install_requires` tells us to use `RUN_DEPENDS` in the build system. In numba passing `RUN_DEPENDS` works as expected for the port I am working on. However for llvmlite, it complains that enum34 is missing for us to compile. It is necessary for us to use `BUILD_DEPENDS`, but it is misleading.

The conversation in question is at [PR 211346](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=211346#add_comment) from Comment 11-13

Sorry for the semantics.

pitrou:
Thanks for the report. Can you submit a PR for the required change?

PR #231, opened 2017-01-10
Fix setup.py for Python2.7

NathanDeMaria:
`python setup.py build` in a `Python2.7` environment without `enum34` installed fails

Steps to reproduce:
```
conda create -n llvm_27 python=2.7 -y
source activate llvm_27
python setup.py build
```

Error:
```
  File "setup.py", line 24, in <module>
    from llvmlite.utils import get_library_files
  File "...", line 6, in <module>
    raise ImportError("could not find the 'enum' module; please install "
ImportError: could not find the 'enum' module; please install it using e.g. 'pip install enum34'
```

Moving the `import`s inside the `distutils.command`s they're used in makes sure the import doesn't happen until after the packages in `install_requires` have been installed.

Fixes https://github.com/numba/llvmlite/issues/194

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,8 +21,6 @@ if os.environ.get('READTHEDOCS', None) == 'True':
     sys.exit("setup.py disabled on readthedocs: called with %s"
              % (sys.argv,))
 
-from llvmlite.utils import get_library_files
-
 import versioneer
 
 versioneer.VCS = 'git'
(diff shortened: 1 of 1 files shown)


sklam:
Thanks!

To expand on the description, triggering the auto installation of the dependencies requires a `pip install .`.  Just running `python setup.py build` will trigger the same ImportError.

PR #243, opened 2017-02-16
Add __hash__ to all IR types

eamartin:
All of the ir.Types are not hashable in Python 3 because they define their own ``__eq__`` methods. It would be nice for types to be hashable so they could be used in sets or as keys to a dict. This fix is also necessary for correctness under Python 2:
```python
In [12]: a, b = ir.VoidType(), ir.VoidType()

In [13]: a == b
Out[13]: True

In [14]: hash(a) == hash(b)
Out[14]: False
```

Overriding ``__hash__`` on the ``Type`` class doesn't work because
> A class that overrides __eq__() and does not define __hash__() will have its __hash__() implicitly set to None. When the __hash__() method of a class is None, instances of the class will raise an appropriate TypeError when a program attempts to retrieve their hash value, and will also be correctly identified as unhashable when checking isinstance(obj, collections.Hashable).

(from the [``__hash__ docs``](https://docs.python.org/3/reference/datamodel.html#object.__hash__)).

There were a couple possible definitions for the hash functions. I initially considering ``hash(self.__class__)``, but subclasses would make this hash inconsistent with equality. The hashes could be improved to take into account more fields (such as the width of IntType). For the classes with instance caches (IntType and the float types), the hash could just be the address/``id`` of the singleton instance. I chose this implementation for it's simplicity and uniformity over all types.

### DIFF
--- a/llvmlite/ir/types.py
+++ b/llvmlite/ir/types.py
@@ -95,9 +95,6 @@ class MetaDataType(Type):
     def __eq__(self, other):
         return isinstance(other, MetaDataType)
 
-    def __ne__(self, other):
-        return not isinstance(other, MetaDataType)
-
 
 class LabelType(Type):
     """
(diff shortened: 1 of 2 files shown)


sklam:
Looks good.  Thanks for the patch.

PR #245, opened 2017-02-22
Add ORC-MCJIT replacement binding

sklam:
Adds binding to OrcMCJITReplacement.

Progress:

* [x] write binding
* [x] write tests
* [ ] write docs

Note:

The Orc-MCJIT is not a direct replacement for MCJIT for all cases.  When testing in numba, the lazy compilation in Orc-MCJIT seems to be non-threadsafe and it is failing any parallel execution of the same function.  (See related LLVM issue https://bugs.llvm.org//show_bug.cgi?id=5184).  In Orc-MCJIT, it seems the lazy compilation cannot be turned off.

### DIFF
--- a/llvmlite/tests/test_binding.py
+++ b/llvmlite/tests/test_binding.py
@@ -692,6 +692,71 @@ class TestMCJit(BaseTest, JITWithTMTestMixin):
         return llvm.create_mcjit_compiler(mod, target_machine)
 
 
+class TestOrcMCJit(BaseTest, JITWithTMTestMixin):
+    """
+    Test JIT engines created with create_orc_mcjit_compiler().
+    """
+
+    def jit(self, mod, target_machine=None):
+        if target_machine is None:
+            target_machine = self.target_machine()
+        return llvm.create_orc_mcjit_compiler(mod, target_machine)
+
+    def test_object_cache_getbuffer(self):   # overrides
+        # The ORC JIT needs a different test because it will fail at the
+        # second phase if loading function "sum" in a module without one.
+        # Instead, we will supply another "sum" implementation and make sure it
+        # is not used.
+        notifies = []
+        getbuffers = []
+
+        def notify(mod, buf):
+            notifies.append((mod, buf))
+
+        def getbuffer(mod):
+            getbuffers.append(mod)
+
+        mod = self.module()
+        ee = self.jit(mod)
+        ee.set_object_cache(notify, getbuffer)
+
+        # First return None from getbuffer(): the object is compiled normally
+        self.assertEqual(len(notifies), 0)
+        self.assertEqual(len(getbuffers), 0)
+        cfunc = self.get_sum(ee)
+        self.assertEqual(len(notifies), 1)
+        self.assertEqual(len(getbuffers), 1)
+        self.assertIs(getbuffers[0], mod)
+        sum_buffer = notifies[0][1]
+
+        # Recreate a new EE, and use getbuffer() to return the previously
+        # compiled object.
+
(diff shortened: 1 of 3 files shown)


seibert:
Does this mean we'll need to write our own MCJIT replacement with the ORC API that is actually thread-safe?

sklam:
Yes.  I think there are two potential approaches:

1. Make a eager-version.  The problem comes from The ORC-MCJIT is lazy.  It puts a stub function that defers the compilation to 1st call.  The stub is not threadsafe though.  Making it eager should remove the problem.

2. Make a threadsafe compile-on-demand layer ourselves.

seibert:
Do we benefit from the lazy compilation, since we are already deferring our call to LLVM to the first function call anyway?  If no benefit, then option 1 seems to make the most sense.

seibert:
And I assume that option 1 means Numba (or llvmlite?) will still need to enforce locking at the Python level to prevent concurrent compilation.

sklam:
The current numba setup won't benefit from lazy compilation.  I prefer option 1 and hope that it actually fixes the issue.  Option 1 will be exactly like what MCJIT is doing---eager compilation all the time.  Numba already has the concurrent compilation lock.

llvmlite/tests/test_binding.py
```diff
@@ -692,6 +692,71 @@ def jit(self, mod, target_machine=None):
...
+
+    def jit(self, mod, target_machine=None):
+        if target_machine is None:
+            target_machine = self.target_machine()
+        return llvm.create_orc_mcjit_compiler(mod, target_machine)
+
+    def test_object_cache_getbuffer(self):   # overrides
+        # The ORC JIT needs a different test because it will fail at the
+        # second phase if loading function "sum" in a module without one.
+        # Instead, we will supply another "sum" implementation and make sure it
```
pitrou: Wow. Is that a documented difference in LLVM? A bug?

sklam: Is the behavior in the MCJIT-version documented? 

It seems OrcJIT is stricter here.  Due to its laziness, Orc first tries to find a IR module that has the symbol.  Once found, it checks if it is already compiled and do its thing.  If not found, it stops and return NULL.

alendit:
What's the status of this? AFAIK ORC supports eager compilation via the basic `IRCompileLayer`. Maybe the `OrcMCJITReplacement` isn't the best way forward? I could add ORC as a separate ExecutionEngine with a superset of `MCJIT`'s python API. This way the numba wouldn't be directly affected and the new EE could be thoroughly tested before any switch is made.

abebeos:
> Yes. I think there are two potential approaches:
> 
>     1. Make a eager-version.  The problem comes from ORC-MCJIT being lazy.  It puts a stub function that defers the compilation to 1st call.  The stub is not threadsafe though.  Making it eager should remove the problem.

@lhames, @sklam 

Is this problem (and/or the overall PR) still actual?
 
>     2. Make a threadsafe compile-on-demand layer ourselves.

lhames:
@abebeos OrcMCJITReplacement was intended as an OrcV1 drop-in replacement for MCJIT. has been deprecated for a while and was removed with the rest of OrcV1 in LLVM 12. I think this PR can be closed.

abebeos:
> @abebeos OrcMCJITReplacement was intended as an OrcV1 drop-in replacement for MCJIT. has been deprecated for a while and was removed with the rest of OrcV1 in LLVM 12. I think this PR can be closed.

οκ.

@esc, this one is closable.

esc:
sure

PR #246, opened 2017-02-22
add requirements.txt for pip dependency resolving

stephengroat:
allows pip to resolve dependencies without setuptools


sklam:
Can you add a python version requirement?  Python version < 3.4 doesn't need enum34.

Thanks!

sklam:
Looks good!  Thanks for the patch!

PR #250, opened 2017-03-02
Disable static linking of libstdc++ by default.

### LINKED ISSUE #249: Option to enable/disable static-linking during build
From https://github.com/numba/llvmlite/issues/93#issuecomment-283378934

> If no-static-linking is the default, it will fit most users that are building from source. Only those that needs to distribute and be compatible with older linux needs to turn on static linking.

sklam:
(Fix #249)

Only enable it for distribution build in the conda recipe.
User can still override with LLVMLITE_CXX_STATIC_LINK.

### DIFF
--- a/conda-recipes/llvmlite/build.sh
+++ b/conda-recipes/llvmlite/build.sh
@@ -7,6 +7,8 @@ if [ -n "$MACOSX_DEPLOYMENT_TARGET" ]; then
 fi
 
 export PYTHONNOUSERSITE=1
+# Enables static linking of stdlibc++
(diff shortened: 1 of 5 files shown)


stephengroat:
I know this was just merged, but I was wondering if/when the v17-dev tage would be updated and if there were any plans for the v17 release

seibert:
We're planning to do the version 0.17 release in the next week along with the update to LLVM 4.0

stephengroat:
thanks so much for the information! super helpful!

PR #253, opened 2017-03-24
Build and test with PyPy master Linux with Travis CI

gmarkall:
This adds support for building and testing llvmlite with PyPy on Linux, as a prerequisite for building and testing Numba on PyPy. There are a few comments I'd make about this PR:

- It relies on a fairly recent build of PyPy, that I've packaged: https://anaconda.org/gmarkall/pypy - this is because I hit a bug in PyPy that was causing a few tests to fail: https://bitbucket.org/pypy/pypy/issues/2475/ctypes-restype-appears-to-have-no-effect - this is now fixed in PyPy.
- The PyPy package is in my personal Anaconda channel at the moment. Perhaps it should be copied to the Numba channel before merge?
- As more PyPy fixes and support are added to Numba, I'm expecting to find and fix more bugs in PyPy. This will require updating of the PyPy conda package, so I think it's worth having a PyPy conda package (instead of just waiting for a PyPy release and using that on Travis).
- The pypy conda package is called `pypy` and not `python`. I considered calling it `python` and giving it a version number that indicated that it was PyPy, but I expect that to cause issues because I think there's a bit of special-casing in Conda for the `python` package, and I didn't want to end up in a rabbit-hole of conda-related issues. If additional conda packages to go with this PyPy package are required later, then I can build them (I think this would be easier than trying to get the `pypy` conda package to work with the rest of the conda package ecosystem).
- Support for PyPy is only tested on Linux at the moment, as it's the platform I use most frequently. I'm hoping to get everything working on Linux first, before worrying about other platforms (and I hope that it will be a relatively small task to add support for testing on Windows/OS X once it's working on Linux).
- There's a couple of tests added for `PassManagerBuilder` functions, which I'd noticed were untested.
- As well as the required modifications to the Travis setup, I added `python --version` to `test.sh` so that it's clear which Python is being used (just in case somehow a wrong Python binary got used when testing with PyPy, e.g. the system Python - this is unlikely, but I wanted it to be unambiguous).

Once this PR is merged, I think the next step for me to look at is to submit a PR to Numba that enables building with PyPy (though no testing, but maybe running `import numba`) as an incremental move towards PyPy support and testing.

Any comments and feedback appreciated - I am still in the Numba team on anaconda.org, so I can copy the package to the Numba channel and update this PR accordingly if required.


gmarkall:
I've updated the comment - I believe this PR is now ready for review.

PR #268, opened 2017-05-22
Support packed struct

### LINKED ISSUE #267: Support packed struct type
It seems that llvmlite doesn't support packed attribute for struct types.

sklam:
Fixes #267.

A small enhancement to support packed struct.


stuartarchibald:
This looks good, thanks for adding it.

PR #272, opened 2017-05-25
cmpop contains comparison type, not lhs

squeaky-pl:


sklam:
Thanks for the patch!

PR #273, opened 2017-05-31
Fix the behavior of module.get_global

choleraehyq:
Fix the behavior of module.get_global to keep consistency with the document.
Document says: 

> Get the global value (a GlobalValue) with the given name. KeyError is raised if it doesn’t exist.

But actually `dict.get` will never raise an `KeyError`. `OrderedDict` is a subclass of `dict`.

See:
https://docs.python.org/2/library/stdtypes.html#dict.get
https://docs.python.org/3/library/stdtypes.html#dict.get

### DIFF
--- a/llvmlite/ir/module.py
+++ b/llvmlite/ir/module.py
@@ -135,7 +135,7 @@ class Module(object):
         """
         Get a global value by name.
         """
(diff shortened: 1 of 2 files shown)


sklam:
Thanks for the patch!

PR #279, opened 2017-06-23
Fix #274 addrspace in gep

### LINKED ISSUE #274: GEPInstr in llvmlite/ir/instructions.py:456  drops the addrspace
I was running `text_complex_kernel` in a personal Numba OpenCL backend and LLVM would complain about the following lines because of the missing `addrspace(4)` in %.80:
```
  %.79 = getelementptr inbounds { double, double }, { double, double } addrspace(4)* %arg.a.4, i64 %.75, i32 1
  %.80 = load double, double* %.79, align 8
```
Tracing the issue I reached instructions.py:456 and noticed that `typ = typ.as_pointer()`was dropping the addrspace, changed it to `typ = typ.as_pointer(ptr.type.addrspace)`and now `text_complex_kernel` generates the following LLVM code and executes correctly.
```
  %.80 = load double, double addrspace(4)* %.79, align 8
```
This problem did not arise in the CUDA backend, probably because the generic address space is 0, therefore `addrspace(0)` is not carried around and it would still be ok to drop it.

I am not totally sure this is an issue, but it is worth reporting.
Thanks.

sklam:
Fixes both constant and non-constant gep

### DIFF
--- a/llvmlite/ir/instructions.py
+++ b/llvmlite/ir/instructions.py
@@ -453,7 +453,7 @@ class GEPInstr(Instruction):
                 isinstance(lasttyp, types.PointerType)):
             typ = lasttyp
         else:
-            typ = typ.as_pointer()
+            typ = typ.as_pointer(ptr.addrspace)
 
         super(GEPInstr, self).__init__(parent, typ, "getelementptr",
                                        [ptr] + list(indices), name=name)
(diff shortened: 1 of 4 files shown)


stuartarchibald:
Thanks. This looks good, will merge on CI pass.

PR #307, opened 2017-10-18
Fixes for LLVM5

sklam:
Based on #306 

This updates llvmlite for the changes in LLVM 5:

* Fix C-API get attribute usage.
* Remove the deprecated get_pointer_to_function since it behavior is unexpected.

### DIFF
--- a/llvmlite/binding/value.py
+++ b/llvmlite/binding/value.py
@@ -1,4 +1,3 @@
-import warnings
 from ctypes import POINTER, c_char_p, c_int, c_size_t, c_uint
 import enum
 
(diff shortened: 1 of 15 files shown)


llvmlite/binding/value.py
```diff
@@ -1,5 +1,5 @@
-
-from ctypes import POINTER, c_char_p, c_int
+import warnings
```
stuartarchibald: Is this used?

sklam: Good catch.  It's now fixed.

stuartarchibald:
Thanks for the fixup... merging.

jakirkham:
Any idea as to when this might land in a release?

seibert:
We are hoping to take a version 0.36 release candidate next week, with the final release a week later.

jakirkham:
Thanks @seibert. Could you please give me a ping when those land?

PR #313, opened 2017-11-08
[rfc] LLVM Vector Type and Operations Support

ekiwi:
This is not polished enough to be merged, however I would like to know if anyone would be interested in support for LLVM vector types.

https://llvm.org/docs/LangRef.html#vector-type
https://llvm.org/docs/LangRef.html#vector-operations

I'd also appreciate feedback on how to improve the code to more naturally integrate with `llvmlite`.

Is there anyone who wants to test this under `python2`?

If there is enough interest to merge this, I will take the time to write some unittests.

### DIFF
--- a/llvmlite/ir/types.py
+++ b/llvmlite/ir/types.py
@@ -332,6 +332,54 @@ class _Repeat(object):
         else:
             raise IndexError(item)
 
+class VectorType(Type):
+    """
+    The type for vectors of primitive data items (e.g. "<f32 x 4>").
+    """
+
+    def __init__(self, elementtype, count):
+        self.elementtype = elementtype
+        self.count = count
+
+    @property
+    def elements(self):
+        return _Repeat(self.elementtype, self.count)
+
+    def __len__(self):
+        return self.count
+
+    def _to_string(self):
+        return "<%d x %s>" % (self.count, self.elementtype)
+
+    def __eq__(self, other):
+        if isinstance(other, VectorType):
+            return self.elementtype == other.elementtype and self.count == other.count
+
+    def __hash__(self):
+        # TODO: why does this not take self.elementtype/self.count into account?
+        return hash(VectorType)
+
+    def format_constant(self, value):
+        itemstring = ", " .join(["{0} {1}".format(x.type, x.get_reference())
+                                 for x in value])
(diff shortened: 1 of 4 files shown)


sklam:
@ekiwi, Thanks.  The PR looks good so far.  I wrote some tests as I was looking through it.  I uploaded my patch to https://gist.github.com/sklam/76f213a2b4e3ef17dfcf0af82286545a .

llvmlite/ir/values.py
```diff
@@ -116,9 +116,7 @@ def __init__(self, typ, constant):
         assert isinstance(typ, types.Type)
         assert not isinstance(typ, types.VoidType)
         self.type = typ
-        if isinstance(constant, (list, tuple)):
-            # Recursively wrap aggregate constants
-            constant = typ.wrap_constant_value(constant)
+        constant = typ.wrap_constant_value(constant)
```
sklam: Good catch

llvmlite/ir/types.py
```diff
@@ -332,6 +332,54 @@ def __getitem__(self, item):
         else:
             raise IndexError(item)
 
+class VectorType(Type):
+    """
+    The type for vectors of primitive data items (e.g. "<f32 x 4>").
```
sklam: should be <4 x f32>

ekiwi:
Thanks for the feedback @sklam .

I am going to write some unittests over the weekend and push them to this branch.

Once everything is in shape I will squash and rebase on master.

cod3monk:
@ekiwi do you plan to continue on this PR? I would be very interested in it making it into the master branch.

ekiwi:
> @ekiwi do you plan to continue on this PR? I would be very interested in it making it into the master branch.

I would like to and if people want this feature that is definitely motivating.
Research is keeping me super busy right now, but I might be able to find a free weekend during the summer.

hassanalinali:
It's been more than a year, can we review this?

seibert:
I think it would be ideal to compare / combine this with #322, which has as similar goal.

hassanalinali:
@seibert What does the other one has in comparison? They look almost the same, but @ekiwi also included support for vector operations. Can't we just merge this one and leave the other one? It might take more time to combine the two than just fix this one.

PR #316, opened 2017-11-27
Remove temporary code to make cross compilation work.

stuartarchibald:
This removes temporary code needed to make linux64->linux32 cross
compilation work, the fix is now upstream so this can be removed.


stuartarchibald:
This needs a run through the build farm before merge.

PR #318, opened 2017-11-28
Ensuring threadsafety in concurrent usage of LLVM C-API

sklam:
~~~(don't merge this until 0.21.0 is released.)~~~

Various race-condition issues had been revealed in the numba testsuite and hacks were made to temporarily workaround the issues.  This patch provides a proper solution to the threadsafety problem by employing a lock on all LLVM C-API calls.  

LLVM uses a LLVMContext to store shared data.  By isolating LLVMContext and disallowing sharing of LLVMContext among threads, threadsafety issues could be avoided.  However, the current llvmlite API does not expose LLVMContext creation and the APIs assume a single global context.  So, we must ensure that the LLVM APIs will not be concurrently used.  By using a global lock for all LLVM API, it can keep the llvmlite API simple and safe.  Later when we discover certain LLVM APIs to be threadsafe, we can disable the lock on them individually.

### DIFF
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -1,5 +1,6 @@
 import ctypes
 import os
+import threading
 
 from .common import _decode_string, _is_shutting_down
 from ..utils import get_library_name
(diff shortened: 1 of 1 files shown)


llvmlite/binding/ffi.py
```diff
@@ -30,6 +31,82 @@ def _make_opaque_ref(name):
 LLVMSectionIteratorRef = _make_opaque_ref("LLVMSectionIterator")
 
 
+class _lib_wrapper(object):
+    """Wrap the libllvmlite with a lock such that only one thread is
+    accessing it at a time.
```
stuartarchibald: Perhaps `Wrap libllvmlite with a lock such that only one thread may access it at a time`?

llvmlite/binding/ffi.py
```diff
@@ -30,6 +31,82 @@ def _make_opaque_ref(name):
...
+
+    This class duck-types a CDLL.
+    """
+    __slots__ = ['_lib', '_fntab', '_lock']
+
+    def __init__(self, lib):
+        self._lib = lib
+        self._fntab = {}
+        # The recursive lock is needed for callbacks that re-enter
```
stuartarchibald: s/recursive/reentrant/?

llvmlite/binding/ffi.py
```diff
@@ -30,6 +31,82 @@ def _make_opaque_ref(name):
...
+    def _handle(self):
+        """The system handle used to access the library.
+
+        For duck-typing a ctypes.CDLL
+        """
+        return self._lib._handle
+
+
+class _lib_fn_wrapper(object):
+    """Wraps and duck-type a ctypes.CFUNCTYPE to provide
```
stuartarchibald: s/duck-type/duck-types/

seibert:
Merging so we can let buildbot grind on this over the weekend.

PR #322, opened 2018-01-22
Adding Vector Type

komalon:


SabirAkhadov:
Is this getting merged soon? We could make use of the vector type for implementing correct struct ABI for Cfunc.

seibert:
@sklam: We also have the much older PR #313.  Not sure if some of that should be pulled forward into this PR.

llvmlite/ir/types.py
```diff
@@ -196,6 +196,39 @@ def __eq__(self, other):
     def __hash__(self):
         return hash(FunctionType)
 
+class VectorType(Type):
+
+    @classmethod
+    def __new__(cls, count, typ):
```
sklam: There's no need to use `__new__` here.  This should be a `__init__` instead.    Also, a `__new__` doesn't need `@classmethod`.

llvmlite/ir/types.py
```diff
@@ -196,6 +196,39 @@ def __eq__(self, other):
...
+        assert isinstance(count, int) and count >= 0
+        assert isinstance(typ, Type)
+        self = super(VectorType, cls).__new__(cls)
+        self.count = count
+        self.type = typ
+        return self
+
+    def __getnewargs__(self):
```
sklam: Don't need this if `__new__` is not defined.

PR #326, opened 2018-02-14
Enable Intel SVML-enabled auto-vectorization for all the transcendentals

anton-malakhov:
This patch allows to enable SVML in LLVM correctly on all the platforms and provides a way to choose faster functions by loosing some accuracy via fastmath=True argument in Numba

Credit to @dvnagorny for the LLVM patch


anton-malakhov:
@seibert Any update on this? Unlike changes in Numba, the changes here are opt-in and will not hurt anything else.

seibert:
We're sorting out one remaining Numba test failure related to the LLVM 6 transition so we can merge those two PRs first.

stuartarchibald:
I've built llvmdev, then llvmlite and finally Numba against this patch and the current test suites for the Numba/llvmlite pass.

seibert:
In that case, I vote we merge this PR while we keep working on the LLVM 6 issues.

stuartarchibald:
Agreed. Thanks for the patch. Merging.

PR #328, opened 2018-02-16
Build C files separately on Linux and support parallel make

mgorny:
Change the Linux Makefile to build each source file separately instead of using one big `$(CXX)` invocation and support using parallel make. This is cleaner, and makes it possible to utilize parallel make on multi-core CPUs as well as distcc.

### DIFF
--- a/ffi/Makefile.linux
+++ b/ffi/Makefile.linux
@@ -5,21 +5,27 @@ CXX ?= g++
 CXX_FLTO_FLAGS ?= -flto
 LD_FLTO_FLAGS ?= -flto -Wl,--exclude-libs=ALL
 
-CXXFLAGS = $(LLVM_CXXFLAGS) $(CXX_FLTO_FLAGS)
+# -fPIC is required when compiling objects for a shared library
+CXX_FPIC_FLAGS ?= -fPIC
+
+CXXFLAGS = $(LLVM_CXXFLAGS) $(CXX_FLTO_FLAGS) $(CXX_FPIC_FLAGS)
 LDFLAGS = $(LLVM_LDFLAGS) $(LD_FLTO_FLAGS)
 LIBS = $(LLVM_LIBS)
 INCLUDE = core.h
-SRC = assembly.cpp bitcode.cpp core.cpp initfini.cpp module.cpp value.cpp \
-	  executionengine.cpp transforms.cpp passmanagers.cpp targets.cpp dylib.cpp \
-	  linker.cpp object_file.cpp
+OBJ = assembly.o bitcode.o core.o initfini.o module.o value.o \
+	  executionengine.o transforms.o passmanagers.o targets.o dylib.o \
+	  linker.o object_file.o
 OUTPUT = libllvmlite.so
 
 all: $(OUTPUT)
 
-$(OUTPUT): $(SRC) $(INCLUDE)
+.cpp.o: $(INCLUDE)
+	$(CXX) -c $(CXXFLAGS) $< -o $@
+
+$(OUTPUT): $(OBJ)
 	# static-libstdc++ avoids runtime dependencies on a
 	# particular libstdc++ version.
-	$(CXX) $(CXX_STATIC_LINK) -shared $(CXXFLAGS) $(SRC) -o $(OUTPUT) $(LDFLAGS) $(LIBS)
+	$(CXX) $(CXX_STATIC_LINK) -shared $(CXXFLAGS) $(OBJ) -o $(OUTPUT) $(LDFLAGS) $(LIBS)
 
 clean:
-	rm -rf test $(OUTPUT)
+	rm -rf test $(OUTPUT) $(OBJ)
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.pyc
 *.pyd
+*.o
 *.so
 *.dylib
 *.dll
--- a/ffi/build.py
+++ b/ffi/build.py
@@ -120,7 +120,8 @@ def main_posix(kind, library_ext):
         os.environ['CXX_STATIC_LINK'] = "-static-libstdc++"
 
     makefile = "Makefile.%s" % (kind,)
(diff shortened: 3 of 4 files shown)


sklam:
I have resolved the merge conflict on this PR.  But I notice that the separate compilation will increase the shared library size by 7MB in exchange for faster compilation time.  @seibert, do you  think it's worth it?

seibert:
I'm puzzled by the size change.  Is there some LTO flag we need to turn on to get the same library size?

Also, the size change is 7MB out of how much?  IIRC, 7MB would actually be a non-trivial fraction of the final shared library size.

mgorny:
Please try stripping. I can't really check that since we're using only shared libraries in Gentoo, so the result is tiny (93k).

sklam:
It's growing from 44MB to 51MB.

epsilon-0:
hey @stuartarchibald 
can we take a look at this PR again?
OpenBSD also has some trouble getting llvmlite to build right now

gmarkall:
@mgorny Many thanks for the PR and your efforts so far!

I'm spending some time going through the llvmlite PR backlog - this PR seems to be one that's been stuck for a while. Looking over past activity, I think the following actions are needed if this is to move forward:

- Merge master and resolve conflicts.
- Checking of the resulting library size by the author and making appropriate changes to remedy the increase if it appears to be excessive (e.g. perhaps using LTO if needed as suggested by @seibert above). Note that we recently added more targets, so the absolute / relative sizes of the library might differ from those previously mentioned above.

A couple of questions for @mgorny:
- Do you feel that there's likely to be further progress on this PR? If not, please let us know and / or close the PR.
   - If further progress is unlikely then if any other interested parties are keen to take over the shepherding of this PR to completion (possibly @epsilon-0?), I think a new PR based on the work of this one could be opened.
- If you'd still like to see progress, what do you think about my two suggestions above?

mgorny:
Well, I don't mind rebasing it and implementing specific suggestions given by others. However, please don't expect me to put too much effort into it — I've already given up on llvmlite and removed it entirely from Gentoo.

gmarkall:
@mgorny Many thanks for your quick response! If you're able to merge master and resolve conflicts, then just make sure it builds as you intended, then I will have a look at the potential issues around the resulting binary size.

mgorny:
I've rebased the commits against `master`, and compile-tested it.

gmarkall:
Many thanks! I'll try this out tomorrow and report back.

gmarkall:
Build times on my machine - 6-core 12-thread Xeon Gold 6128 w/ 96GB RAM (`real`, measured with `time python setup.py develop`):

* `split-c-build`: 13 seconds
* `master`: 24 seconds

`libllvmlite.so` sizes:

* `split-c-build`: 115266440 bytes (110MB)
* `master`: 115266440 bytes (110MB)

This appears not to affect the size of libllvmlite.so when building in this way.

mgorny:
Thanks. Have you been building with LTO enabled?

gmarkall:
> Thanks. Have you been building with LTO enabled?

Yes, I believe so - `-flto` appears in my compile and link flags - e.g.:

```
g++ -c   -fPIC -I/home/gmarkall/miniconda3/envs/numbanp120/include -std=c++14 -fno-exceptions -fno-rtti -D_GNU_SOURCE -D_DEBUG -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -fno-rtti -g -DHAVE_SVML -flto -fPIC assembly.cpp -o assembly.o
```

and

```
g++  -shared   -fPIC -I/home/gmarkall/miniconda3/envs/numbanp120/include -std=c++14 -fno-exceptions -fno-rtti -D_GNU_SOURCE -D_DEBUG -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -fno-rtti -g -DHAVE_SVML -flto -fPIC assembly.o # ...
```

ffi/Makefile.linux
```diff
@@ -4,22 +4,27 @@ CXX ?= g++
...
-	  linker.cpp object_file.cpp custom_passes.cpp
+OBJ = assembly.o bitcode.o core.o initfini.o module.o value.o \
+	  executionengine.o transforms.o passmanagers.o targets.o dylib.o \
+	  linker.o object_file.o custom_passes.o
 OUTPUT = libllvmlite.so
 
 all: $(OUTPUT)
 
-$(OUTPUT): $(SRC) $(INCLUDE)
+.cpp.o: $(INCLUDE)
```
gmarkall, line 21: What are these lines doing? If I comment them out, I still seem to get the exact same invocation of `g++` for compiling, and the same `.so` output. (My makefile expertise is limited - is there some subtlety I'm missing here?

mgorny: They define how to create an `.o` file from `.cpp`.

You get the same result most likely because GNU make has an implicit rule for building C++ files. However, POSIX does not specify such a rule (for C++, it does for plain C), so other make implementations may be missing one.

gmarkall: Many thanks for the explanation!

gmarkall (approved):
This looks good to me.

I'm going to leave this as "4 - Waiting for reviewer" just to check with @sklam - do you still see a size increase with this PR?

gmarkall:
I've moved this to 0.39RC as it seems that this could go in by then (assuming @sklam is happy with it too).

sklam:
I got the same libllvmlite.so size for `master` and for this branch.

gmarkall:
> I got the same libllvmlite.so size for `master` and for this branch.

OK - I think this is OK to merge then.

esc:
@mgorny thank you for the patch! @gmarkall thank you for the review!

gmarkall:
@mgorny Many thanks for the patch and for your continued efforts on it towards merging :slightly_smiling_face:

mgorny:
Thanks! Feel free to ping me in case of any regressions.

PR #330, opened 2018-03-01
Add support for LLVM fence instruction

obilaniu:
The `fence` instructions are useful for shared-memory scenarios on weak-memory-model machines like PowerPC and ARM.

More specifically, it is necessary to use fences to render safe Numba's reference count increment/decrement.

We add here an `llvmlite.ir.instructions.Fence` instruction and a method `IRBuilder.fence(targetscope, ordering, name='')`.

The legal values of `targetscope` are `None`, `"singlethread"` and other target-specific supported values.

The legal values of `ordering` are `"acquire"`, `"release"`, `"acq_rel"` and `"seq_cst"`.

### DIFF
--- a/llvmlite/ir/instructions.py
+++ b/llvmlite/ir/instructions.py
@@ -676,3 +676,34 @@ class LandingPadInstr(Instruction):
                               clauses=''.join(["\n      {0}".format(clause)
                                                for clause in self.clauses]),
                               ))
+
+class Fence(Instruction):
+    """
+    The `fence` instruction.
+
+    As of LLVM 5.0.1:
+
+    fence [syncscope("<target-scope>")] <ordering>  ; yields void
+    """
+
+    VALID_FENCE_ORDERINGS = {"acquire", "release", "acq_rel", "seq_cst"}
+
+    def __init__(self, parent, ordering, targetscope=None, name=''):
+        super(Fence, self).__init__(parent, types.VoidType(), "fence", (), name=name)
+        if ordering not in self.VALID_FENCE_ORDERINGS:
+            raise ValueError("Invalid fence ordering \"{0}\"! Should be one of {1}."
+                             .format(ordering, ", ".join(self.VALID_FENCE_ORDERINGS)))
+        self.ordering = ordering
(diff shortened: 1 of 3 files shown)


llvmlite/ir/instructions.py
```diff
@@ -676,3 +676,28 @@ def descr(self, buf):
                               clauses=''.join(["\n      {0}".format(clause)
                                                for clause in self.clauses]),
                               ))
+
+class Fence(Instruction):
+    """
+    The `fence` instruction.
+
+    As of LLVM 5.0.1:
```
stuartarchibald: Thanks for the patch. Will this be valid under LLVM 6.0 too?

obilaniu: The syntax for `fence` changed from [LLVM 4](https://releases.llvm.org/4.0.0/docs/LangRef.html#fence-instruction) to [LLVM 5](https://releases.llvm.org/5.0.0/docs/LangRef.html#fence-instruction), which is why I mentioned it, but since then there have been no changes in [LLVM 6](https://releases.llvm.org/6.0.0/docs/LangRef.html#fence-instruction) nor in [LLVM trunk](https://www.llvm.org/docs/LangRef.html#fence-instruction).

stuartarchibald: Ah great, thanks for confirming.

obilaniu:
@stuartarchibald Now that mainline llvmlite supports LLVM 6, may I call for this PR to be merged, with a view to the eventual merger also of Numba [PR 2780](https://github.com/numba/numba/pull/2780) and thereby a partial solution to Numba issue [2451](https://github.com/numba/numba/issues/2451)?

stuartarchibald:
Yes. Please could you add a test to exercise the code path to demonstrate that the added functionality behaves as expected? (code does not need to be compiled, just needs to ensure that what is generated is valid). Thanks.

obilaniu:
@stuartarchibald Done.

stuartarchibald:
@obilaniu thanks for adding the test, this looks good. Merging.

PR #337, opened 2018-03-13
[WIP] Add support for optimization remarks in pass managers.

sklam:
This enables optimization passes to write diagnostic remarks to a YAML file.  The remarks contains info about applied optimizations and missed opportunities with references to the source file if source location (debug info) is present.

The added feature is localized to pass managers.  The `.run()` method of module and function pass-managers takes a new optional argument `remarks_file` for the filepath where the remark info will be stored.  The feature is disabled if `remarks_file` is `None` (the default).

Related read: https://llvm.org/devmtg/2016-11/Slides/Nemet-Compiler-assistedPerformanceAnalysis.pdf

### DIFF
--- a/ffi/passmanagers.cpp
+++ b/ffi/passmanagers.cpp
@@ -6,12 +6,14 @@
 #include "llvm/IR/DiagnosticInfo.h"
 #include "llvm/IR/DiagnosticPrinter.h"
 #include "llvm/Support/raw_ostream.h"
-
 #include "llvm-c/Transforms/Scalar.h"
 #include "llvm-c/Transforms/IPO.h"
 #include "llvm/IR/LegacyPassManager.h"
 #include "llvm/Transforms/Scalar.h"
 #include "llvm/Transforms/IPO.h"
+#include "llvm/Support/ToolOutputFile.h"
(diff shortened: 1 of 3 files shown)


whitequark:
Why a file and not a string?

sklam:
> Why a file and not a string?

Right, I agree.  It's better to have a string.  The current implementation is based on some code I found in LLVM.  I need to see if I can make the opt-remarks print to a string instead.

alendit:
What's the progress on this? I've built something similar which works with the current version, should I open a PR?

gmarkall:
I'm spending some time going through the llvmlite PR backlog - this PR seems to be one that's been stuck for a while.

@sklam Do you think that you're likely to pick this up again?

@alendit If you still have something similar which works with the current version, opening a PR to replace this one would be great - it may well be easier for us to make further progress if you have something already working and ready-to-go (as I know that @sklam has many other Numba and llvmlite-related tasks at hand that push this down the priority list).

sklam:
sorry, i won't have time to finish this up

stuartarchibald:
Closing as #837 implemented the feature based on this PR. Thanks all!

PR #338, opened 2018-03-13
Add 3 Bit Manipulation Intrinsics

njwhite:
https://llvm.org/docs/LangRef.html#llvm-ctpop-intrinsic etc


sklam:
Thanks for your contributions!

PR #339, opened 2018-03-19
Add cttz & ctlz

njwhite:
Some more bit manipulation intrinsics for numba/numba#2824

CC @sklam


sklam:
Patch looks good.  Thanks for your contribution!

PR #340, opened 2018-03-20
This moves llvmlite to use LLVM 6.0.0 as its backend.

stuartarchibald:
Changes include:
 * Updates to documentation
 * Updates to conda recipes and wheel builds
 * Inclusion and application of necessary patches from LLVM 7.0svn
   (001-Transforms-Add-missing-header-for-InstructionCombini.patch) to
   allow some LLVM-C symbols to correctly export.
 * Removal of the ELF relocation patch (6.0 source appears to handle this
   correctly now).
 * Rewrite of the LLVMPY_LinkModules internals with respect to the new
   DiagnosticHandler classes and API.
 * Removal of the code that switched out the diagnostic handler in the
   Pass Manager (6.0 source suggests that the root cause is likely to be
   handled correctly).
 * Update the target machine code model handling as `CodeModel::JITDefault`
   and `CodeModel::Default` were both removed. If unspecified the new default
   is `CodeModel::Large`.
 * The `unit-at-a-time` operations on the PassManagerBuilder are removed in
   LLVM 6.0 and so are also removed in llvmlite.


stuartarchibald:
SHA changes from rebase on master following SVML patch.

stuartarchibald:
SHA changes from rebase on master following build number patch

ffi/targets.cpp
```diff
@@ -194,18 +194,24 @@ LLVMPY_CreateTargetMachine(LLVMTargetRef T,
...
         cm = CodeModel::Small;
     else if (cms == "kernel")
         cm = CodeModel::Kernel;
     else if (cms == "medium")
         cm = CodeModel::Medium;
     else if (cms == "large")
         cm = CodeModel::Large;
-    else
-        cm = CodeModel::Default;
+    else if (cms == "default") // As per LLVM 5, needed for AOT
```
sklam: `As per LLVM 5` ?  Should it be LLVM 6?

ffi/targets.cpp
```diff
@@ -194,18 +194,24 @@ LLVMPY_CreateTargetMachine(LLVMTargetRef T,
...
+        cm = CodeModel::Small;
+    else { // catches "jitdefault" and not set, as per LLVM 5, needed for MCJIT
```
sklam: Same here

ffi/targets.cpp
stuartarchibald: I think it should be `5` as `default` and `jitdefault` don't exist in 6, I added them back in as per the behaviour in 5.

ffi/targets.cpp
```diff
@@ -194,18 +194,24 @@ LLVMPY_CreateTargetMachine(LLVMTargetRef T,
...
     else if (cms == "kernel")
         cm = CodeModel::Kernel;
     else if (cms == "medium")
         cm = CodeModel::Medium;
     else if (cms == "large")
         cm = CodeModel::Large;
-    else
-        cm = CodeModel::Default;
+    else if (cms == "default") // As per LLVM 5, needed for AOT
+        cm = CodeModel::Small;
```
sklam: "default" -> `CodeModel::Small`?

This doesn't match the docstring:

```
+        Specifying codemodel='default' will result in a code model being
+        picked based on platform bitness (32="small", 64="large").
```

stuartarchibald:
Thanks for the review @sklam I think I've addressed your comments. I also switched the default code model in use back to `jitdefault` too as per the previous release (as I think this is a more common use).

sklam:
Great! Merging.

seibert:
The 0.38.0 release candidate was tagged on April 13, and we plan to tag and build the final release on April 23.

PR #343, opened 2018-04-02
Fix undefined behavior bug due to Twine usage in LLVM

sklam:
This adds a patch to LLVM source code to fix an undefined behavior bug.  A new test is added for the llvmdev package to verify the patch worked.  Note, the patched code will only affect branch weight printing at WriteCFG during CFG DOT file generation.

### DIFF
--- a/conda-recipes/llvmdev/cfg_test.ll
+++ b/conda-recipes/llvmdev/cfg_test.ll
@@ -1,0 +1,56 @@
+; ModuleID = 'foo'
+source_filename = "<string>"
+target datalayout = "e-m:o-i64:64-f80:128-n8:16:32:64-S128"
+target triple = "x86_64-apple-darwin17.4.0"
+
+@.const.foo = internal constant [4 x i8] c"foo\00"
+@".const.Fatal error: missing _dynfunc.Closure" = internal constant [38 x i8] c"Fatal error: missing _dynfunc.Closure\00"
+@PyExc_RuntimeError = external global i8
+@".const.missing Environment" = internal constant [20 x i8] c"missing Environment\00"
+
+; Function Attrs: norecurse nounwind
+declare i32 @"_ZN8__main__7foo$241Ex"(i64* noalias nocapture %retptr, { i8*, i32 }** noalias nocapture readnone %excinfo, i8* noalias nocapture readnone %env, i64 %arg.x) local_unnamed_addr #0
+
+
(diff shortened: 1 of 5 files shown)


sklam:
I have tested the patched llvm with numba and it passes.

conda-recipes/llvmdev/cfg_test.ll
```diff
@@ -0,0 +1,56 @@
+; ModuleID = 'foo'
+source_filename = "<string>"
+target datalayout = "e-m:o-i64:64-f80:128-n8:16:32:64-S128"
+target triple = "x86_64-apple-darwin17.4.0"
```
stuartarchibald: Does it matter that the triple is wrong for linux but the test only guards against windows https://github.com/numba/llvmlite/pull/343/files#diff-a5674b1fbcdc1e5882531629a69766cbR71?

seibert:
Should we merge this and build new llvmdev packages, or do you want to go hunting for other incorrect usage of Twine?

stuartarchibald:
RE: https://github.com/numba/llvmlite/pull/343#issuecomment-378667637 I'm inclined to get this in and see how much corruption it fixes, then go after the one in `CodeGen`, so we know the impact of each patch?

conda-recipes/llvmdev/cfg_test.ll
sklam: It doesn't matter as we don't go thru codegen.  We only ask `opt` to make a graph of the CFG.  I have run this on Linux.  In fact, this patch runs fine on our buildfarm.

sklam:
Yes, I want to see the impact of each patch.  The problem in `CodeGen` is limited because it is Darwin specific and requires globals to be merged.  

See https://github.com/llvm-mirror/llvm/blob/master/lib/CodeGen/GlobalMerge.cpp#L486-L489

Also, there are ways to turn off that globalmerge pass.

conda-recipes/llvmdev/cfg_test.ll
stuartarchibald: Ok, thanks.

stuartarchibald:
I'm happy with this, let's merge and rebuild the stack.

sklam:
A bug is filed to LLVM bugtracker: https://bugs.llvm.org/show_bug.cgi?id=37019

jakirkham:
Did the LLVM developers get in touch with you about this bug?

PR #351, opened 2018-04-29
Python 3.7 compat: Properly escape repl in re.sub

mingwandroid:
With `python -c "import re; print(re.sub(r'\s+',r'\s*','      here are  some   words'))"`

Python 2.7 and 3.6 give:
```
\s*here\s*are\s*some\s*words
```

While Python 3.7 gives:
```
  File "/opt/conda/envs/py37/lib/python3.7/sre_parse.py", line 1021, in parse_template
    this = chr(ESCAPES[this][1])
KeyError: '\\s'

During handling of the above exception, another exception occurred:

  File "<string>", line 1, in <module>
  File "/opt/conda/envs/py37/lib/python3.7/re.py", line 192, in sub
    return _compile(pattern, flags).sub(repl, string, count)
  File "/opt/conda/envs/py37/lib/python3.7/re.py", line 309, in _subx
    template = _compile_repl(template, pattern)
  File "/opt/conda/envs/py37/lib/python3.7/re.py", line 300, in _compile_repl
    return sre_parse.parse_template(repl, pattern)
  File "/opt/conda/envs/py37/lib/python3.7/sre_parse.py", line 1024, in parse_template
    raise s.error('bad escape %s' % this, len(this))
re.error: bad escape \s at position 0
```

From https://docs.python.org/3.7/whatsnew/3.7.html
.. this change in behaviour seems to be due to:
```
Unknown escapes consisting of '\' and an ASCII letter in replacement templates for
re.sub() were deprecated in Python 3.5, and will now cause an error.
```

The URL for this bug is:
https://bugs.python.org/issue27030

And the commit is:
https://github.com/python/cpython/commit/9bd85b83f66d31e39282244e455275bd9cd91bb1

### DIFF
--- a/llvmlite/tests/test_ir.py
+++ b/llvmlite/tests/test_ir.py
@@ -46,7 +46,7 @@ class TestBase(TestCase):
             return c
 
         pattern = ''.join(map(escape, pattern))
-        regex = re.sub(r'\s+', r'\s*', pattern)
+        regex = re.sub(r'\s+', r'\\\s*', pattern)
         self.assertRegexpMatches(text, regex)
 
     def assert_ir_line(self, line, mod):
--- a/llvmlite/tests/test_ir.py
+++ b/llvmlite/tests/test_ir.py
@@ -46,7 +46,7 @@ class TestBase(TestCase):
             return c
 
         pattern = ''.join(map(escape, pattern))
-        regex = re.sub(r'\s+', r'\\\s*', pattern)
+        regex = re.sub(r'\s+', r'\\s*', pattern)
         self.assertRegexpMatches(text, regex)
 
     def assert_ir_line(self, line, mod):


mingwandroid:
@sklam, @seibert, @stuartarchibald please review.

mingwandroid:
Hmm, while it seems to be fine in this isolated test, I'm still encountering the same failure during `llvmlite`'s tests.

stuartarchibald:
I think the reason this patch was failing is that the replacement should either be:
```
regex = re.sub(r'\s+', r'\\s*', pattern)
```
or
```
regex = re.sub(r'\s+', '\\\s*', pattern)
```

stuartarchibald:
Windows test is failing because llvmdev still has that dep on vs2008 from py2.7 which means it always ends up downgrading all the packages to py2.7. https://github.com/numba/llvmlite/pull/353 and a stack rebuild should fix.

stuartarchibald:
Thanks for testing @mingwandroid . I've just force pushed to get a rebuild with the new llvmdev for LLVM 6 (build 3) on windows that should have the right dependency chain that doesn't cause a Python version downgrade.

stuartarchibald:
I pushed this through the Numba build farm. The 32bit OS builders failed as there's no shipped Py 3.7 packages available on the `c3i_test` channel yet, however the rest of the builds went fine.

stuartarchibald:
Agreed. Post 0.23.1 tag this can go in.

stuartarchibald:
Thanks for the reviews. Merging.

PR #353, opened 2018-05-08
Fix llvmdev build recipe.

stuartarchibald:
This adds a selector such that windows does not use Anaconda
defined compilers which have a dependency on vc9.


nehaljwani:
If the selectors don't work for whatever reason, you can try this:
```
{% set on_win = SUBDIR in ('win-64', 'win-32') %}
...
    {% if not on_win %}
     - {{ compiler('c') }}
     - {{ compiler('cxx') }}
    {% endif %}
```

PR #356, opened 2018-05-16
fix build on OpenBSD.

sigalpes:
Hi,

this pull request fixes build on OpenBSD.

Best regards.


stuartarchibald:
Many thanks for the patch. I was wondering if OpenBSD and FreeBSD are sufficiently similar that in fact they could just use the same Makefile, this reducing maintenance burden etc? Then here https://github.com/numba/llvmlite/pull/356/files#diff-464cc6be0bee1e15453cc1f2d3686d17R136 could just be something like `sys.platform in ('openbsd', 'freebsd')` or, if we know enough about BSDs to be confident, just see if 'bsd' is in the `sys.platform` string?

sigalpes:
Hi,
thank you for your response.
We could use the same Makefile than FreeBSD, I think it's a good idea.
I would prefer to use `sys.platform in ('openbsd', 'freebsd')` instead of 'bsd' is in the sys.platform string.
Thank you.

stuartarchibald:
@sigalpes I'm happy with your chosen preference. Would you like to make this change in the current PR? Thanks.

stuartarchibald:
@sigalpes looks good. Thanks. Merging.

PR #361, opened 2018-05-31
Add SVML detection and a function to declare support.

stuartarchibald:
This adds detection of SVML support in a LLVM build at compile
time. This is then exposed such that it can be queried at run time
from Python in a manner that is suitable for use in the SVML
initialisation and detection sequence.

Closes #3006
Fixes #2998

### DIFF
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1,8 +1,9 @@
 v0.23.2
 -------
 
-This is a bug fix release to assist in addressing a critical bug in Numba:
-https://github.com/numba/numba/issues/3006
+This is a bug fix release to assist in addressing a critical Numba issue that
+can affect users who download llvmlite packages from sources other than PyPI
+(pip), Anaconda, or Intel Python: https://github.com/numba/numba/issues/3006
 
 Support for SVML is now detected at compile time and baked into a function that
 is exposed by llvmlite. This function can be queried at runtime to find out if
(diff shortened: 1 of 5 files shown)


stuartarchibald:
This is a first attempt, I had to somewhat guess as there's 3 platforms and also now 2 or 3 LLVM build variants.

stuartarchibald:
To manually test this I've used a windows 64 bit machine with a Miniconda3, and on it created two conda packages of `llvmdev`, one as normal, and one in which I've deliberately removed the SVML patch. I've then built `llvmlite` against each of them in turn and checked that the `llvmlite.binding.has_svml` function reports correctly.

stuartarchibald:
I've also run this patch through the build farm, it passed everywhere.

seibert:
Can you also add the Changelog update for 0.23.2 to this PR?

stuartarchibald:
@seibert change log done in https://github.com/numba/llvmlite/pull/361/commits/86f8206cca22294d7cc69d7b3beed6da51210956

CHANGE_LOG
```diff
@@ -1,3 +1,19 @@
+v0.23.2
+-------
+
+This is a bug fix release to assist in addressing a critical bug in Numba:
```
seibert: "This is a bug fix release to assist in addressing a critical Numba issue that can affect users who download llvmlite packages from sources other than PyPI (pip), Anaconda, or Intel Python:"

Just to make it clear that this doesn't affect a large number of people.

stuartarchibald: good idea, thanks, done.

PR #363, opened 2018-06-09
Workaround for incorrect vectorization factor computed for SVML functions

anton-malakhov:
Resolves https://github.com/numba/numba/issues/3016


anton-malakhov:
@seibert @stuartarchibald @sklam This patch fixed the tests in my builds. The CI here does not build llvmdev, so the results here are not telling anything.

conda-recipes/D47188-svml.patch
```diff
@@ -560,30 +533,41 @@ index 37ffac1..8ad2131 100644
...
-@@ -4925,7 +4926,10 @@ void InnerLoopVectorizer::widenInstruction(Instruction &I) {
+@@ -4925,7 +4932,8 @@ void InnerLoopVectorizer::widenInstruction(Instruction &I) {
          VectorF = Intrinsic::getDeclaration(M, ID, TysForDecl);
        } else {
          // Use vector version of the library call.
 -        StringRef VFnName = TLI->getVectorizedFunction(FnName, VF);
-+        // TEMPORARY. For backporting (down to release 5.0.0) purposes let's use any() from FastMathFlags instead of 
-+        // still absent isFast(). It isn't equivalent for full mask but I think it's enough there
-+        bool IsFast = CI->getFastMathFlags().any();
++        bool IsFast = CI->getFastMathFlags().isFast();
```
anton-malakhov: hmm.. I havn't noticed that before, is it what you implemented @stuartarchibald ? Did you need it for LLVM 6.0.0? I havn't noticed any issue with `.isFast()`

anton-malakhov: I guess it is up to the difference between earlier version and published version of the patch

anton-malakhov:
I can reproduce the issue on windows, so this patch is not enough there

conda-recipes/D47188-svml.patch
stuartarchibald: Think it was the latter, I don't recall writing that :)

conda-recipes/D47188-svml.patch
```diff
@@ -560,30 +533,46 @@ index 37ffac1..8ad2131 100644
...
-@@ -4917,6 +4917,7 @@ void InnerLoopVectorizer::widenInstruction(Instruction &I) {
+@@ -3974,6 +3974,17 @@ static unsigned getVectorCallCost(CallInst *CI, unsigned VF,
+   if (!TLI || !TLI->isFunctionVectorizable(FnName, VF) || CI->isNoBuiltin())
+     return Cost;
+ 
++  // this goes against LLVM coding philosophy, but it'll stop bleeding
++  bool IgnoreMe;
++  StringRef FuncName = TLI->getVectorizedFunction(FnName, VF, IgnoreMe, true);
++#if LLVM_ON_WIN32
++  StringRef SvmlPrefix("\0_svml", 6); // nobody knows why symbols are like this
```
stuartarchibald: Should `SvmlPrefix` be used in the comparison in the condition below i.e. `FuncName.startswith(SvmlPrefix)` ?

conda-recipes/llvmdev_manylinux1/meta.yaml
```diff
@@ -1,7 +1,7 @@
 {% set shortversion = "6.0" %}
 {% set version = "6.0.0" %}
 {% set sha256 = "1ff53c915b4e761ef400b803f07261ade637b0c269d99569f18040f3dcee4408" %}
-{% set build_number = "1" %}
+{% set build_number = "4" %}
```
stuartarchibald: I think this needs to be "2", the wheel builds move at a different pace to conda builds.

conda-recipes/D47188-svml.patch
```diff
@@ -154,23 +122,28 @@ index 0000000..f68c086
...
-+def round      : SvmlVariant;
 +def sind       : SvmlVariant;
 +def sinh       : SvmlVariant;
 +def sqrt       : SvmlVariant;
 +def tan        : SvmlVariant;
 +def tanh       : SvmlVariant;
-+def trunc      : SvmlVariant;
++
++// While SVML doesn't provide _ha versions of
++// the following symbols let's disable their vectorization.
```
stuartarchibald: Is this conveying the following, "SVML does not provide _ha version of the following symbols so prevent them from being vectorized"?

conda-recipes/D47188-svml.patch
```diff
@@ -1,36 +1,5 @@
...
- test/Transforms/LoopVectorize/X86/svml-calls.ll |  40 +++++++++
- utils/TableGen/CMakeLists.txt                   |   1 +
- utils/TableGen/SVMLEmitter.cpp                  | 114 ++++++++++++++++++++++++
- utils/TableGen/TableGen.cpp                     |   8 +-
- utils/TableGen/TableGenBackends.h               |   1 +
- utils/vim/syntax/llvm.vim                       |   1 +
- 22 files changed, 366 insertions(+), 99 deletions(-)
- create mode 100644 include/llvm/IR/SVML.td
- create mode 100644 utils/TableGen/SVMLEmitter.cpp
+From https://reviews.llvm.org/D47188 rebased on top of LLVM 6.0.0
```
stuartarchibald: It would be nice if this remained as a git formatted patch.

conda-recipes/D47188-svml.patch
anton-malakhov: ah, sharp eye! thanks! it's bad copy-paste. Looks like the llvmdev test (in conda-forge) doesn't work as intended as well since it didn't catch it

conda-recipes/D47188-svml.patch
anton-malakhov: yes, in order to speedup publishing the patch upstream, we decided to do it in smaller steps.

conda-recipes/D47188-svml.patch
anton-malakhov: I'm not sure how to add these comments in git-formatted patch. Or do you suggest to drop them away? it does not look like it breaks anything, many other patches have arbitrary headers

conda-recipes/llvmdev_manylinux1/meta.yaml
anton-malakhov: will do, but it is inconsistent with another recipe above

stuartarchibald:
Thanks for the effort on this @anton-malakhov, build farm seems to be happy with it. Merging.

PR #365, opened 2018-06-11
Add printing to the bound llvm type

alendit:
The `LLVMPY_TypeRef` class is a completely opaque type right now. This adds a thin wrapper which prints the LLVM IR representation of the LLVM Type.

If this change is ok, I'll adjust the documentation, too.

I have one further PR prepared which allows extracting `TypeRef`s from a `ModuleRef`, similar to the `get_function` and `functions`. I'd like to wait for the feedback on this PR first, though.

My envisioned use case is making the interop with external modules easier. Let say I've compiled some C code to a `c.ll` file. I'd like to use functions and structs defined there from a module created with `llvmlite`. Right now I'd have to manually forward declare everything. With these changes I can pretty much automate everything.

### DIFF
--- a/llvmlite/tests/test_binding.py
+++ b/llvmlite/tests/test_binding.py
@@ -794,7 +794,23 @@ class TestValueRef(BaseTest):
         mod = self.module()
         glob = mod.get_global_variable("glob")
         tp = glob.type
-        self.assertIsInstance(tp, ffi.LLVMTypeRef)
+        self.assertIsInstance(tp, llvm.TypeRef)
+
+    def test_type_printing_variable(self):
+        mod = self.module()
+        glob = mod.get_global_variable("glob")
+        tp = glob.type
+        self.assertEqual(str(tp), 'i32*')
+
+    def test_type_printing_function(self):
+        mod = self.module()
+        fn = mod.get_function("sum")
+        self.assertEqual(str(fn.type), "i32 (i32, i32)*")
+
+    def test_type_printing_struct(self):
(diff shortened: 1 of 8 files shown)


llvmlite/tests/test_binding.py
```diff
@@ -491,11 +494,12 @@ def test_parse_bitcode_error(self):
         self.assertIn("Invalid bitcode signature", str(cm.exception))
 
     def test_bitcode_roundtrip(self):
-        bc = self.module().as_bitcode()
+        bc = self.module(asm=asm_mul).as_bitcode()
```
sklam: Why does this test has to be changed?

alendit: Since we reuse the context for the changes, the types get a numerical suffix each time a module is parsed (i.e. `struct.somestruct.1` etc). This fails the tests since the produced IR is slightly different, even though it's sematically equivalent. `asm_mul` doesn't declare any types, so it works fine.

Another solution would be to allow the user to instantiate custom llvm contexts instead of relying on the global one. Since it's the preferred method in the newer LLVM version, it might be the better solution in the mid-term, but would be outside of the scope of this PR.

alendit:
@sklam sorry, it totally slipped my mind to answer to your review. I'll update the docs in the coming days.

llvmlite/tests/test_binding.py
sklam: I see.  And yes, we need to move to the new way of llvm-context at some point.

Issue #350, opened 2018-04-27
(new) Warning at shutdown


dscole:
Getting the following warning at shutdown with llvmlite-0.23.0 and numba-0.38.0:

```
Exception TypeError: "'NoneType' object is not callable" in <bound method ModuleRef.__del__ of <llvmlite.binding.module.ModuleRef object at 0x7fdf4a8f46d0>>
```
Not sure how to track it down further.

asmeurer:
I get a similar warning, in Python 2 only

```
Exception TypeError: "'NoneType' object is not callable" in <bound method ExecutionEngine.__del__ of <llvmlite.binding.executionengine.ExecutionEngine object at 0x11d2b2390>> ignored
```

(you can reproduce this by running the [sympy](http://github.com/sympy/sympy) llvm tests, `./bin/test llvm`)

The `__del__` method is https://github.com/numba/llvmlite/blob/44a515a346d420ee999e329960d50f6768551dd5/llvmlite/binding/ffi.py#L241-L242

I'm not really clear on [`__del__`](https://docs.python.org/2/reference/datamodel.html#object.__del__) documentation if you're supposed to be able to fix this by making something start with an underscore. Or should it just be guarded with a check if it's `None`? 

I also don't know why it's Python 2-only. I guess the del algorithm changed in Python 3 so that `close` doesn't get deleted when `__del__` is called.

asmeurer:
Guarding against None removes the warnings for me. https://github.com/numba/llvmlite/pull/368

will133:
The guard against the close does not seem to work for me.  I dug into it a little bit and the problem at least for me has to do with the _dispose function in:

https://github.com/numba/llvmlite/blob/80d8f86ff81467a52658df95e58a4524bfbf5403/llvmlite/binding/module.py#L77-L78

The LLVMPY_DisposeModule function seems to raise the "NoneType is not callable exception" error internally during exit on python 2.7.  If i do the following:

```python
    def _dispose(self):
        try:
            self._capi.LLVMPY_DisposeModule(self)
        except TypeError as e:
            if 'object is not callable' in str(e):
                pass
            else:
                raise
```
I can get rid of the error.  Do you know if that's the right thing to do?  Or if there's something wrong going underneath the DisposeModule call?

This issue is also reported on the numba bug (https://github.com/numba/numba/issues/2818) where it can be reproduced by doing (from @stuz5000):

```python
import numba
import numpy as np


@numba.jit
def f(a):
[...]

PR #368, opened 2018-06-21
Guard against close being None in ObjectRef.__del__

asmeurer:
This can happen if it is called at interpreter shutdown.

Partial fix to #350 (I didn't try to fix any other `__del__` methods).


asmeurer:
This also seems to fix warnings like `Exception AttributeError: '_as_parameter_' in <bound method ModuleRef.__del__ of <llvmlite.binding.module.ModuleRef object at 0x116c1a550>> ignored`, though I'm not really clear why.

PR #369, opened 2018-06-22
Recipe fixes for Conda Build 3

nehaljwani:
Newer versions of Conda Build are more stricter about the build/host split.


stuartarchibald:
@nehaljwani thanks for the patch. Looks good and the build farm is happy with it. Merging.

PR #379, opened 2018-07-18
ARM aarch64 test on jetson tx2

stuartarchibald:


stuartarchibald:
A Py 3.6 build of Numba worked against this patch on the build farm.

PR #383, opened 2018-08-02
Add type getting

alendit:
This PR adds `get_type` method and `struct_types` iterator to the `ModuleRef` class, similar to `get_function` and `functions`. 

The purpose is to facilitate interaction with the IR/bitcode module created by other means than `llvmlite`. You this extension you can extract types from the existing modules and use them in the module created in `llvmlite` before you link in the other module. By automatically extracting the types we make sure that the types are identical and the linker recognizes them as the same.


seibert:
Thanks for this contribution!  I've tagged @sklam to review.

alendit:
@sklam Note: `get_struct_type` might be a better name than `get_type`. I'll leave the final call to you.

PR #389, opened 2018-08-23
Add symbols from static object files

alendit:
Hi,

this adds a new `add_object_file` method which takes either an `ObjectFileRef` instance or a string path and exposes the symbols defined in the static object file to the execution engine. 

Cheers.


alendit:
There were some spurious failures on OSX. Could someone with an access to a Mac check if `test_add_object_file_from_filesystem` runs properly?

llvmlite/tests/test_binding.py
```diff
@@ -1273,22 +1274,82 @@ def test_inlineasm(self):
...
+
+        jit = llvm.create_mcjit_compiler(self.module(self.mod_asm),
+            target_machine)
+
+        jit.add_object_file(obj)
+
+        sum_twice = CFUNCTYPE(c_int, c_int, c_int)(
+            jit.get_function_address("sum_twice"))
+
+        assert sum_twice(2, 3) == 10
```
sklam: Better use `self.assertEqual`.

llvmlite/tests/test_binding.py
```diff
@@ -1273,22 +1274,82 @@ def test_inlineasm(self):
...
+                target_machine)
+
+            jit.add_object_file(temp_path)
+        finally:
+            os.unlink(temp_path)
+
+        sum_twice = CFUNCTYPE(c_int, c_int, c_int)(
+            jit.get_function_address("sum_twice"))
+
+        assert sum_twice(2, 3) == 10
```
sklam: Better use `self.assertEqual`.

llvmlite/tests/test_binding.py
```diff
@@ -1273,22 +1274,82 @@ def test_inlineasm(self):
...
 class TestObjectFile(BaseTest):
+
+    mod_asm = """
+        ;ModuleID = <string>
+        target triple = "{triple}"
+
+        declare i32 @sum(i32 %.1, i32 %.2)
+
+        define i32 @sum_twice(i32 %.1, i32 %.2) {{
+            %.3 = add i32 %.1, %.2
```
sklam: Should make `sum_twice` reference `sum`.  The current tests passes without the `.add_object_file`.

llvmlite/tests/test_binding.py
```diff
@@ -1273,22 +1274,82 @@ def test_inlineasm(self):
...
+            target_machine)
+
+        jit.add_object_file(obj)
+
+        sum_twice = CFUNCTYPE(c_int, c_int, c_int)(
+            jit.get_function_address("sum_twice"))
+
+        assert sum_twice(2, 3) == 10
+
+    def test_add_object_file_from_filesystem(self):
```
sklam: For this test, I'm sometimes getting. `LLVM ERROR: Malformed MachO file` or just a bus error.

sklam: @stuartarchibald said he cannot replicate this on Linux.

Will need to do a debug build of llvm to check this.

alendit: On a related note: is there a way to build conda `llvmdev` recipe with debugging flags?

alendit:
Huh, again an segfault on mac os. @sklam do you think you'd get a change to look into it on mac os. Otherwise I could try to get access to a mac so system somewhere.

llvmlite/tests/test_binding.py
sklam: This problem seems to be resolved by llvm7.

sklam:
Pushed a merge master commit to kick the CI.

llvmlite/tests/test_binding.py
sklam: The problem was not resolved by llvm7.  There was a ownership issue with the `LLVMPY_CreateObjectFile`.  It was more apparent in the test against pypy as the interpreter in Azure.  I fixed it in https://github.com/numba/llvmlite/pull/389/commits/da02109703211a0b0ab0798e2068483c6db002dc.

stuartarchibald:
Well debugged @sklam :)

alendit:
Great catch!

PR #394, opened 2018-09-13
Enable  IntelJIT events for LLVM for VTune support

anton-malakhov:
As title. @sklam @seibert


conda-recipes/llvmdev/build.sh
```diff
@@ -51,7 +51,8 @@ if [[ $(uname) == Darwin ]]; then
   # this causes link failures building the santizers since they respect DARWIN_osx_ARCHS. We may as well
   # save some compilation time by setting this for all of our llvm builds.
   _cmake_config+=(-DDARWIN_osx_ARCHS=x86_64)
-#elif [[ $(uname) == Linux ]]; then
+elif [[ $(uname) == Linux ]]; then
+  _cmake_config+=(-DLLVM_USE_INTEL_JITEVENTS=ON)
```
stuartarchibald: What effect does this flag have on ARM?

anton-malakhov: Well, besides it is just useless there, it might fails to compile because implementation defines inline assembly in order to define atomic operation: https://github.com/llvm-mirror/llvm/blob/master/lib/ExecutionEngine/IntelJITEvents/ittnotify_config.h#L268-L271
Though I'm not sure if it can go that far, cmake might just disable it earlier. Do you see a failure already?

stuartarchibald: Indeed, can't imagine it'll do anything useful :). This was more about will it fall over when building as a result of this change. I'll run it on the build farm and see.

stuartarchibald:
+1 for the proposed change, but please could the build number(s) in the conda build `meta.yaml` be incremented to reflect that this build is technically different to the previous. Once done I'll do a rebuild on the farm to check this works for all architectures. Thanks.

stuartarchibald:
Thanks for the fix @anton-malakhov

stuartarchibald:
This built fine on all platforms for both llvmdev and llvmlite. Am running Numba smoke tests against these builds with https://github.com/numba/numba/pull/3307

stuartarchibald:
Numba smoke tests were ok.

stuartarchibald:
Thanks for the contribution. Merging.

PR #400, opened 2018-10-06
conda recipe: add make as build dep

nehaljwani:
Don't assume that make is part of the build container


conda-recipes/llvmlite/meta.yaml
```diff
@@ -21,6 +21,7 @@ requirements:
     - {{ compiler('cxx') }}  # [not (osx or armv6l or armv7l or aarch64)]
     # The DLL build uses cmake on Windows
     - cmake          # [win]
+    - make           # [unix]
```
stuartarchibald: I think this ought to be `# [unix and not (armv6l or armv7l or aarch64)]` as we use native make on ARM?

stuartarchibald:
Thanks for the fix up @nehaljwani, merging.

PR #403, opened 2018-10-11
Add shutdown detection to ObjectRef __del__ method.

### LINKED ISSUE #350: (new) Warning at shutdown

stuartarchibald:
On older Pythons (<3.4) the shutdown sequence sets globals etc
to None to help gc. This means that the calling of a global as
part of `__del__` as occurs in `ffi::ObjectRef` leads to error
messages like:
`Exception TypeError: "'NoneType' object is not callable" in...`

Using the already existing shutdown detection mechanism in the
`__del__` it's possible to avoid calling the global and as a
result stops these message/calling None. The LLVM object will
leak either way, at least this way is leaks quietly.

Fixes #350
Fixes numba/numba#2818

### DIFF
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -165,7 +165,8 @@ class OutputString(object):
         # Avoid errors trying to rely on globals and modules at interpreter
         # shutdown.
         if not _is_shutting_down():
-            self.close()
+            if self.close is not None:
+                self.close()
 
     def __str__(self):
         if self._ptr is None:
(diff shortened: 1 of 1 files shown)


seibert:
Hah, that is a clever way to capture the global so that it doesn't disappear.  Took me a minute to see what you were doing there.  :)

PR #414, opened 2018-11-02
Add requirement for graphviz

zhangysh1995:
BTW, this file says

>  python < 3.4

, but README says 

> 2.7 or 3.4+

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-enum34; python_version < '3.4'
+graphviz; enum34; python_version < '3.4'


seibert:
The `python < 3.4` marker on that line is indicating that `enum34` is only required for Python versions less than 3.4, which is basically 2.7 at this point.  Normally `graphviz` would be added on a separate line, since it doesn't depend on which Python version you have.

However, `graphviz` is an optional dependency (you don't need it to use most of llvmlite), which is why we don't list it in the `requirements.txt`.

Thanks for the suggestion, but we want to keep the `requirements.txt` as is.

PR #426, opened 2018-11-26
Fix pickling of IR functions and add tests

alendit:
Pickling of the `ir.Function` instances was broken due to interaction of optional arguments in `AttributeSet` constructor and it's children. This PR also adds a couple of tests for this.


alendit:
I also had to replace identity check on `Undefined` with an `instanceof` check. Apparently, unpickling doesn't work well with singletons, although I though to have read otherwise. There is also a test for it now.

PR #427, opened 2018-11-26
Decimal and Half Precision Float Types

hassanalinali:
I've added support for 16 bits and 128 bits wide floating points, I'm still unsure for the naming so you can comment on that if you wish. There are 2 more types specified in [https://releases.llvm.org/6.0.0/docs/LangRef.html#floating-point-types](LLVM's documentation) (ppc_fp128 and x86_fp80) but I'm unsure of the use cases and if someone wants them.

### DIFF
--- a/llvmlite/ir/types.py
+++ b/llvmlite/ir/types.py
@@ -316,7 +316,35 @@ class DoubleType(_BaseFloatType):
         return _format_double(value)
 
 
-for _cls in (FloatType, DoubleType):
+class DecimalType(_BaseFloatType):
+    """
+    The type for 128 bits floats.
+    """
+    null = '0.0'
+    intrinsic_name = 'f128'
+
+    def __str__(self):
+        return 'fp128'
+
+    def format_constant(self, value):
+        return _format_double(value)
+
+
+class HalfPrecisionFloatType(_BaseFloatType):
+    """
+    The type for 16 bits floats.
+    """
+    null = '0.0'
+    intrinsic_name = 'f16'
+
+    def __str__(self):
+        return 'half'
+
+    def format_constant(self, value):
+        return _format_double(value)
+
+
+for _cls in (FloatType, DoubleType, DecimalType, HalfPrecisionFloatType):
     _cls._create_instance()
 
 


seibert:
Hi, thanks for the contribution!  Can you add documentation to `docs/source/user-guide/ir/types.rst` and tests to `llvmlite/tests/test_ir.py` (in particular adding the new types to the list in `TestTypes.assorted_types()` and `TestTypes.test_str()`)?

hassanalinali:
Thank you for your guide. I will update the documentation and tests as well. Let me know if you agree on the class names and if you also want the 2 other float types specified in the LLVM specification ( They are not useful for me, but if someone needs them, I can also include it)

hassanalinali:
I changed the name of the half type because it was too long and I updated the tests and documentation, I also tried to closely follow what you had before for float and double. Let me know if it's ok.

hassanalinali:
@seibert May I ask for a review for this pull request?

hassanalinali:
I realized now that the `format` method might need a change, do you mind guiding me with the required modifications?

the-sushi:
It might be better to go with Quad instead of Decimal (following Half and Double, though at that point perhaps there should be a Single).

gmarkall:
Many thanks for the PR and your efforts so far!

I'm spending some time going through the llvmlite PR backlog - this PR seems to be one that's been stuck for a while. Since it seemed to get stuck waiting for further feedback from a reviewer I've added it to the next milestone and will take a look to see if I can provide some helpful feedback.

gmarkall:
In order to see how this does with current CI I've merged master and resolved conflicts, then pushed to this PR. Note that #509 already added a half-type, so only the 128-bit type is now added by this PR.

gmarkall (changes requested):
Many thanks for the PR - I've now reviewed this, and I think there are a couple of changes and additions that need to be made:

- I agree with @the-sushi's earlier comment that the type name should be `QuadType` - a Decimal type generally doesn't mean something with a floating-point representation, so `QuadType` will both be more consistent with the other floating-point types, and less confusing for users.
- I think `_triop_intrinsic` will need updating to accept the Quad type, otherwise it won't be possible to use the Quad type for FMA operations.
- I think the way the representation is produced will be insufficient to output all the data representing a Quad - can you add a test similar to `test_half_repr` that tests the representation, and modify the `format_constant` function if required please?

I hope this provides some helpful directions - if you're unsure about any of the above or have any questions, please let me know. Also, I am aware it's been a long time since this PR received any attention, so if you don't have time to work on this now please let me know and we can look at finding some effort from elsewhere to finish up this PR (I think it's fundamentally OK, just needs some modifications to get it over the line).

Many thanks in anticipation!

PR #434, opened 2018-12-14
Add another thread for RPi builds.

stuartarchibald:
This adds another thread for LLVM building, RPi3B+ can handle it.


seibert:
looks good

PR #447, opened 2019-01-02
[WIP] Use setuptools pkg_resources to load archived shared lib properly.

sklam:
Use `pkg_resources` to get auto extract archived shared library for frozen application.

### DIFF
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -116,8 +116,17 @@ if os.name == 'nt':
     os.environ['PATH'] += ';' + _lib_dir
 
 _lib_name = get_library_name()
+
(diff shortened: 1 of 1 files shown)


franzhaas:
Dear sklam,

I am late to the party, but i do believe that this solution might be improved upon.

What do you think about this (untested):

import importlib.resources                                                                  #this is part of the python 3.7 and newer
with importlib.resources.path(__name__, _lib_name) as _lib_path:
    lib = ctypes.CDLL(_lib_path)

The problem with pkg_resources is that in my experience it does not work well when runing out of single file executables. It litteres temporary files...
Also it may or may not be present at runtime, making the system unecessary complex.

PR #448, opened 2019-01-05
Fix "SyntaxWarning: invalid escape sequence \d"

jstasiak:
This needs to be a raw string literal so that Python won't try to interpret `\d`.


seibert:
Good catch.  Thanks!

PR #449, opened 2019-01-07
Consolidate the two vector type PRs

sklam:
Closes #313 and #322.

Minor adjustments:
- rename .elementtype to .element
- adjust some docstrings


seibert:
This looks good.  Will merge when CI passes.

jstasiak:
This is a really welcome addition (I'm writing a toy compiler and could use this)! Is there a plan to release new llvmlite version with this change? (no biggie if not, I'll just vendor it)

seibert:
The next llvmlite release will contain these changes.  Our plan is to do a release candidate at the end of February.

If you are using conda, we also build development packages from master that you can install with:
```
conda install -c numba/label/dev llvmlite
```

PR #452, opened 2019-01-11
Some adjustments to pr426

sklam:
(base on #426)

* `Undefined` equality should be checked by object identity; not by `isinstance`.
* Adds more tests to IR object pickling
* Adds singleton tests for Undefined.


seibert:
Look good!

PR #454, opened 2019-01-22
Truncate long label names when adding label suffix.

### LINKED ISSUE #453: Long label name crashes llvm7
Some builder API like the `.if_then` will add suffix (i.e. '.if' and '.endif').  When nested a lot, the label names can grew to a point that it corrupts LLVM7.  Older LLVM works without the problem.

sklam:
Fixes #453.
Also the long label names are not useful anyway.

Label names will now be truncated when too long.  The cutoff point is picked arbitrarily.

### DIFF
--- a/llvmlite/ir/builder.py
+++ b/llvmlite/ir/builder.py
@@ -134,6 +134,22 @@ def _castop(opname, cls=instructions.CastInstr):
     return wrap
 
 
+def _label_suffix(label, suffix):
+    """Returns (label + suffix) or a truncated version if it's too long.
+    Parameters
+    ----------
+    label : str
+        Label name
+    suffix : str
+        Label suffix
+    """
+    if len(label) > 50:
+        nhead = 25
+        return ''.join([label[:nhead], '..', suffix])
+    else:
+        return label + suffix
+
+
 class IRBuilder(object):
     def __init__(self, block=None):
         self._block = block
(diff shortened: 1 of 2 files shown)


llvmlite/ir/builder.py
```diff
@@ -134,6 +134,22 @@ def wrapped(self, val, typ, name=''):
...
+    Parameters
+    ----------
+    label : str
+        Label name
+    suffix : str
+        Label suffix
+    """
+    if len(label) > 50:
+        nhead = 25
+        return ''.join([label[:nhead], '..', suffix])
```
seibert: My only concern with this approach is that it is possible to get to get a short cycle (period 4 if `.endif` is the suffix) of repeating labels:
```
thefunc.endif.endif.endif...endif
thefunc.endif.endif.endif...endif.endif
thefunc.endif.endif.endif...endif.endif.endif
thefunc.endif.endif.endif...endif.endif.endif.endif
thefunc.endif.endif.endif...endif
thefunc.endif.endif.endif...endif.endif
thefunc.endif.endif.endif...endif.endif.endif
[...and so on...]
```
Would that create a problem?

sklam: Label names are automatically deduplicated via https://github.com/numba/llvmlite/blob/21bce30c67a457ad5eac131bdaca769ddee55afe/llvmlite/ir/values.py#L234-L235

seibert: OK, then this seems fine.

seibert:
I think this is ready to merge.

PR #460, opened 2019-02-07
LLVM 7 changed memset intrinsic signature, adjust it

alendit:
Since LLVM 7 memset doesn't take an alignment as a third parameter and respects first argument's alignment instead, see https://releases.llvm.org/7.0.0/docs/LangRef.html#llvm-memset-intrinsics

Also fixes `memcpy` which has experienced the same change between 6 and 7.

### DIFF
--- a/llvmlite/ir/module.py
+++ b/llvmlite/ir/module.py
@@ -179,7 +179,7 @@ class Module(object):
         elif len(tys) == 2:
             if intrinsic == 'llvm.memset':
                 tys = [tys[0], types.IntType(8), tys[1],
-                       types.IntType(32), types.IntType(1)]
+                    types.IntType(1)]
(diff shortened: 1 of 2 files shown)


alendit:
Continuing the discussion on gitter: `memset` and co work with 5 arguments still because of the autoupgrade code in http://llvm.org/doxygen/AutoUpgrade_8cpp_source.html (lines 675ff.) The official signature seems to have changed to the 4 argument variant and the LLVM devs apparently just forgot to update the "Semantics" section in the IR Reference. 

I think it's better to update to the newer signature. The required adjustment in numba is also minimal (2 lines in cgutils.py).

seibert:
We need to reopen this PR and figure out why is caused failures with LLVM 8.

Issue #415, opened 2018-11-02
Any way to convert ModuleRef to Module?


zhangysh1995:
I want to read in a `bitcode` or `llvmir` file, and convert the results (ModuleRef) from `parse_bitcode` or `parse_assembly` to `Module`.

I checked the constructor of Module, there is no options to do this.

zhangysh1995:
Because I need to **add functions or instrument** a Module, so I need to do the conversion.

hosewiejacke:
Even though I'd like to have this feature too things don't look good according to one of the contributors:

> llvmlite has two separate parts 1. ir which is pure python implementation of LLVM IR building and 2. binding which contains wrappers for LLVM C/C++ API. ValueRef is a part of binding and, even though it is a semantic equivalent of Value from ir, there is no connection between them in llvmlite. So, when you create a ModuleRef from assembly/bitcode there is not connection to the ir subsystem, either.

Source: @alendit at https://github.com/numba/llvmlite/issues/342

zhangysh1995:
@hosewiejacke So, it is possible to convert `ModuleRef` to `Module` in LLVM? But it's not implemented in `llvmlite`.

zhangysh1995:
I checked the implementation of LLVM, but didn't see **any explicit API** to convert between these two representations.

zhangysh1995:
I found how to convert to Module in LLVM and I think it's possible to implement it in llvmlite.
```` C
#include "llvm-c/Core.h"

ModuleRef moduleref; // assume we have this
Module m = unwrap(moduleref);
````

hosewiejacke:
> @hosewiejacke So, it is possible to convert `ModuleRef` to `Module` in LLVM? But it's not implemented in `llvmlite`.

I don't know whether it's possible or not in LLVM. I stumbled over the comment from alendit when I was looking for this myself. Depending on what you're trying to accomplish using the python bindings from LLVM7 might be an option.

alendit:
`ModuleRef` is just a typedef for `llvm::Module*` to facilitate communication with the `llvm-c` API. Both have nothing to do with `llvmlite.ir.Module`. There is currently to way to convert from C/C++ module (e.g. `ModuleRef`) to `ir.Module`. 

One way to do it, would be to write a parser to convert llvm IR/bitcode to python objects. Another would be to add a `NodeVisitor` which would walk the `ModeRef` instance and emit the correct `ir.Value` objects which would be used to create a `ir.Module`. 

[...]

Issue #342, opened 2018-03-28
Get Value from ValueRef


brendabrandy:
I am currently trying to manipulate an existing llvm IR file. I have the following IR file that I compiled from a simple hello world function: 
```
@.str = private unnamed_addr constant [13 x i8] c"Hello World\0A\00", align 1

; Function Attrs: noinline nounwind optnone ssp uwtable
define i32 @main() #0 {
  %1 = alloca i32, align 4
  store i32 0, i32* %1, align 4
  %2 = call i32 (i8*, ...) @printf(i8* getelementptr inbounds ([13 x i8], [13 x i8]* @.str, i32 0, i32 0))
  ret i32 0
}
```

How can I change "Hello World" to "Goodbye World"? I can currently obtain the ValueRef object but I am confused as to how to get the actual Value of the object so I can manipulate its contents.

Thanks!

alendit:
Short answer: you can't right now. 

Long answer: llvmlite has two separate parts 1. `ir` which is pure python implementation of LLVM IR building and 2. `binding` which contains wrappers for LLVM C/C++ API. `ValueRef` is a part of `binding` and, even though it is a semantic equivalent of `Value` from `ir`, there is no connection between them in `llvmlite`. So, when you create a `ModuleRef` from assembly/bitcode there is not connection to the `ir` subsystem, either.

I've made #365 which makes an initial attempt to bridge the gap between `TypeRef` and `Type`, but the better solution would be to add conversion between `ModuleRef` and `Module` and all the child types. I'm not sure if it's the direction the authors want to go into, so I'll wait for the reaction to the PR, before I add any further implementations.

PR #466, opened 2019-03-06
Add ValueRef iterators

pearu:
This PR implements the following features:
- [x] `ValueRef` properties `is_function`, `is_argument`, `is_block`, `is_instruction`, `is_operand`
- [x] `ValueRef` new iterator `blocks` for functions
- [x] `ValueRef` new iterator `arguments` for functions
- [x] `ValueRef` new iterator `instructions` for basic blocks
- [x] `ValueRef` new iterator `operands` for instructions
- [x] `ValueRef` new property `opcode` that returns instruction's Opcode name
- [x] `ValueRef` new properties `module`, `function`, `block`, and `instruction` as parent values for `function`, `block` or `argument`, `instruction`, `operand` value instances, respectively
- [x] Disallow using methods that are available only for certain Value kinds (functions, arguments, blocks, instructions, operands)
- [x] Documentation of the new features
- [x] Add unittests.
- [x] `ValueRef` new iterator `attributes` for functions and arguments.

The PR also provides solution to issues #415 and #342 .


llvmlite/binding/module.py
```diff
@@ -240,6 +240,24 @@ def _next(self):
         return ffi.lib.LLVMPY_FunctionsIterNext(self)
 
 
+class _BlocksIterator(_Iterator):
+
+    def _dispose(self):
+        self._capi.LLVMPY_DisposeBlocksIter(self)
+
+    def _next(self):
+        return ffi.lib.LLVMPY_BlocksIterNext(self)
```
sklam: `LLVMPY_BlocksIterNext` is missing the type declaration for its `.restype`. it is being assumed to be `ctypes.c_int`.  You will need to add `ffi.lib. LLVMPY_BlocksIterNext.restype = ffi.LLVMValueRef`.

llvmlite/binding/module.py
```diff
@@ -240,6 +240,24 @@ def _next(self):
...
+
+
+class _InstructionsIterator(_Iterator):
+
+    def _dispose(self):
+        self._capi.LLVMPY_DisposeInstructionsIter(self)
+
+    def _next(self):
+        return ffi.lib.LLVMPY_InstructionsIterNext(self)
```
sklam: Same as above.  Need to define `.restype`.

llvmlite/binding/module.py
pearu: Fixed.

llvmlite/binding/module.py
pearu: Fixed.

sklam:
@pearu, once there are tests, i think it's ready for review.

There're already things in numba that I can rewrite using the new API you added.  Btw, can you tell me your usecase?  Are you working on an analysis pass on the IR?  Perhaps, even a transformation pass?

pearu:
@sklam The main use case is to analyze the rbc/numba generated IRs for completeness as well as to update the generated IRs with new functions:

1. Since the RBC server is not expected to be numba-aware (e.g. the server is implemented in C++), I would need to model the RBC server using llvmlite only. The RBC server has application-specific IR module generator to which the client IR will be joined.

2. During the development of RBC, I have encountered cases where IRs are incomplete (they lack function definitions that the functions in IR are calling) and these leads to crashing the server when calling the functions. So the idea is that before sending the IRs to server, the client will analyze the IRs in order to catch such issues as early as possible.

3. A part of this work is also educational. RBC tries to reuse numba tools as much as possible but sometimes it is not easy to use numba directly as for RBC the numba API would have to be refactored to separate IR producing and IR compiling parts.

I'll add unittests for the iterators but if you can tell beforehand which parts of the PR would not be needed, that would be great. For instance, I have introduced new attributes such as `is_<value kind>` that usage sofar is to ensure that the ValueRef instances contain iterator specific data. When user llvmlite specific algorithms are correct, the `is_<value kind>` attributes might not be needed.

sklam:
For 2), is it sufficient just to scan for declare-only function?  I believe DCE will kill unused declarations or you can find the "use-list" for globals to see if a global is used (see http://llvm.org/doxygen/group__LLVMCCoreValueUses.html).

sklam:
Let's keep the `is_<kind>` attributes.  I think they can be useful if we need to add more APIs to traverse the IR.  The current implementation is getting the kind base on usage.  But the kind could be unknown is other API.  There's a [`LLVMGetValueKind`](http://llvm.org/doxygen/group__LLVMCCoreValueGeneral.html#ga04768f6545009fb71119499c6735b544) when we need to tell the `kind` dynamically.

pearu:
@sklam please review.

llvmlite/binding/value.py
```diff
@@ -140,6 +186,7 @@ def add_function_attribute(self, attr):
         attr : str
             attribute name
         """
+        assert self.is_function
```
sklam: I think it's better to explicitly raise an exception instead of using `assert` in case someone runs with `python -O` (assertions optimized out).

llvmlite/binding/value.py
```diff
@@ -163,6 +210,185 @@ def is_declaration(self):
...
+            itr = _AttributeSetIterator(it)
+        return itr
+
+    @property
+    def blocks(self):
+        """
+        Return an iterator over this function's blocks.
+        The iterator will yield a ValueRef for each block.
+        """
+        assert self.is_function
```
sklam: same here.

llvmlite/binding/value.py
```diff
@@ -163,6 +210,185 @@ def is_declaration(self):
...
+        parents.update(function=self)
+        return _BlocksIterator(it, parents)
+
+    @property
+    def arguments(self):
+        """
+        Return an iterator over this function's arguments.
+        The iterator will yield a ValueRef for each argument.
+        """
+        assert self.is_function
```
sklam: same here

llvmlite/binding/value.py
```diff
@@ -163,6 +210,185 @@ def is_declaration(self):
...
+        parents.update(function=self)
+        return _ArgumentsIterator(it, parents)
+
+    @property
+    def instructions(self):
+        """
+        Return an iterator over this block's instructions.
+        The iterator will yield a ValueRef for each instruction.
+        """
+        assert self.is_block
```
sklam: same here

llvmlite/binding/value.py
```diff
@@ -163,6 +210,185 @@ def is_declaration(self):
...
+        parents.update(block=self)
+        return _InstructionsIterator(it, parents)
+
+    @property
+    def operands(self):
+        """
+        Return an iterator over this instruction's operands.
+        The iterator will yield a ValueRef for each operand.
+        """
+        assert self.is_instruction
```
sklam: same here

llvmlite/binding/value.py
```diff
@@ -163,6 +210,185 @@ def is_declaration(self):
...
+        it = ffi.lib.LLVMPY_InstructionOperandsIter(self)
+        parents = self._parents.copy()
+        parents.update(instruction=self)
+        return _OperandsIterator(it, parents)
+
+    @property
+    def opcode(self):
+        assert self.is_instruction
```
sklam: same here

llvmlite/binding/value.py
pearu: Agreed and fixed.

llvmlite/binding/value.py
pearu: Fixed.

llvmlite/binding/value.py
pearu: Fixed.

llvmlite/binding/value.py
pearu: Fixed.

llvmlite/binding/value.py
pearu: Fixed.

Issue #468, opened 2019-03-10
Support loading from egg files


aiudirog:
I have been working on a Hadoop cluster environment with PySpark where I don't have permissions to add packages to nodes directly, so I have been deploying extra libraries using egg files. My current project gets a 2000x speed increase using Numba and so I wanted to deploy that as well. However, I discovered that the way libllvmlite.so is loaded can't be done from within an egg file. After a few hours of testing different options I did find two ways of making it work:
1. Using SparkContext.addFile to add libllvmlite.so to the working directory and then patching ffi.py to load from the current directory.
2. Using pkg_resources to load from the extracted egg file in the cache.

I bundled both of these options up into a single patch which skips the pkg_resource option if it isn't importable to prevent adding a new required dependency: https://github.com/aiudirog/llvmlite/commit/4b144662b6f56afd21f351f4dcc562c0a455ac9b

I am not sure how to create proper test cases for this, so any help would be appreciated.

aiudirog:
Closing as pull request has been merged. Thanks for the quick turnaround!

PR #469, opened 2019-03-10
Support loading from current directory and egg files

aiudirog:


aiudirog:
Pull request for Issue #468

aiudirog:
Just noticed this pull request which solves the same issue: https://github.com/numba/llvmlite/pull/447

sklam:
It looks like this PR will try more search paths than #447.  I am inclined to close #447 in favor for this one.

PR #470, opened 2019-03-12
Fix leak on string returning APIs.

sklam:
* All string returning C-API must return a `LLVMPY_CreateString` allocated string and the Python side must deallocate using `LLVMPY_DisposeString` or via `OutputString`.
* Added `ret_string` and `ret_bytes` to wrap the string return-value for cleanup.
* All string returning C-API must use `c_void_p` to avoid `ctypes` auto-conversion of `c_char_p` into a python string.

### DIFF
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -149,8 +149,20 @@ class OutputString(object):
     """
     _as_parameter_ = _DeadPointer()
 
-    def __init__(self, owned=True):
-        self._ptr = ctypes.c_char_p(None)
+    @classmethod
+    def from_return(cls, ptr):
+        """Constructing from a pointer returned from C-API.
+        The pointer must be allocated with LLVMPY_CreateString.
+
+        Note
+        ----
+        Because ctypes auto convert *restype* of *c_char_p* into a python
+        string, we must use *c_void_p* to obtain the raw pointer.
+        """
+        return cls(init=ctypes.cast(ptr, ctypes.c_char_p))
+
+    def __init__(self, owned=True, init=None):
+        self._ptr = init if init is not None else ctypes.c_char_p(None)
         self._as_parameter_ = ctypes.byref(self._ptr)
         self._owned = owned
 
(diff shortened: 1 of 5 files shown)


stuartarchibald:
Thanks for the patch. Wondering about whether this is leaking too:
https://github.com/numba/llvmlite/blob/fcadf8af11947f3fd041c5d6526c5bf231564883/ffi/value.cpp#L359-L363
?

llvmlite/binding/ffi.py
```diff
@@ -149,8 +149,20 @@ class OutputString(object):
...
-    def __init__(self, owned=True):
-        self._ptr = ctypes.c_char_p(None)
+    @classmethod
+    def from_return(cls, ptr):
+        """Constructing from a pointer returned from C-API.
+        The pointer must be allocated with LLVMPY_CreateString.
+
+        Note
+        ----
+        Because ctypes auto convert *restype* of *c_char_p* into a python
```
stuartarchibald:
```suggestion
        Because ctypes auto-converts *restype* of *c_char_p* into a python
```

llvmlite/binding/ffi.py
```diff
@@ -149,8 +149,20 @@ class OutputString(object):
     """
     _as_parameter_ = _DeadPointer()
 
-    def __init__(self, owned=True):
-        self._ptr = ctypes.c_char_p(None)
+    @classmethod
+    def from_return(cls, ptr):
+        """Constructing from a pointer returned from C-API.
```
stuartarchibald:
```suggestion
        """Constructing from a pointer returned from the C-API.
```

stuartarchibald:
@sklam do you have a view on https://github.com/numba/llvmlite/pull/470#issuecomment-484820396? Else I think this can be merged, will probably help https://github.com/numba/numba/issues/4368.

sklam:
@stuartarchibald 

> do you have a view on #470 (comment)? 

`LLVMPY_PrintValueToString` is not leaking.  All caller is using the `OutputString` to clean up.

stuartarchibald:
Thanks for checking, this can be merged as convenient.

PR #478, opened 2019-04-04
Upgrade to llvm8

stuartarchibald:
As title.


stuartarchibald:
Travis CI is failing because it does wheel builds and there's been no LLVM 8 based wheel build out at this point, the rest of the builds seem fine.

seibert:
This looks good.  Ready to merge.

conda-recipes/0001-Revert-Limit-size-of-non-GlobalValue-name.patch
```diff
@@ -0,0 +1,70 @@
+From 008dc9e0a47d9ec70b0583b01bc71bd892f3de24 Mon Sep 17 00:00:00 2001
+From: Stuart Archibald <redacted>
+Date: Mon, 8 Apr 2019 16:30:24 +0100
+Subject: [PATCH] Revert "Limit size of non-GlobalValue name"
+
+https://reviews.llvm.org/D41296 introduced a limit on the length of a
+non-GlobalValue name, this patch is a direct revert of that change (as code gen
+can sometimes legitimately create very long names).
```
aaronpuchert: Can I ask why? The limit is pretty generous with 1024 characters. Note that label names are there to increase readability, and I don't see how someone can reasonably read anything with an identifier that long.

And if you're not reading the IR anyway, why not just number the labels, as other IR generators do?

stuartarchibald: Before adding this patch some effort was put in to reduce/shorten label names at the site of code generation but we concluded it was too risky to ship as we couldn't absolutely guarantee the fix.

stuartarchibald: xref: https://github.com/numba/numba/issues/3876

aaronpuchert: Hmm, but why have the names at all? When I compile this code with Clang:

```c
int f(int a, int b)
{
    if (a > b)
        return a;
    else
        return b;
}
```

I get:
```llvm
define dso_local i32 @f(i32, i32) #0 {
  %3 = alloca i32, align 4
  %4 = alloca i32, align 4
  %5 = alloca i32, align 4
  store i32 %0, i32* %4, align 4
  store i32 %1, i32* %5, align 4
  %6 = load i32, i32* %4, align 4
  %7 = load i32, i32* %5, align 4
  %8 = icmp sgt i32 %6, %7
  br i1 %8, label %9, label %11

; <label>:9:                                      ; preds = %2
  %10 = load i32, i32* %4, align 4
  store i32 %10, i32* %3, align 4
  br label %13

; <label>:11:                                     ; preds = %2
  %12 = load i32, i32* %5, align 4
  store i32 %12, i32* %3, align 4
  br label %13

; <label>:13:                                     ; preds = %11, %9
  %14 = load i32, i32* %3, align 4
  ret i32 %14
}
```

This is completely fine. It's hardly readable, but do you care about that? Your users aren't reading the LLVM IR, are they? The local value names are basically a debugging tool and I won't even get them from Clang unless I specify `-fno-discard-value-names`.

aaronpuchert: I've had a look at the upstream patch that you reverted and the discussion indicates that they wanted to shorten the value names generated by intermediate passes. It doesn't seem to be intended to limit the length of names in a parsed file. And even if it was intended for that, the error shouldn't be “Terminator found in the middle of a basic block!”, as you wrote in numba/numba#3669. So perhaps you're right and this is indeed a bug.

However, they did this because they would generate names so long that the optimizer could run out of memory, and perhaps this can happen to you as well. So I think the path pursued by #454 is the right one, and in my opinion you should try to make sure that your names stay at reasonable lengths.

stuartarchibald: > I've had a look at the upstream patch that you reverted and the discussion indicates that they wanted to shorten the value names generated by intermediate passes. It doesn't seem to be intended to limit the length of names in a parsed file. And even if it was intended for that, the error shouldn't be “Terminator found in the middle of a basic block!”, as you wrote in [numba/numba#3669](https://github.com/numba/numba/issues/3669). So perhaps you're right and this is indeed a bug.

Thanks for taking a look at the upstream patch. I have a reproducer for the `Terminator found in the middle of a basic block!` issue:
```
curl https://gist.githubusercontent.com/stuartarchibald/4bca524a6ae6a34f885f8ccb6ab1d992/raw/da755ef0f2953154753434efc70fff65b449c561/before.ll -O; opt -verify before.ll -o /dev/null 2>&1; opt -instcombine  before.ll -S -o newafter.ll; opt  -verify newafter.ll -o /dev/null
```
these two are similar, but are more likely a direct result of label truncation so are less suspicious as potential LLVM bugs:
https://github.com/numba/numba/issues/3680
https://github.com/numba/numba/issues/3666

> However, they did this because they would generate names so long that the optimizer could run out of memory, and perhaps this can happen to you as well. So I think the path pursued by #454 is the right one, and in my opinion you should try to make sure that your names stay at reasonable lengths.

#454 fixes one of the easier places where long labels may occur, and indeed this is probably the best approach to take. However, moving all our supported platforms to LLVM 8 and the debugging required to do that took up a huge amount of time (PPC64LE sadly didn't make it, too many problems), this meant that the extra issues/fixes landed relatively late in a release cycle. The knock-on effect being that this patch to allowed us to ship, confident in that we we're not breaking users with the upgrade as we could not guarantee we'd caught all the long labels in the code generation changes.

aaronpuchert: Did you file that bug with LLVM upstream? Otherwise I would do so, and perhaps investigate how to fix it.

Can I understand your second statement as saying that the patch is primarily a temporary safety measure and you won't need it anymore when you feel confident that overly long labels can no longer occur?

stuartarchibald: > Did you file that bug with LLVM upstream? Otherwise I would do so, and perhaps investigate how to fix it.

I don't recall doing so, it wasn't apparent at the time that it was definitely a bug as it coincided with numerous bug reports on the Numba issue tracker caused by https://reviews.llvm.org/D41296 and it wasn't clear what was to blame/that they may be different problems. If you are convinced that this is a bug we'll file upstream.

> Can I understand your second statement as saying that the patch is primarily a temporary safety measure and you won't need it anymore when you feel confident that overly long labels can no longer occur?

Correct, at some point we hope to adjust label generation in a way that we are confident won't break things and this patch reversion will be removed. We are aware the current situation is not ideal, but sometimes just have to take a view on where the most risk lies.

aaronpuchert: I'm now convinced it's a bug, as it means that in some cases LLVM can't parse its own output. I've added a [comment](https://reviews.llvm.org/D41296#1607582) to the upstream change and asked for opinions on how to fix this. Let's see where it goes. It shouldn't be hard to fix, but it will probably touch some core parts of LLVM, so there'll be some discussion about which is the right way. I'll inform you when a fix has been posted for review or merged to the code line.

stuartarchibald: Thanks very much, following up on this is much appreciated.

aaronpuchert: FYI, it seems that truncation during parsing will now always lead to an error (https://reviews.llvm.org/D145282).

PR #483, opened 2019-05-09
Depend on enum34 using PEP 508 environment markers

niallwingham:
The recommended way to specify dependencies specific to a python
version is to use environment markers.

One downside of using if statements to specify dependencies is that it
can break applications using package managers like pipenv or poetry,
where the version of python running setup.py is not necessarily the
same verion of python being deployed.

It would certainly be fair to blame this on the package managers
themselves, but given the PEP and the fact that other packages have
solved this in the same way, I think it's worth fixing on this end.

See:
* https://www.python.org/dev/peps/pep-0508/#environment-markers
* https://github.com/hynek/argon2_cffi/issues/48
* https://github.com/hynek/argon2_cffi/commit/8eed81aa

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -158,11 +158,8 @@ packages = ['llvmlite',
             'llvmlite.tests',
             ]
 
-install_requires = []
-setup_requires = []
-if sys.version_info < (3, 4):
-    install_requires.append('enum34')
-    setup_requires.append('enum34')
+install_requires = ['enum34; python_version < "3.4"']
+setup_requires = ['enum34; python_version < "3.4"']
 
 
 with open('README.rst') as f:


niallwingham:
@korijn @esc this addresses #483

Korijn:
> ... where the version of python running setup.py is not necessarily the same verion of python being deployed.

This had not even occurred to me yet as being the root cause. Great analysis & fix!

esc:
@niallwingham thanks for contributing this PR! This is an interesting find. @Korijn have you tried this, can you confirm, that it works?

esc:
@Korijn @niallwingham can you suggest any way to test that this does work? Ideally in an automated fashion, but something manual would be a good start. Just so that I can verify this does indeed solve the problem.

Korijn:
Automated would be quite tricky. Manually, you could go about it like this:

* Build a pre-release distribution (use e.g. `0.29a0` as the version string for alpha in accordance with [PEP 440](https://www.python.org/dev/peps/pep-0440/)) on this branch. I'm not sure if it matters if you go with `sdist` or `bdist_wheel`.
* Upload the pre-release to PyPI.
* Install pipenv into a Python 2.7 environment
* Use your python 2.7-backed pipenv to create a new environment that requires Python 3.5 or higher and the llvmlite pre-release. You can put the following Pipfile in a new directory and call `pipenv lock`.

```
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true

[packages]
llvmlite = "0.29a0"

[requires]
python_version = "3.6"

[pipenv]
allow_prereleases = true
```

* Inspect the lockfile to see if enum34 was included as a dependency or not.

Do note that this is all very sensitive to system configuration, so it might be hard to get the circumstances just right.

esc:
FWIW, the `setup.py` from Numba suffers from the same shortcoming:

https://github.com/numba/numba/blob/master/setup.py#L309

Korijn:
BTW our workaround is precompiling our Numba code and wrapping that in packages, so our runtime environments don't have to depend on Numba or llvmlite.

dsevero:
+1

Just came back from reading a huge log from our application using poetry and the culprit is `enum34` being installed on `python==3.6.7` due to `numba=0.43.1` and `llvmlite==0.28.1`!

dsevero:
I'm guessing the numba package still needs fixing, since the same happens on https://github.com/numba/numba/blob/master/setup.py#L309-L310

PR #484, opened 2019-05-10
Revert "LLVM 7 changed memset intrinsic signature, adjust it"

seibert:
Reverts numba/llvmlite#460.  This seems to be causing issues with LLVM 8.


sklam:
There were no issue with the #460 changes with LLVM 8.  In turns out numba needed updating to match the new intrinsic signature.

PR #485, opened 2019-05-15
Revert "Revert "LLVM 7 changed memset intrinsic signature, adjust it""

sklam:
Reverts numba/llvmlite#484

### DIFF
--- a/llvmlite/ir/module.py
+++ b/llvmlite/ir/module.py
@@ -179,7 +179,7 @@ class Module(object):
         elif len(tys) == 2:
             if intrinsic == 'llvm.memset':
                 tys = [tys[0], types.IntType(8), tys[1],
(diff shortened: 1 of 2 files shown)


seibert:
Given the issues we had with Numba failing after this PR, are we also going to break other users of llvmlite if we make this change?

sklam:
Yes, this is a breaking change.

seibert:
I think we're now big enough that we have to go through the deprecation process then.  We can start that in the 0.30 dev cycle.

stuartarchibald:
xref: https://github.com/numba/llvmlite/pull/511, deprecation notice issued.

stuartarchibald:
Corresponding doc update for 0.31 noting removal. https://github.com/numba/llvmlite/pull/533

PR #487, opened 2019-05-20
Update changelog for 0.29.0

sklam:
as titled


CHANGE_LOG
```diff
@@ -1,11 +1,20 @@
-v0.29.0
--------
+v0.29.0 (RC1: May 20, 2019)
+---------------------------
 
-In development.
+This release upgrades to LLVM 8.0 for all supported platforms except PPC64LE.
+Due to numerous problems with LLVM 8.0 running on PPC64LE,
+we have decided to use LLVM 7.1, which is more stable on PPC64LE.
```
stuartarchibald: Want to mention that we build a few new non-host targets on `llvmdev`? I've not seen what happens from the `llvmlite` side yet! https://github.com/numba/llvmlite/blob/master/conda-recipes/llvmdev/build.sh#L31-L32

sklam: good point, I will add those

seibert:
Looks good, merging!

PR #500, opened 2019-07-18
Obey CXXFLAGS and CPPFLAGS

fschlimb:
Including CXXFLAGS and CPPFLAGS from user env (ignored otherwise).

PR #502, opened 2019-08-02
Add -fPIC flag for manylinux1 wheel building

sklam:
as titled


buildscripts/manylinux1/build_llvmlite.sh
```diff
@@ -33,7 +33,8 @@ python setup.py clean
 # Build wheel
 distdir=$outputdir/dist_$(uname -m)_$pyver
 rm -rf $distdir
-python setup.py bdist_wheel -d $distdir
+# -fPIC is needed since llvm8 when building wheels
+CXX='g++ -fPIC' python setup.py bdist_wheel -d $distdir
```
jvesely: This forces g++ compiler. Isn't `CXXFLAGS="-fPIC" enough?

stuartarchibald: This needs resolving to be merged. I think @jvesely's point is valid under the assumption that the build process acknowledges `CXXFLAGS`, I don't recall if this is true and think https://github.com/numba/llvmlite/pull/500 may be needed?

jvesely: This is a slightly bigger issue than just switching to a different compiler. LLVM packages from apt.llvm.org are built using clang/llvm thus `llvm-config --cxxflags` may include flags that are not supported by gcc/g++.
This leads to errors such as:
```
    g++: error: unrecognized command line option ‘-Wcovered-switch-default’; did you mean ‘-Wno-switch-default’?
    g++: error: unrecognized command line option ‘-Wstring-conversion’; did you mean ‘-Wint-conversion’?
```

jvesely: Even if the build process does not recognize `CXXFLAGS`, `CXX="$CXX -fPIC"` should work.
Fixing it in the build system would be nicer so pip builds from source work correctly.

sklam:
I've merge in #500 into this PR to make CXXFLAGS work.
Now the setup.py is setting `CXXFLAGS=$CXXFLAGS -fPIC` on linux for `bdist_wheel` command.

buildscripts/manylinux1/build_llvmlite.sh
stuartarchibald: This is a good point, thanks for making it, however in the interests of getting mainline working again, I'm going to put this on a new issue and review the following as is.

stuartarchibald: https://github.com/numba/llvmlite/issues/514 tracks

jvesely: I think that the latest version from @sklam, should work.
The Makefile honors `CXXFLAGS` env var and setup.py appends `-fPIC` to the env var if needed.
However, I still see the issue trying to `pip install git+https://github.com/numba/llvmlite.git@master` on pypy3[0]

[0] https://travis-ci.com/jvesely/PsyNeuLink/jobs/240332917#L1007

sklam: I was able to replicate the problem and realized that the patch that I have is only for `bdist_wheel`.  But, `pypy3` doesn't use (and probably doesn't have) `bdist_wheel`.  To workaround the problem, you can set `CXXFLAGS='-fPIC'`.  That worked for my local testing.

jvesely: This isn't `pypy3` specific. `pip install` from source uses `setup.py install` instead of `bdist_wheel`. The same happens when llvmlite is in `requirements.txt` and gets installed as a dependency:
```
 LLVM_CONFIG=/usr/bin/llvm-config-64 pip3 install --user -U llvmlite --no-binary ':all:'
Collecting llvmlite
  Using cached https://files.pythonhosted.org/packages/8b/b0/df26861e6ce2fc91c8bb93ea808fa2e631ee8a29fc4c3bc96626b78dae74/llvmlite-0.30.0.tar.gz
Installing collected packages: llvmlite
  Running setup.py install for llvmlite ... error
```
The error complains about relocation.
 I don't know what logic pip uses when deciding to build local wheels vs. install.

PR #509, opened 2019-09-11
Add Half-Precision Type

njwhite:
In the style of Float and Double

### DIFF
--- a/llvmlite/ir/types.py
+++ b/llvmlite/ir/types.py
@@ -257,6 +257,13 @@ def _as_float(value):
     return struct.unpack('f', struct.pack('f', value))[0]
 
 
+def _as_half(value):
+    """
+    Truncate to half-precision float.
+    """
+    return struct.unpack('e', struct.pack('e', value))[0]
+
+
 def _format_float_as_hex(value, packfmt, unpackfmt, numdigits):
     raw = struct.pack(packfmt, float(value))
     intrep = struct.unpack(unpackfmt, raw)[0]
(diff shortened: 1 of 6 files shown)


seibert:
related: https://github.com/numba/numba/issues/4402

llvmlite/ir/types.py
```diff
@@ -257,6 +257,17 @@ def _as_float(value):
...
 
+def _as_half(value):
+    """
+    Truncate to half-precision float.
+    """
+    try:
+        return struct.unpack('e', struct.pack('e', value))[0]
+    except struct.error:
+        # 'e' only added in Python 3.6+
+        return _as_float(value)
```
stuartarchibald: Might this be a source of confusion, a user asks for a half but when formatted it appears to not have been correctly truncated?

llvmlite/tests/test_valuerepr.py
```diff
@@ -1,18 +1,22 @@
...
-    ArrayType)
+    ArrayType, HalfType)
 from . import TestCase
 
 
 int8 = IntType(8)
 int16 = IntType(16)
 
 
+PY36 = sys.version_info[:2] >= (3, 6)
```
stuartarchibald:
```suggestion
PY36_OR_LATER = sys.version_info[:2] >= (3, 6)
```

stuartarchibald:

llvmlite/ir/types.py
njwhite: I could use numpy to round it ([the logic is complicated!](https://bugs.python.org/issue11734)), but I don't think llvmlite depends on that at the minute. Is that OK? I'd rather use a library to do the rounding - especially as python 2.x is decreasing in popularity.

llvmlite/tests/test_valuerepr.py
njwhite: I've fixed this

llvmlite/ir/types.py
stuartarchibald: @sklam do you have a view on this. Perhaps just documenting that it's different or issuing a warning is the easiest thing to do?

sklam: If not rounded properly, at `parse_assembly()`, it will raise `RuntimeError` with:

```python
<string>:5:27: error: floating point constant invalid for type
@"myhalf" = constant half 0x3f87006680000000
```

I think we should just document it.  Telling user to consider doing something like `float(np.float16(1.12312321e-2))` to make sure the value can be exactly represented as half.

njwhite: @sklam @stuartarchibald I've pushed a change to the documentation that covers this

PR #510, opened 2019-09-11
Add fp16 Intrinsics

njwhite:
..because NVVM IR [doesn't support half types](https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html#type-system)!


llvmlite/ir/builder.py
```diff
@@ -1016,3 +1030,24 @@ def fma(self, a, b, c):
         """
         Perform the fused multiply-add operation.
         """
+
+    def convert_from_fp16(self, a, to=None, name=''):
+        """
+        Convert from an i16 to the given FP type
+        """
+        if not to:
+            raise TypeError("expected a float return type")
```
stuartarchibald: Is this branch exercised in the tests?

stuartarchibald:
This will also need a merge from master once the thing causing the test failures is fixed (probably #513)

stuartarchibald:

llvmlite/ir/builder.py
njwhite: I've added another test so all the error messages are covered now

njwhite:
Thanks for the review @stuartarchibald, I've pushed fixes to address all your comments -

PR #512, opened 2019-09-21
fix for block labels which contain "interesting" characters

trbabb:
llvm will fail to parse the IR from a module if it contains a block with an "interesting" label (i.e. one which is not alphanumeric). These labels need to be quoted and escaped.

Rather than failing with a mysterious parse error, this allows llvmlite to accept the full space of strings as block labels.

(My use case is that the labels were b64 encoded hashes of AST structures. The b64 alphabet includes +, =, and /, which would randomly cause llvmlite to puke).

### DIFF
--- a/llvmlite/ir/values.py
+++ b/llvmlite/ir/values.py
@@ -6,6 +6,7 @@ Instructions are in the instructions module.
 from __future__ import print_function, absolute_import
 
 import string
+import re
 
 from .. import six
 from . import types, _utils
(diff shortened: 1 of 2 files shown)


esc:
@trbabb thanks for submitting this! I have queued it for a review.

stuartarchibald:

trbabb:
Looks like travis-ci is busted for reasons unrelated to this PR, but otherwise does the current change look OK?

stuartarchibald:
@sklam please could you take a look at this. Thanks.

sklam:
The PR looks okay.  I would recommend refactoring the code so that the quotation check is in `NamedValue` but we are too close to cutting a release.  We will do the refactoring later since it is fixing a bug regarding BB name.

PR #513, opened 2019-09-23
Remove restriction on sphinx version from Anaconda distro

stuartarchibald:
As title.


stuartarchibald:
Hopefully this should fix the build constraint issue which will allow PR testing to work again.

seibert:
looks like this fixes CI

PR #515, opened 2019-09-30
Remove conda restriction on appveyor

stuartarchibald:
As title.


esc:
Appveyor is no longer used and the config was removed with #565 -- so I will close this now as it is no longer needed.

PR #518, opened 2019-10-01
Fix use of -fPIC flag in wheels

stuartarchibald:
As title


sklam:
Manually tested that manylinux1 wheel building is working

PR #524, opened 2019-10-13
Allow llvm>=9.0

jschueller:
llvm lite builds fine with LLVM 9.0 that just got released

### DIFF
--- a/ffi/build.py
+++ b/ffi/build.py
@@ -12,6 +12,7 @@ import subprocess
 import shutil
 import sys
 import tempfile
+from distutils.version import LooseVersion
 
 
 here_dir = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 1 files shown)


jschueller:
ping @sklam

stuartarchibald:
Thanks for the PR. `llvmlite` and also the entire Numba stack will actually need testing on all the arch/OS combinations to ensure that it does indeed work ok before declaring it as such, thanks for your understanding.

esemeniuc:
A quick fix is to run
```

stuartarchibald:
Unless I am mistaken, the above patch is not adding support, it is just removing the version check.

fabaff:
This is kind of a blocker for Fedora. Fedora started to ship the release candidates of llvm 10 in the meantime.

stuartarchibald:
LLVM 9 support is now on master, it's finally passed testing on all platforms/OS: https://github.com/numba/llvmlite/pull/548

There's also now an env var to permit building against unsupported LLVMs http://llvmlite.pydata.org/en/latest/admin-guide/install.html#compiling-llvmlite

PR #526, opened 2019-10-15
ENH: Better error message

larsoner:
On Ubuntu 19.10 I got:
```
Building llvmlite requires LLVM 7.0+ Be sure to set LLVM_CONFIG ...
```
Which is misleading because it has LLVM 9 by default. This PR updates the message to say:
```
Building llvmlite requires LLVM 7.0, 7.1, or 8.0, got '9.0.0'. Be sure to set LLVM_CONFIG ...
```

### DIFF
--- a/ffi/build.py
+++ b/ffi/build.py
@@ -112,11 +112,11 @@ def main_posix(kind, library_ext):
     if not (out.startswith('8.0.') or out.startswith('7.0.')
             or out.startswith('7.1.')):
         msg = (
-            "Building llvmlite requires LLVM 7.0+ Be sure to "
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Thanks for the patch. https://github.com/numba/llvmlite/pull/543 Made a similar change, so am going to close. Thanks again!

larsoner:
It might also be worth taking the extra information that was provided by this PR (printing the obtained version), but not a big deal I suppose

stuartarchibald:
@larsoner if you are happy to rebase this on master with such a change I'm happy to review. Thanks.

PR #528, opened 2019-11-05
Add cttz and ctlz to irbuilder docs.

ivirshup:
Added `cttz` intrinsic to sphinx docs.

Should this be kept consistent with the doc-string? I think it would be useful if the doc-string had information about the flag as well.

### DIFF
--- a/docs/source/user-guide/ir/ir-builder.rst
+++ b/docs/source/user-guide/ir/ir-builder.rst
@@ -205,6 +205,11 @@ Integer
 
      Arithmetic, signed, right-shift *lhs* by *rhs* bits.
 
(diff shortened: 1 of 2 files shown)


esc:
@ivirshup consistency is a good value to code by, so I would say, yes.

ivirshup:
👍

ivirshup:
I've also added docs for `ctlz`, since that also is used in `numba.unsafe.numbers.leading_zeros`. (part of: https://github.com/numba/numba/pull/4780)

esc:
@ivirshup thanks for submitting this, I think it is ready for merge.

PR #532, opened 2019-12-05
Only use pkg_resources to load library if other methods fail

cgohlke:
Delay import of pkg_resources because it is excruciatingly slow.
See https://github.com/pypa/setuptools/issues/510

On my Windows workstation, `import llvmlite.binding` takes 3.2 s, about 3 s of which are spent in `import pkg_resources`.

### DIFF
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -125,25 +125,12 @@ if os.name == 'nt':
 _lib_name = get_library_name()
 
 
-# Possible CDLL loading paths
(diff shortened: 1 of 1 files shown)


seibert:
Would it make sense for us to switch to `importlib_resources` instead?

gmarkall:
#801 switches to using importlib instead - that PR is presently active - should we perhaps close this one?

gmarkall:
Many thanks @cgohlke !

PR #549, opened 2020-02-05
Miscalled method and missing parameter in the documentation

nofiv:
The method in IRBuilder is called branch_indirect, not indirectbr (https://github.com/numba/llvmlite/blob/master/llvmlite/ir/builder.py#L803)


esc:
@nofiv thanks for submitting this, does this only concern the documentation?

nofiv:
I've noticed later that there's a missing parameter in an example. Yes, it's related only to the documentation, the example as well.

esc:
@nofiv thanks it LGTM, @sklam ?

stuartarchibald:
This looks good, thanks for the fix. I think the original `indirectbr` error comes from using the `__repr__` of an `instructions.IndirectBranch` created via the `builder` API function fixed in this PR.

PR #554, opened 2020-03-16
azure-pipelines: Bump VM images.

jvesely:
The old ones will be removed 03/23/2020.
Allow Visual Studio 15 2017 as cmake generator.

Signed-off-by: Jan Vesely <jano.vesely@gmail.com>

PR #560, opened 2020-03-23
ENH: Better error message

larsoner:
Rebase of #526 after #543 was merged.

### DIFF
--- a/ffi/build.py
+++ b/ffi/build.py
@@ -112,11 +112,10 @@ def main_posix(kind, library_ext):
     if not (out.startswith('8.0.') or out.startswith('7.0.')
             or out.startswith('7.1.')):
         msg = (
-            "Building llvmlite requires LLVM 7.0.x, 7.1.x or 8.0.x, got %r. "
+            "Building llvmlite requires LLVM 7.0.x, 7.1.x or 8.0.x, got {!r}. "
             "Be sure to set LLVM_CONFIG to the right executable path.\n"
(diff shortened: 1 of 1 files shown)


ffi/build.py
```diff
@@ -112,11 +112,11 @@ def main_posix(kind, library_ext):
...
             or out.startswith('7.1.')):
         msg = (
-            "Building llvmlite requires LLVM 7.0.x, 7.1.x or 8.0.x Be sure to "
-            "set LLVM_CONFIG to the right executable path.\n"
+            "Building llvmlite requires LLVM 7.0.x, 7.1.x or 8.0.x, got %r. "
+            "Be sure to set LLVM_CONFIG to the right executable path.\n"
             "Read the documentation at http://llvmlite.pydata.org/ for more "
             "information about building llvmlite.\n"
-            )
+            % (out.strip(),))
```
stuartarchibald: Need this be a tuple?

stuartarchibald: Also, perhaps use new style formatting?

stuartarchibald:
CI failures probably require merge of https://github.com/numba/llvmlite/pull/554

ffi/build.py
larsoner: > Need this be a tuple?

No -- in general I find it more future compatible to make it one in case other things need to be added in the future, but I'll revert

> Also, perhaps use new style formatting?

Sure, I'll change it. I was trying to conform to the local file convention since `format` is unused and `%` used 5 times, but happy either way

PR #565, opened 2020-03-28
Set AppVeyor to python 3.6

ARF1:
AppVeyor does not currently support python 3.8. This sets it to the minimum requirement of python 3.6.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,7 +8,7 @@ environment:
 
   matrix:
     # Since appveyor is quite slow, we only use a single configuration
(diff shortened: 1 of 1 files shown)


esc:
Dear @ARF1 , thank you again for submitting this and apologies that it was never processed! We actually moved from Appveyor to Azure Pipelines sometime last year so this change is now no longer needed. In fact, we just removed the Appveyor config with #687 so I will close this PR now. Thank you again for your efforts and sorry again that this patch never made it.

PR #568, opened 2020-03-31
Remove unused ninja build requirements on windows

ARF1:
Ninja is not used. It gets downloaded and installed needlessly.

Also, the ninja cmake generator has issues. It does not play nice with Visual Studio Build tools. Using the Visual Studio generator is more reliable in my experience.

### DIFF
--- a/conda-recipes/llvmdev/meta.yaml
+++ b/conda-recipes/llvmdev/meta.yaml
@@ -65,8 +65,6 @@ requirements:
     - cmake # [not (armv6l or armv7l)]
     # Needed to unpack the source tarball
     - m2w64-xz  # [py27 and win]
-    # ninja not currently used, bld.bat needs an update
-    - ninja  # [win]
     # Needed to build LLVM
     - python >=3
     # need vs2015_runtime to build, do not want it at run time


esc:
This looks good, are the passing CI tests a good enough indication that this doesn't break anything?

ARF1:
@esc For what it's worth, it builds on my machine without `ninja`.

stuartarchibald:
> This looks good, are the passing CI tests a good enough indication that this doesn't break anything?

Public CI does not test conda builds of llvm, or indeed any sort of build of llvm.

esc:
Maybe this should go through the (internal) build farm?

stuartarchibald:
I think that changing the recipe would technically require a build number bump and rebuilding everything, for a noncritical change I'd suggest delaying it and including instead in the LLVM 9 upgrade. An alternative to this would be that it might be good to get ninja working again!

ARF1:
@stuartarchibald No problem with waiting for llvm 9. I am happy to carry my llvmdev patches locally.

Re "getting ninja working again": I could never get the ninja cmake generator to play nice with Visual Studio Build Tools: With VS Build Tools 2017 one can install multiple toolset, i.e. v140 and v141. Ninja will never find v140 and error out if v141 is not installed. (And I think a similar problem exists for Build Tools 2019 with v141 / v142).

With the Visual Studio cmake generators there is no problem, they just use whatever toolset is available without fuss - with the option of specifying which toolset to use.

Maybe there are ways of making this work with ninja but I never could.

esc:
I discovered today, that `conda-forge` feedstock for `llvmdev` uses `ninja`:

https://github.com/conda-forge/llvmdev-feedstock/blob/master/recipe/bld.bat#L11

esc:
@ARF1 thank you again for submitting this and apologies for the delay in processsing this. The changes have now been folded into #670 and so I will close this as it has been superseded.

PR #576, opened 2020-04-15
Update the release date

stuartarchibald:
As title.


stuartarchibald:
Same as https://github.com/numba/llvmlite/pull/575 but targets branch https://github.com/numba/llvmlite/tree/release0.32

sklam:
need to update this to mention https://github.com/numba/llvmlite/pull/577

PR #582, opened 2020-04-20
Adds override for LLVM version check, re-formats docs.

stuartarchibald:
Builds on: https://github.com/numba/llvmlite/pull/581

Adds override for LLVM version check, re-formats docs.

* Adds environment variable to remove LLVM version check.
* Reformats install guide doc:
  * 80 char limit
  * Move static linking prose to FAQ
  * Add section on using unsupported LLVMs


docs/source/admin-guide/install.rst
```diff
@@ -71,34 +71,46 @@ prebuilt LLVM binary package and skip this step::
 
 The LLVM build process is fully scripted by conda-build_, and the `llvmdev recipe <https://github.com/numba/llvmlite/tree/master/conda-recipes/llvmdev>`_ is the canonical reference for building LLVM for llvmlite.  Please use it if at all possible!
```
esc, line 72: Not part of this PR, but this line would be wrapped.

PR #580, opened 2020-04-21
Trove classifiers may be out of date.

esc:
https://github.com/numba/llvmlite/blob/master/setup.py#L181-L183

It should probably read: `3.6 3.7 3.8`. I can prep a PR if deemed correct.


esc:
I have converted the issue to a pull-request and it is now read for review.

esc:
I should also note that, the `setup.py` still contains some flake8 issues that need to be addressed in a subsequent PR.

jccurtis:
Excellent, thanks folks! I was just about to make this PR. This causes the release on https://pypi.org/project/llvmlite/ to let `python2.7` think it can use it. I know this is deprecated but it has caused some confusing CI failures since LLVM is not installed in the CI chain. Thanks @esc and @stuartarchibald. Is there a plan to bump the version on pypi?

PR #585, opened 2020-04-30
Update changelog for 0.32.1 on release branch

sklam:
as titled.

like https://github.com/numba/llvmlite/pull/586 but on release branch

### DIFF
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1,5 +1,5 @@
 v0.32.1 (May 7, 2020)
-----------------------
+---------------------
 
 This is a small patch release that addresses some packaging issues:
 


CHANGE_LOG
```diff
@@ -1,3 +1,20 @@
+v0.32.1 (May 7, 2020)
+----------------------
```
stuartarchibald, line 2: Just noticed... this is too many dashes?

stuartarchibald, line 2:
```suggestion
v0.32.1 (May 7, 2020)
---------------------
```

stuartarchibald:
Thanks for the fix, looks good.

PR #586, opened 2020-04-30
Update changelog for 0.32.1 on master

sklam:
as titled.

like https://github.com/numba/llvmlite/pull/585 but on master branch

### DIFF
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -5,7 +5,7 @@ In development
 
 
 v0.32.1 (May 7, 2020)
-----------------------
+---------------------
 
 This is a small patch release that addresses some packaging issues:
 


CHANGE_LOG
```diff
@@ -3,6 +3,24 @@ v0.33.0dev
 
 In development
 
+
+v0.32.1 (May 7, 2020)
+----------------------
```
stuartarchibald, line 8: Just noticed... this is too many dashes?

stuartarchibald, line 8:
```suggestion
v0.32.1 (May 7, 2020)
---------------------
```

stuartarchibald:
Thanks for the fix, looks good.

PR #590, opened 2020-05-21
add python_requires to setup.py

esc:
As of version 0.32.0 (?) the lllvmlite package only supports Python 3.6
and beyond. As outlines in
https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires
- the appropriate line is added to the `setup.py` file to prevent `pip`
from installing the project on other Python versions.


esc:
In case anyone is wondering, the Numba `setup.py`  features this already, I just checked.

PR #593, opened 2020-06-02
Fix CUDA with LLVM9

gmarkall:
The `llvm.nvvm.atomic.load.add` intrinsics were removed and `atomicrmw` is used in its place. The auto-upgrade of IR replaces any of the intrinsics with `atomicrmw`, which is a problem when sending the optimized IR to NVVM, which doesn't accept `atomicrmw`.

This commit adds patches to revert the removal of these intrinsics and corresponding auto-upgrade of the IR.


sklam:
I'm doing a quick test for this patch in our buildfarm.

sklam:
I confirm that the build is working fine in our buildfarm and it is resolving the problems with atomics

PR #594, opened 2020-06-03
Update changelog for 0.33.0

sklam:
as titled

### DIFF
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1,5 +1,5 @@
-v0.33.0rc1
-----------
+v0.33.0 (June 10, 2020)
+-----------------------
 
 This release upgrades to LLVM 9 and drops support for older LLVM versions.
 
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1,7 +1,33 @@
-v0.33.0dev
+v0.33.0rc1
 ----------
 
-In development
+This release upgrades to LLVM 9 and drops support for older LLVM versions.
+
+Pull requests:
+
+* PR #593: Fix CUDA with LLVM9
+* PR #592: Fix meta.yaml
+* PR #591: buildscripts: Unpin wheel
+* PR #590: add python_requires to setup.py
+* PR #586: Update changelog for 0.32.1 on master
+* PR #585: Update changelog for 0.32.1 on release branch
+* PR #580: Trove classifiers may be out of date.
+* PR #582: Adds override for LLVM version check, re-formats docs.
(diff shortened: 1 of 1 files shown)


CHANGE_LOG
```diff
@@ -1,7 +1,33 @@
...
+This release upgrades to LLVM 9 and drops support for older LLVM versions.
+
+Pull requests:
+
+* PR #593: Fix CUDA with LLVM9
+* PR #592: Fix meta.yaml
+* PR #591: buildscripts: Unpin wheel
+* PR #590: add python_requires to setup.py
+* PR #586: Update changelog for 0.32.1 on master
+* PR #585: Update changelog for 0.32.1 on release branch
```
stuartarchibald, line 13: Do these need to go in ? I removed similar from the Numba change log.

CHANGE_LOG
```diff
@@ -1,7 +1,33 @@
...
+* PR #580: Trove classifiers may be out of date.
+* PR #582: Adds override for LLVM version check, re-formats docs.
+* PR #581: Add FAQ entry on LLVM version support.
+* PR #577: llvmlite wheel building fixes
+* PR #576: Update the release date
```
stuartarchibald, line 18: Is this on master?

CHANGE_LOG
```diff
@@ -1,7 +1,33 @@
...
+
+Pull requests:
+
+* PR #593: Fix CUDA with LLVM9
+* PR #592: Fix meta.yaml
+* PR #591: buildscripts: Unpin wheel
+* PR #590: add python_requires to setup.py
+* PR #586: Update changelog for 0.32.1 on master
+* PR #585: Update changelog for 0.32.1 on release branch
+* PR #580: Trove classifiers may be out of date.
```
stuartarchibald, line 14: This should be 2 lines down, after #581

CHANGE_LOG
sklam: right, they don't need to

CHANGE_LOG
sklam: good catch.

CHANGE_LOG
sklam: i guess my script is going by some sort of dates

CHANGE_LOG
```diff
@@ -1,7 +1,30 @@
-v0.33.0dev
+v0.33.0rc1
```
stuartarchibald, line 1: I think this has the proposed release date in it now and no `rc1`?

PR #596, opened 2020-06-11
Use std::make_unique on LLVM 10

mgorny:
LLVM 10 removes llvm::make_unique in favor of std::make_unique.
However, this requires C++14 and is therefore unsuitable for LLVM 9
that forces -std=c++11.  Update the code to use both conditionally.
This fixes all issues with LLVM 10.


esc:
@mgorny thanks very much for your contribution. I cherry-picked the commit into this PR here: https://github.com/numba/llvmlite/pull/599

esc:
I will close this issue now, as the suggested change has been cherry-picked into another PR which is likely to make it into `master`(`main`)  soon.

PR #602, opened 2020-06-25
add missing targets to wheel build

### LINKED ISSUE #601: nvptx target missing in 0.33 when installed using pip
Hi, has the PTX target been removed in 0.33?

Trying to use target triple `nvptx64-nvidia-cuda` results in:
```
cls = <class 'llvmlite.binding.targets.Target'>, triple = 'nvptx64-nvidia-cuda'

    @classmethod
    def from_triple(cls, triple):
        """
        Create a Target instance for the given triple (a string).
        """
        with ffi.OutputString() as outerr:
            target = ffi.lib.LLVMPY_GetTargetFromTriple(triple.encode('utf8'),
                                                        outerr)
            if not target:
>               raise RuntimeError(str(outerr))
E               RuntimeError: No available targets are compatible with triple "nvptx64-nvidia-cuda"

../.local/lib/python3.7/site-packages/llvmlite/binding/targets.py:197: RuntimeError
```

Reverting to 0.32.1 fixes the issue and my program runs without issues.
Rebuilding 0.33.0 from source using system LLVM also works without any changes to the original program.

esc:
Fixes #601: add the missing GPU hosts to the wheel build for llvmdev.

### DIFF
--- a/conda-recipes/llvmdev_manylinux1/build.sh
+++ b/conda-recipes/llvmdev_manylinux1/build.sh
@@ -28,7 +28,8 @@ _cmake_config+=(-DHAVE_TERMIOS_H=OFF)
 _cmake_config+=(-DCLANG_ENABLE_LIBXML=OFF)
 _cmake_config+=(-DLIBOMP_INSTALL_ALIASES=OFF)
 _cmake_config+=(-DLLVM_ENABLE_RTTI=OFF)
-_cmake_config+=(-DLLVM_TARGETS_TO_BUILD=host)
+_cmake_config+=(-DLLVM_TARGETS_TO_BUILD="host;AMDGPU;NVPTX")
+_cmake_config+=(-DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=WebAssembly)
 _cmake_config+=(-DLLVM_INCLUDE_UTILS=ON) # for llvm-lit
 # TODO :: It would be nice if we had a cross-ecosystem 'BUILD_TIME_LIMITED' env var we could use to
 #         disable these unnecessary but useful things.


stuartarchibald:
Perhaps roll this into the LLVM 10 patch series unless you were considering rebuilding the entire Numba stack wheels as an option?

jvesely:
Will this be only in 0.34+? Or do you plan to release 0.33.1?

esc:
@stuartarchibald @jvesely perhaps this should be a topic of discussion during the next core-dev meeting? @jvesely what would be your preference?

jvesely:
> @stuartarchibald @jvesely perhaps this should be a topic of discussion during the next core-dev meeting? @jvesely what would be your preference?

I don't really have one. I can keep the affected machines on 0.32.1 for the time being.
In general the sooner, the better. I might also be worth considering not having 0.33.x as a broken release wrt nvptx/amdgpu.

esc:
The decision in the core-dev meeting yesterday was, that we fix the wheels on PyPi but not do a new release.

sklam:
To test, we can add use:

```bash
$ llvm-config --targets-built
AMDGPU NVPTX WebAssembly X86
```

in the recipe's test section to check for NVPTX in the output of `llvm-config`

esc:
@jvesely sorry for being late on the update of this. I ran out of time to produce new wheels for 0.33.x -- however, the 0.34.0rc1 wheels (available since last week) should contain this fix already.

jvesely:
> @jvesely sorry for being late on the update of this. I ran out of time to produce new wheels for 0.33.x -- however, the 0.34.0rc1 wheels (available since last week) should contain this fix already.

thanks! I can confirm that 0.34.0rc1 works OK, at least for cuda.

esc:
@jvesely excellent, thanks for following up!

Issue #553, opened 2020-02-27
Build issue with LLVM-8.0.1 & llvmlite-0.31.0: 'libllvmlite.so' failed


openedev:
Hi,

I found a build issue with LLVM-8.0.1 for llvmlite-0.31.1 release. I have built the LVMM-8.0.1 and link to llvmlite using buildroot environment. Let me know if I miss anything.

Build Log:
[...]

Issue #542, opened 2020-01-04
Build breaks on FreeBSD: it needs -fPIC


yurivict:
```
got version from file /wrkdirs/usr/ports/devel/py-llvmlite/work-py36/llvmlite-0.31.0/llvmlite/_version.py {'version': '0.31.0', 'full': 'fe7d985f6421d87f613bd414479d29d912771562'}
running build_ext
/usr/local/bin/python3.6 /wrkdirs/usr/ports/devel/py-llvmlite/work-py36/llvmlite-0.31.0/ffi/build.py
LLVM version... clang++ -std=c++11 -stdlib=libc++ -shared -I/usr/local/llvm80/include -std=c++11 -fno-exceptions -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -fno-rtti -g assembly.cpp bitcode.cpp core.cpp initfini.cpp module.cpp value.cpp  executionengine.cpp transforms.cpp passmanagers.cpp targets.cpp dylib.cpp  linker.cpp object_file.cpp -o libllvmlite.so -L/usr/local/llvm80/lib -lLLVM-8
ld: error: can't create dynamic relocation R_X86_64_64 against local symbol in readonly segment; recompile object files with -fPIC or pass '-Wl,-z,notext' to allow text relocations in the output
```

using cmake to build

llvmlite-0.31.0

stuartarchibald:
Thanks for the report. I'm not sure what the build target is from the above (wheel-like?!) if so perhaps add `pic = sys.platform.startswith(('linux', 'freebsd','openbsd'))` here: https://github.com/numba/llvmlite/blob/eaa3412a1f2bbd7a42b42555bbbce865f0f6cd23/setup.py#L130 ?
if that fixes it please let us know, thanks.

nehaljwani:
I was trying to build llvmlite-v0.30.0 with a build of llvmdev-v8.0.0 (both built using devtoolset-2) and faced the same error. I ended up exporting `CXXFLAGS="${CXXFLAGS-} -fPIC"`

I am not sure why you are doing it only for building a wheel. In both cases you want to shove object files into a shared library, so they need to have position independent code (at least on Linux).

You do not see this problem with the conda recipe because the activation scripts for the compilers inject `-fPIC` by default:
[...]

Issue #522, opened 2019-10-10
Build error (symbol relocation) when installing from source


trbabb:
I've attempted to build the llvmlite from source on an Ubuntu 18.04 instance. I'm building outside of Conda, following the directions in the docs— I've built llvm 8.0 with the supplied patches, which succeeded (but only after doing `export RECIPE_DIR=/path/to/llvmlite/conda-recipes/llvmdev`— that probably ought to be in the docs). 

I'm avoiding Conda because I need to deploy to virtualenv (many of them per machine), and I'm not familiar enough with Conda to figure out how to make them play together (assuming it's possible); the Conda docs seem to suggest that attempting this is brittle.

With `export LLVM_CONFIG=/path/to/my/patched/llvm8/bin/llvm-config` running either `python3 setup.py build` or `pip3 install /path/to/local/llvmlite` (the latter being the deploy path for virtualenv) will fail with:

```
[...]

PR #627, opened 2020-08-21
Always build with -fPIC on Linux and BSD

gmarkall:
**EDIT:** Title changed to reflect the fact that this PR now always builds with `-fPIC` on Linux and BSD, instead of requiring the user to specify it as an option. Original PR comment follows:

On systems that do not build PIC by default, linking libllvmlite.so fails (see e.g. the error reported in #553). This PR adds an option to the `build_ext` command in setup.py (`--pic`) to compile with `-fPIC`.

Related issues: #553, #542, #522 - this PR may address some of these issues.

In my case, I need to use this option when building on Ubuntu 18.04 with the system compiler, but from the above issues it seems that there are other systems that require the flag as well.


esc:
@gmarkall thanks for submitting this, I have added it to the queue for review.

ffi/Makefile.freebsd
```diff
@@ -1,6 +1,7 @@
 
 CXX = clang++ -std=c++11 -stdlib=libc++
-CXXFLAGS = $(LLVM_CXXFLAGS)
+CXXFLAGS := $(CPPFLAGS) $(CXXFLAGS) $(LLVM_CXXFLAGS) $(CXX_FLTO_FLAGS)
+LDFLAGS := $(LDFLAGS) $(LLVM_LDFLAGS) $(LD_FLTO_FLAGS)
 LDFLAGS = $(LLVM_LDFLAGS)
```
stuartarchibald, line 5: Suggest removing this as its overwriting the line above.

ffi/Makefile.freebsd
```diff
@@ -1,6 +1,7 @@
 
 CXX = clang++ -std=c++11 -stdlib=libc++
-CXXFLAGS = $(LLVM_CXXFLAGS)
+CXXFLAGS := $(CPPFLAGS) $(CXXFLAGS) $(LLVM_CXXFLAGS) $(CXX_FLTO_FLAGS)
+LDFLAGS := $(LDFLAGS) $(LLVM_LDFLAGS) $(LD_FLTO_FLAGS)
```
stuartarchibald, line 4: Probably worth adding https://github.com/numba/llvmlite/blob/4acef2d5165768f43b315dd47ed99a5f8f63b3a8/ffi/Makefile.linux#L5-L6 above too.

ffi/Makefile.freebsd
gmarkall: Will remove.

stuartarchibald (approved):
Thanks for the fixes. Patch looks good, also tests out ok on OpenBSD.

stuartarchibald:
@esc any chance you could trigger a test wheel build of this please?

esc:
BFID: `llvmlite_wheel_13`

esc:
So,  the wheel-build was fine for all architectures, i.e. build artifacts were produced. We currently don't have any automation however to actually run tests on the wheels.

esc:
I spot checked OSX and Linux x86_64  and both are fine.

stuartarchibald:
> I spot checked OSX and Linux x86_64 and both are fine.

Thanks for building and testing @esc, think this is good to go.

esc:
Agreed!

PR #635, opened 2020-09-30
Fixes #629: Default to 0 if None is passed as a value for Int32

### LINKED ISSUE #629: Can not use zeroinitializer vector as mask in shuffle_vector
code like this doesn't work, but it should
```python
vector = ir.VectorType(int32, 3)([1, 2, 3])
mask = ir.VectorType(int32, 3)(None)
builder.shuffle_vector(vector, vector, mask)
```
instead we get
```
IndexError: mask values need to be in range(0, 6)
```

NavyaZaveri:
Fixes #629.

### DIFF
--- a/llvmlite/ir/types.py
+++ b/llvmlite/ir/types.py
@@ -1,3 +1,4 @@
+
 """
 Classes that are LLVM types
 """
@@ -242,6 +243,11 @@ class IntType(Type):
             return str(val).lower()
         else:
             return str(val)
+
+    def wrap_constant_value(self, value):
+        if value is None:
+            return 0
+        return value
 
     @property
     def intrinsic_name(self):
@@ -513,7 +519,6 @@ class BaseStructType(Aggregate):
     def gep(self, i):
         """
         Resolve the type of the i-th element (for getelementptr lookups).
-
         *i* needs to be a LLVM constant, so that the type can be determined
         at compile-time.
         """
@@ -563,7 +568,6 @@ class IdentifiedStructType(BaseStructType):
     A type which is a named alias for another struct type, akin to a typedef.
     While literal struct types can be structurally equal (see
     LiteralStructType), identified struct types are compared by name.
-
     Do not use this directly.
     """
     null = 'zeroinitializer'
--- a/llvmlite/ir/types.py
+++ b/llvmlite/ir/types.py
@@ -1,4 +1,3 @@
-
 """
 Classes that are LLVM types
 """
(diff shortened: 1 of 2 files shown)


esc:
@NavyaZaveri thanks for submitting this, I have added it to the queue for review.

llvmlite/ir/types.py
```diff
@@ -1,3 +1,4 @@
+
```
stuartarchibald, line 1: Probably don't need these line changes?

llvmlite/ir/types.py
```diff
@@ -513,7 +519,6 @@ def format_constant(self, value):
     def gep(self, i):
         """
         Resolve the type of the i-th element (for getelementptr lookups).
-
```
stuartarchibald, line 516: Probably don't need these line changes?

llvmlite/ir/types.py
```diff
@@ -563,7 +568,6 @@ class IdentifiedStructType(BaseStructType):
     A type which is a named alias for another struct type, akin to a typedef.
     While literal struct types can be structurally equal (see
     LiteralStructType), identified struct types are compared by name.
-
```
stuartarchibald, line 566: Probably don't need these line changes?

llvmlite/ir/types.py
```diff
@@ -243,6 +244,11 @@ def format_constant(self, val):
         else:
             return str(val)
 
+    def wrap_constant_value(self, value):
+        if value is None:
+            return 0
+        return value
+
```
stuartarchibald, line 251: Could a test be added for this please? Thanks.

stuartarchibald (commented):
Thanks for the patch, think this should work, there's a few minor things to resolve else looks good. Thanks again!

NavyaZaveri:
Thanks for the review @stuartarchibald 

Is there an official formatter used for the project? I couldn't find anything in the docs 

It's a bit PITA to losslessly format multiline asm strings for the tests  :(  

Also there is conflicting info on max line [here](https://github.com/numba/llvmlite/blob/master/.landscape.yml) and [here](https://github.com/numba/llvmlite/blob/master/.flake8)

llvmlite/tests/test_ir.py
```diff
@@ -2064,6 +2064,10 @@ def test_vector(self):
             ', '.join(map('i32 {}'.format, vals)))
         self.assertEqual(str(vec), vec_repr)
 
+    def test_non_nullable_int(self):
+        constant = ir.Constant(ir.IntType(32), None).constant
+        assert constant == 0
```
stuartarchibald, line 2069:
```suggestion
        self.assertEqual(constant, 0)
```
`llvmlite` testing uses unittest assertions

stuartarchibald (approved):
Thanks for the patch and fixes, looks good.

PR #626, opened 2020-08-20
Update SVML patch

Hardcode84:
Patch was updated to fix SVML calling convention issues uncovered by llvm 10. In previous versions of patch SVML calling convention was selected based on compilation settings. So if you try to call 256bit vector function from avx512 code function will be called with avx512 cc which is incorrect. To fix this SVML cc was separated into 3 different cc for 128, 256 and 512bit vector lengths which are selected based on actual input vector length.

CC @stuartarchibald

PR #636, opened 2020-10-06
Llvm 10.0.1 1

esc:
Building out `llvmdev` version 10.0.1-1 -- which includes #626


esc:
Since #626 was merged to master and #636 has produced a 10.0.1-1 build, I think this does not need to be merged an can be closed?

esc:
Will close now.

PR #641, opened 2020-10-13
fix llvmdev build number

esc:
The current version of `llvmdev` on `numba/label/main` is 10.0.1-1 -- as
this was build from the PR at:

https://github.com/numba/llvmlite/pull/636

To sync the state of the repository with the state of the build
artifacts, we must increment the llvmdev build number.


stuartarchibald (approved):
Thanks for the patch. This synchronizes with what is built.

PR #642, opened 2020-10-13
pin to correct version of llvmdev

esc:
llvmlite will need to be built against the correct version and build
number of llvmdev. This pins that version number as part of the
`conda-recpie`.


stuartarchibald:
Buildfarm ID: `llvmlite_57`

stuartarchibald:
> Buildfarm ID: `llvmlite_57`

Apart from known issues, this passed.

stuartarchibald:
Spot check in build logs suggests build 1 is being picked up.

stuartarchibald (approved):

PR #643, opened 2020-10-13
Update changelog for 0.35.0rc1

sklam:
as titled.

### DIFF
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1,16 +1,14 @@
-v0.35.0rc1
-----------
+v0.35.0rc1 (Oct 14, 2020)
+-------------------------
 
 The main feature in this release is a Numba specific LLVM pass for pruning
 reference-count operations. We plan to generalize this custom LLVM pass
 once it is proven stable so that it can be configured for other uses.
 In addition, this release contains an updated SVML patch that fixes an issue
-for AVX512.
+for AVX512, and a patch that fixes build issues on Linux and BSD.
 
 Pull requests:
 
-* PR #614: Fix wheel building.
-* PR #616: fix release date in changelog
 * PR #617: Fix wheel building.
 * PR #618: Fix changelog on master
 * PR #626: Update SVML patch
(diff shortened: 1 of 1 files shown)


esc (approved):
LGMT!

CHANGE_LOG
```diff
@@ -1,7 +1,36 @@
-v0.35.0dev
------------
+v0.35.0rc1
```
stuartarchibald, line 1: Want to put a stub in for the date so we don't forget to add it later?

CHANGE_LOG
```diff
@@ -1,7 +1,36 @@
...
+The main feature in this release is a Numba specific LLVM pass for pruning
+reference-count operations. We plan to generalize this custom LLVM pass
+once it is proven stable so that it can be configured for other uses.
+In addition, this release contains an updated SVML patch that fixes an issue
+for AVX512.
+
+Pull requests:
+
+* PR #614: Fix wheel building.
+* PR #616: fix release date in changelog
```
stuartarchibald, line 13: Is this on mainline?

CHANGE_LOG
```diff
@@ -1,7 +1,36 @@
...
------------
+v0.35.0rc1
+----------
 
-In development
+The main feature in this release is a Numba specific LLVM pass for pruning
+reference-count operations. We plan to generalize this custom LLVM pass
+once it is proven stable so that it can be configured for other uses.
+In addition, this release contains an updated SVML patch that fixes an issue
+for AVX512.
```
stuartarchibald, line 8: It also sorts out issues with fPIC and support for BSD.

stuartarchibald (commented):
Few minor comments else looks good, thanks.

CHANGE_LOG
sklam: good catch

CHANGE_LOG
sklam: I'll add that

CHANGE_LOG
sklam: i'll also need to update my script to check for the target branch

stuartarchibald (approved):

PR #647, opened 2020-11-06
Add support for general attrs on call and loop-rotate pass

stuartarchibald:
As title.


sklam (approved):
Thanks for the patch. LGTM

PR #655, opened 2020-11-19
Switch encoding to UTF-8 from latin1

gmarkall:
Opening this PR to see what happens on CI.

This change was originally made in PR #53, but may no longer be required (and may cause issues with comments in IR that use non-latin1 characters).

### DIFF
--- a/llvmlite/binding/common.py
+++ b/llvmlite/binding/common.py
@@ -2,12 +2,12 @@ import atexit
 
 
 def _encode_string(s):
-    encoded = s.encode('latin1')
+    encoded = s.encode('utf-8')
     return encoded
 
 
 def _decode_string(b):
-    return b.decode('latin1')
+    return b.decode('utf-8')
 
 
 _encode_string.__doc__ = """Encode a string for use by LLVM."""


gmarkall:
@sklam What do you think of this? Do you think the original reason you needed to switch to `latin1` was due to some LLVM bug that's now fixed?

gmarkall:
I've also run the Numba tests locally on Windows with this change, and it did not seem to cause any issues.

sklam:
Might need to check NVVM can handle the char sequence https://github.com/numba/llvmlite/pull/53/files#diff-fa06129bdaa2e360fbc421276bba2cc7270638e8675da33555b97c8e84942e28R864-R877

gmarkall:
NVVM seems to be fine with this - I've added a Numba PR for the test too: https://github.com/numba/numba/pull/6517.

I'm taking this out of WIP and marking for actual review, as it seems not to cause any issues with current LLVM / NVVM versions.

stuartarchibald:
Please remind me, was the conclusion here that this was probably ok? IIRC NVVM was of most concern but that is reported fine by @gmarkall.

Given it's a small patch, perhaps we merge and publish new development builds, in the worse case the revert patch is small.

esc:
> Given it's a small patch, perhaps we merge and publish new development builds, in the worse case the revert patch is small.


@stuartarchibald agreed, let's do it!

stuartarchibald (approved):
Thanks for the patch.

PR #661, opened 2020-12-01
Phi instructions should invalidate their string cache when new incomings are added

trbabb:
Stringifying a phi (including for debugging purposes) can cause them to freeze themselves in a stale state, and subsequently added incoming instructions will be incorrectly omitted from the instruction.

Any time a phi is mutated by adding an incoming expression, its string cache should be immediately invalidated.

### DIFF
--- a/llvmlite/ir/instructions.py
+++ b/llvmlite/ir/instructions.py
@@ -544,6 +544,7 @@ class PhiInstr(Instruction):
     def add_incoming(self, value, block):
         assert isinstance(block, Block)
         self.incomings.append((value, block))
+        self._clear_string_cache()
 
     def replace_usage(self, old, new):
         self.incomings = [((new if val is old else val), blk)


jceipek:
Here's a concrete example of where this could cause a problem:

```python
print("before:", entry_phi) # before: %".7" = phi i64 [5, %"entry"]
entry_phi.add_incoming(v, body_blk) # This isn't represented in the final output because the line above populates the cache
print("after:", entry_phi) # after: %".7" = phi i64 [5, %"entry"]
```

Leads to this compile error:
```python
compile error
  File "./codexec", line 169, in handle_object_code
    self.module.verify()
  File "/opt/venv/lib/python3.8/site-packages/llvmlite/binding/module.py", line 115, in verify
    raise RuntimeError(str(outmsg))
RuntimeError: PHINode should have one entry for each predecessor of its parent basic block!
  %.7 = phi i64 [ 5, %entry ]
```

esc:
@trbabb thank you for submitting this. I have added it to the queue for review.

trbabb:
Any chance this and #710 could make it into the next release? It'd really help clean up some workarounds! Thanks!

gmarkall (changes requested):
@trbabb Many thanks for the PR, and apologies for the delay in its review. I'm spending some time going through the llvmlite PR backlog so I'm hoping to help move this forward now.

The changes seem like a good idea in general - as a general question (not necessarily needing to be addressed in this PR) are there other instructions that should invalidate their string cache when modified?

Could you please add a test of the modified functionality? (Perhaps based on @jceipek's example above).

PR #666, opened 2020-12-10
fix rst syntax in install docs

esc:
Some missing whitespace and incorrect indentation caused the
installation instructuions to be malformed.

Before:

Screen Shot 2020-12-10 at 14 28 38

After:

Screen Shot 2020-12-10 at 14 29 39


stuartarchibald (approved):
Thanks for the patch. Can confirm that this renders correctly.

PR #570, opened 2020-04-01
Improvements to llvmdev build scripts

ARF1:
Improvements:
- allow setting the targets to be built with the environment variable `LLVM_TARGETS_TO_BUILD`
- allow building on all compatible Visual Studio versions
- quote path strings to fix issues with paths with spaces

PR #668, opened 2020-12-11
Modify cmake options to work with VS2019

seibert:


seibert:
checking on build farm: llvmlite_66

esc:
`llvmlite_66` passed 🎉

conda-recipes/llvmlite/bld.bat
```diff
@@ -1,12 +1,22 @@
...
+    @rem VS2017
+    @rem set CMAKE_GENERATOR_ARCH=Win64
+    set CMAKE_GENERATOR_ARCH=x64
 )
-set CMAKE_GENERATOR=Visual Studio 15 2017 %CMAKE_GENERATOR_ARCH%
+@rem for older VS:
+@rem set CMAKE_GENERATOR=Visual Studio 15 2017
+@rem do not set CMAKE_GENERATOR_TOOLKIT
+set CMAKE_GENERATOR=Visual Studio 16 2019
+set CMAKE_GENERATOR_TOOLKIT=v141
```
stuartarchibald, line 19: Does this now mean llvmlite cannot be built in a conda-recipe with VS 2017?

ffi/build.py
```diff
@@ -45,35 +52,34 @@ def run_llvm_config(llvm_config, args):
...
-            return True  # keep those we don't recognize
-        ver = int(m.group(1))
-        return ver >= 14
-    generators = list(filter(drop_old_vs, generators))
-
-    generators.append('Visual Studio 15 2017' + (' Win64' if is_64bit else ''))
+        arch = os.environ.get("CMAKE_GENERATOR_ARCH", None)
+        toolkit = os.environ.get("CMAKE_GENERATOR_TOOLKIT", None)
+        generators.append(
+            (os.environ.get("CMAKE_GENERATOR"), arch, toolkit)
```
stuartarchibald, line 69: This branch is taken as `os.environ.get("CMAKE_GENERATOR")` is true-ish, perhaps store as a var and reuse?

stuartarchibald (commented):
Thanks for the patch, glad to see this working! Couple of minor things to resolve else good to go.

conda-recipes/llvmlite/bld.bat
seibert: build.py will always try VS2019 toolkit v141 and then VS2017 if the generator select by the environment variables fails.  (See generators.extend() line later in patch.)

stuartarchibald: Found it, thanks.

stuartarchibald (approved):
Thanks for the fixes, looks good.

sklam:
Two concerns:
- Do we need to sync up the llvmdev recipe to use VS2019 [instead of VS2017](https://github.com/seibert/llvmlite/blob/49fda327d9805690610f8f5230f0c63561e49838/conda-recipes/llvmdev/bld.bat#L7)? 
- Do the llvmlite meta.yaml needs to remove [this line](https://github.com/seibert/llvmlite/blob/49fda327d9805690610f8f5230f0c63561e49838/conda-recipes/llvmlite/meta.yaml#L41)?

seibert:
To the first question, yes.  We should review and merge #570, which makes equivalent fixes to the llvmdev script for VS2019.  (It has to do fancy things in cmd.exe since there is no equivalent to build.py to put the toolkit searching logic into.)

As for the second question, I don't know.  There is no conda package for anything after `vs2015_runtime`, so I assume it continues to work for later VS releases.  I'm assuming things are OK since all the tests pass.

seibert:
note that in terms of binary compatibility, this PR is continuing to use the same compilers as VS2017 on VS2019 by setting the toolkit, so it should not matter whether we build with VS2017 or VS2019.

esc:
Running again at: `llvmlite_win_only_03`.

esc:
`llvmlite_win_only_03` passed.

sklam:
> this PR is continuing to use the same compilers as VS2017 on VS2019 by setting the toolkit, so it should not matter whether we build with VS2017 or VS2019.

I have memory of how 141 compiler on 2019 and 2017 are not binary compatible. That's why I stopped updating everything to VS2019 back then.

esc:
Should we perhaps make a package from these changes and then give it a spin through the Numba test-suite?

sklam:
Yes, we should

esc (approved):
Tested on the farm, builds fine on new Windows workers.

PR #676, opened 2020-12-19
removed misleading sentence

abebeos:
Section is about "compiling LLVM", so going straight to the point.

### DIFF
--- a/docs/source/admin-guide/install.rst
+++ b/docs/source/admin-guide/install.rst
@@ -66,8 +66,6 @@ Compiling LLVM
 
 The LLVM build process is fully scripted by conda-build_, and the `llvmdev recipe <https://github.com/numba/llvmlite/tree/master/conda-recipes/llvmdev>`_ is the canonical reference for building LLVM for llvmlite.  Please use it if at all possible!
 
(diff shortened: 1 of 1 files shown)


esc:
@abebeos thank you for submitting this. I rendered the documentation just now and it looks like:

Screen Shot 2021-01-07 at 12 12 13

It seems the command `conda install -c numba llvmdev` is now out of context? Or did I miss something?

abebeos:
[...]
> It seems the command `conda install -c numba llvmdev` is now out of context? Or did I miss something?

yes. actually now redundant, so removed.

esc:
@abebeos thank you for fixing this up. @stuartarchibald @seibert @sklam what are your thoughts on this. Do you think we should perhaps move the hint elsewhere instead of deleting it outright?

abebeos:
I've now understood that I had... misunderstood.

Closing, as it is in essence correct (should just be rewritten at some point, at this moment I've no suggestions on how to achieve better clarity).

esc:
@abebeos thank you for following up. I agree that the overall structure isn't entirely favorable but I also don't quite know yet how to improve it.

PR #678, opened 2020-12-21
Don't need to use inner context

dbwodlf3:
in llvmlite/binding/module.py

I can find 'context' in constructor of ModuleRef, and it use only one function.
ffi.lib.LLVMPY_CloneModule
but ffi.lib.LLVMPY_CloneModule , don't use context.

https://github.com/numba/llvmlite/blob/master/ffi/module.cpp#L281

it use only ModuleRef.

thanks.

### DIFF
--- a/llvmlite/binding/module.py
+++ b/llvmlite/binding/module.py
@@ -18,8 +18,7 @@ def parse_assembly(llvmir, context=None):
     strbuf = c_char_p(llvmir)
     with ffi.OutputString() as errmsg:
         mod = ModuleRef(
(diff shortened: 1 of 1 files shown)


esc:
@dbwodlf3 thank you for submitting this. May I ask why this was closed?

dbwodlf3:
because CI say 'this commit can't build.'. and also I didn't compile by hand. I just edit python script in package. So. i am going to edit src code and try to build. and if it be succeed, then i will submit pr again. thanks.

Issue #632, opened 2020-09-16
ObjectFileRef seem's to return wrong section content


RobinDavid:
While experimenting with llvmlite I found some discrepancies between what is returned by the `data()` function of ``ObjectFileRef`` and the real bytes in the file. Here is an example *(where I use lief as ground truth but other disassemblers
returns the same)*:

```python
import llvmlite.binding
import lief

elf = "7f454c4602010100000000000000000001003e000100000000000000000000000000000000000000e0000000000000000000000040" \
      "000000000040000500010048c1e2204809c24889d048c1c03d4831d048b90120000480001070480fafc80000000000000000000000" \
      "00000000000000000000000000002f0000000400f1ff00000000000000000000000000000000070000001200020000000000000000" \
      "001f00000000000000002e74657874005f5f617279626f002e6e6f74652e474e552d737461636b002e737472746162002e73796d74" \
      "6162003c737472696e673e000000000000000000000000000000000000000000000000000000000000000000000000000000000000" \
      "00000000000000000000000000000000000000000000001f0000000300000000000000000000000000000000000000a80000000000" \
      "0000380000000000000000000000000000000100000000000000000000000000000001000000010000000600000000000000000000" \
      "000000000040000000000000001f000000000000000000000000000000100000000000000000000000000000000f00000001000000" \
      "000000000000000000000000000000005f000000000000000000000000000000000000000000000001000000000000000000000000" \
      "0000002700000002000000000000000000000000000000000000006000000000000000480000000000000001000000020000000800" \
      "0000000000001800000000000000"
raw = bytes.fromhex(elf)

obj = llvmlite.binding.ObjectFileRef.from_data(raw)
for s in obj.sections():
    if s.is_text():
        print(s.data().hex())
        # 48c1e2204809c24889d048c1c03d4831d048b9012000000070706572000000

p = lief.parse(raw)
print(bytes(p.get_section('.text').content).hex())
# 48c1e2204809c24889d048c1c03d4831d048b90120000480001070480fafc8
```

Maybe I am misunderstanding the way ObjectFileRef works but in that case you would be eager to know what I did wrong.

Looking forward forward for your help.
Regards

RobinDavid:
I forgot to mention that I tried with version 0.34.0 of llvmlite

gmarkall:
Indeed, objdump produces similar output to lief:

```
$ objdump -s test.elf 

test.elf:     file format elf64-x86-64

Contents of section .text:
[...]

PR #684, opened 2020-12-30
fixes ppc64le segfault under numba static reloc mode

abebeos:
re numba/numba#6606 (re numba/numba#4026)

This fixes segmentation-faults on ppc64le, when using numba in static reloc_mode.

Patch Source:

https://reviews.llvm.org/rG2812c1515627904e31605bbd4f25a887a1f8eb12

### DIFF
--- a/conda-recipes/llvmdev/meta.yaml
+++ b/conda-recipes/llvmdev/meta.yaml
@@ -45,6 +45,8 @@ source:
     # * https://github.com/conda-forge/llvmdev-feedstock/blob/c706309/recipe/patches/expect-fastmath-entrypoints-in-add-TLI-mappings.ll.patch
     - ../intel-D47188-svml-VF.patch
     - ../expect-fastmath-entrypoints-in-add-TLI-mappings.ll.patch
+    # * https://github.com/numba/numba/issues/6606
+    - ../weak_GV_add_TOC_nop_by_stefanp.patch
     {% endif %}
     # Reverts a patch limiting non-GlobalValue name length
     - ../0001-Revert-Limit-size-of-non-GlobalValue-name.patch
--- a/conda-recipes/weak_GV_add_TOC_nop_by_stefanp.patch
+++ b/conda-recipes/weak_GV_add_TOC_nop_by_stefanp.patch
@@ -1,0 +1,272 @@
+[PowerPC] Fix missing nop after call to weak callee.
+
+2812c1515627
+
+Authored by stefanp on Tue, Dec 8, 6:57 AM.
+
+Description
+
+[PowerPC] Fix missing nop after call to weak callee.
+
+Weak functions can be replaced by other functions at link time. Previously it
+was assumed that no matter what the weak callee function was replaced with it
+would still share the same TOC as the caller. This is no longer true as a weak
+callee with a TOC setup can be replaced by another function that was compiled
+with PC Relative and does not have a TOC at all.
+
+This patch makes sure that all calls to functions defined as weak from a caller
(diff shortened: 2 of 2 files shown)


abebeos:
```sh
Failing Tests (3):
    LLVM-Unit :: Support/./SupportTests/CrashRecoveryTest.Basic
    LLVM-Unit :: Support/./SupportTests/CrashRecoveryTest.DumpStackCleanup
    LLVM :: JitListener/multiple.ll

  Expected Passes    : 16897
  Expected Failures  : 52
  Unsupported Tests  : 18254
  Unexpected Failures: 3
  ```

esc:
@abebeos thank you for reporting this, I have added it to the queue for review.

abebeos:
@esc , note that this affects a quite old bug (https://github.com/numba/numba/issues/4026).

Can i trigger myself somehow your ppc64le tests for this, or is this a manual process on your side?

esc:
@abebeos for now, triggering our internal build farm is a manual process not exposed to the public.

esc:
@abebeos hey, thanks for updating this issue. I was wondering, do you have a reproducer for this on the Numba side? I'd need one of those to verify this patch. Also, please rebase this PR to remove the llvm and lld archives, we do not want to vendor those as part of this repo, thanks!

abebeos:
> 
> 
> @abebeos hey, thanks for updating this issue. I was wondering, do you have a reproducer for this on the Numba side? I'd need one of those to verify this patch. 

Reproduction is described within #6606, i you need something more specific, please let me know.

> Also, please rebase this PR to remove the llvm and lld archives, we do not want to vendor those as part of this repo, thanks!

Oh my, forgot to change the branch during some tests. Will cleanup soon.

abebeos:
I validated the patch on Power8 Hardware (py 3.6, linux, ppc64le): 

* testsuite with numba-pic, 0 failures
* testsuite with numba-static: 0 failures

esc:
@abebeos may I ask, what is `numba-pic` and `numba-static`?

abebeos:
Of course!

This ppc bug happens only when numba uses `reloc_mode=static`. If you have `reloc_mode=pic`, then ppc64le will work fine.

numba 0.52 was set to 'pic' for ppc (see https://github.com/numba/numba/commit/31ddd81d559c4aa79fb25e4c7b45b44451df95d5#diff-807036486bf0f1ad036353b9e9aa94b9644c5de3cb136bffbbaa0792f6be9bdd), leaving kind of a dormant bug in the code.

This fix removes the dormant bug, so numba works again with `reloc_mode=static` with ppc (and continues to work with `pic`, too).

So, in order to test the fix, one must run with the original 0.52 numba (pic), and with the 0.52 switched back to static.

abebeos:
> @abebeos hey, thanks for updating this issue. I was wondering, do you have a reproducer [...]

@esc, "reproducer" = a minimal (python) code which triggers the error, something like this here: https://github.com/numba/llvmlite/issues/632#issue-702534951 ?

esc:
@abebeos exactly, something I can copy into my editor, run and which will expose the bug in question.

abebeos:
Ok, thanks! Is there any config-option (cli, file) to configure numba to use "static" instead of the hardcoded "pic"?

esc:
> Ok, thanks! Is there any config-option (cli, file) to configure numba to use "static" instead of the hardcoded "pic"?

I don't think so, no, it has to be set at compile time.

gmarkall:
Given that this PR comes from a user / repo that no longer exists and has not had any activity for a year, I'm going to close it. If we should continue work in this direction, a new PR should be opened from an extant repository.

PR #689, opened 2021-01-13
fixup azure badge to point at master branch

esc:
As title. Followup from #687


stuartarchibald (approved):
Thanks for the patch.

PR #691, opened 2021-01-15
[WIP] fixes ppc64le segfault numba static reloc mode for llvm-11.0.1

abebeos:
closes numba/numba#6606 (was/re numba/numba#4026)

This fixes segmentation-faults on ppc64le, when using numba in static reloc_mode (this is a latent bug, as numba is currently set to `reloc_mode=pic` for ppc, which works fine).

For a summary elaboration, please see: https://github.com/numba/numba/issues/4026#issuecomment-760883075

Patch Source:

https://reviews.llvm.org/rG2812c1515627904e31605bbd4f25a887a1f8eb12

(similar PR for llvm-10.x: #684)


gmarkall:
Given that this PR comes from a user / repo that no longer exists and has not had any activity for a year, I'm going to close it. If we should continue work in this direction, a new PR should be opened from an extant repository.

Issue #698, opened 2021-02-09
Installation does not check pre-requisites


GBR-613:
About a week ago I tried to install llvmlite using pip, and the installation failed with a red herring message. The real root of the problem was that I used Python v3.9.1 for which the package was not still available.  (I suppose now it may be, and the problem is temporarily not reproduceable.)  

If versions of llvmlite are specific per Python's version, then the installation code should check whether the used version is supported. In such the way Python packages work.

I suppose checking of other pre-requisites may matter too.

gmarkall:
Hi @GBR-613 - many thanks for reporting this issue. Do you have a copy of the error message that you received?

GBR-613:
@gmarkall Yes, I opened issue #697 about it.
The misleading message is an issue itself:
But lack of checking of Python version (at least) is separate issue, too.

esc:
This is probably quite easy to solve by adapting the following from Numba:

https://github.com/numba/numba/blob/master/setup.py#L34
https://github.com/numba/numba/blob/master/setup.py#L411

We should probably get this into 0.36.0RC2, it seems easy enough to fix.

(I do recall, however, that when we did this for Numba, it was not effective, as previous releases did not have this limitation included and therefore `pip` choose to select a previous release instead and again yield a potentially misleading error messages. Having said that, we should do this anyway, and cross any future bridged when we get to them. Probably when Python 3.10 is released later on in the year.)

esc:
I have made a corresponding PR here: https://github.com/numba/llvmlite/pull/699

esc:
#699 is now merged, so we can close this. We will see if it helps when Python 3.10 comes out in autumn.

PR #699, opened 2021-02-09
adapt Python version clamp from Numba

esc:
This adds code to limit the Python versions for which llvmlite can be installed. Code is adapted from the Numba project.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,39 @@ from distutils.spawn import spawn
 import os
 import sys
 
+
+_version_module = None
+try:
+    from packaging import version as _version_module
+except ImportError:
+    try:
+        from setuptools._vendor.packaging import version as _version_module
+    except ImportError:
+        pass
+
+
+min_python_version = "3.6"
+max_python_version = "3.10"  # exclusive
+
+
+def _guard_py_ver():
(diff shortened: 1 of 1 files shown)


esc:
Fixes #698

esc:
what are the chances of this going in with 0.36.0? I think it fixes something fairly important and could be considered critical.

gmarkall:
Since it seems aimed to have an effect on installation of wheels with pip, would it be prudent to make a new RC for 0.36 with it if it's to go in for 0.36?

esc:
@gmarkall yes, a second RC for this with wheels would be required.

stuartarchibald (approved):
Thanks for the patch. On inspection this looks the same as the equivalent in Numba, and testing with Python 3.5 gives:
```
$ python setup.py install
  File "setup.py", line 55, in <module>
    _guard_py_ver()
  File "setup.py", line 52, in _guard_py_ver
    raise RuntimeError(msg.format(cur_py, min_py, max_py))
RuntimeError: Cannot install on Python version 3.5.6; only versions >=3.6,<3.10 are supported.
```
which is expected. I've not tested against a Python 3.10 alpha but the logic seems valid.

PR #702, opened 2021-02-25
Implement minimal support for call-site argument attributes

dnadlinger:
This implements support for adding argument attributes at call
sites (`call`, `invoke`), which is necessary to actually emit
functions with attributes such as `sret` (for which leaving off
the attribute at the call site, i.e. specifying it only in the
function declaration, leads to miscompilations).

I opted for a fairly minimal implementation, just optionally
taking a dictionary of `AttributeSet`s in the constructur, as
the vast majority of call sites will not need to specify extra
attributes. Not sure whether a dictionary indexed by positional
indices is the most idiomatic solution…

I also added a commit removing the `tail` parameter from the
InvokeInstr constructor, as the tail call kind attributes aren't
actually supported on `invoke` by LLVM. Previously, that
parameter was just silently ignored, which doesn't seem
desirable.

### DIFF
--- a/llvmlite/ir/instructions.py
+++ b/llvmlite/ir/instructions.py
@@ -154,10 +154,12 @@ class CallInstr(Instruction):
 
 class InvokeInstr(CallInstr):
     def __init__(self, parent, func, args, normal_to, unwind_to, name='',
-                 cconv=None):
+                 cconv=None, fastmath=(), attrs=(), arg_attrs=None):
         assert isinstance(normal_to, Block)
         assert isinstance(unwind_to, Block)
-        super(InvokeInstr, self).__init__(parent, func, args, name, cconv)
+        super(InvokeInstr, self).__init__(parent, func, args, name, cconv,
+                                          tail=False, fastmath=fastmath,
+                                          attrs=attrs, arg_attrs=arg_attrs)
         self.opname = "invoke"
         self.normal_to = normal_to
         self.unwind_to = unwind_to
--- a/llvmlite/ir/instructions.py
+++ b/llvmlite/ir/instructions.py
@@ -5,7 +5,7 @@ Implementation of LLVM IR instructions.
 from llvmlite.ir import types
 from llvmlite.ir.values import (Block, Function, Value, NamedValue, Constant,
                                 MetaDataArgument, MetaDataString, AttributeSet,
-                                Undefined)
+                                Undefined, ArgumentAttributes)
 from llvmlite.ir._utils import _HasMetadata
 
 
@@ -63,13 +63,20 @@ class FastMathFlags(AttributeSet):
 
 class CallInstr(Instruction):
     def __init__(self, parent, func, args, name='', cconv=None, tail=False,
-                 fastmath=(), attrs=()):
+                 fastmath=(), attrs=(), arg_attrs=None):
         self.cconv = (func.calling_convention
                       if cconv is None and isinstance(func, Function)
                       else cconv)
         self.tail = tail
         self.fastmath = FastMathFlags(fastmath)
         self.attributes = CallInstrAttributes(attrs)
+        self.arg_attributes = {}
+        if arg_attrs:
+            for idx, attrs in arg_attrs.items():
+                if not (0 <= idx < len(args)):
+                    raise ValueError("Invalid argument index {}"
+                                     .format(idx))
+                self.arg_attributes[idx] = ArgumentAttributes(attrs)
 
         # Fix and validate arguments
         args = list(args)
@@ -111,8 +118,13 @@ class CallInstr(Instruction):
         return self.callee
 
     def _descr(self, buf, add_metadata):
-        args = ', '.join(['{0} {1}'.format(a.type, a.get_reference())
-                          for a in self.args])
+        def descr_arg(i, a):
+            if i in self.arg_attributes:
+                attrs = ' '.join(self.arg_attributes[i]._to_list()) + ' '
+            else:
+                attrs = ''
+            return '{0} {1}{2}'.format(a.type, attrs, a.get_reference())
+        args = ', '.join([descr_arg(i, a) for i, a in enumerate(self.args)])
 
         fnty = self.callee.function_type
         # Only print function type if variable-argument
--- a/docs/source/user-guide/ir/ir-builder.rst
+++ b/docs/source/user-guide/ir/ir-builder.rst
@@ -576,13 +576,11 @@ The following methods are all :ref:`terminators <terminator>`:
 Exception handling
 ------------------
 
-* .. method:: IRBuilder.invoke(self, fn, args, normal_to, unwind_to, name='', cconv=None, tail=False)
+* .. method:: IRBuilder.invoke(self, fn, args, normal_to, unwind_to, name='', cconv=None)
 
      Call function *fn* with arguments *args*, a sequence of values.
 
-     * *cconv* is the optional calling convention.
-     * *tail*, if ``True``, is a hint for the optimizer to perform
-       tail-call optimization.
+     *cconv* is the optional calling convention.
 
      If the function *fn* returns normally, control is transferred
      to *normal_to*. Otherwise, it is transferred to *unwind_to*,
--- a/docs/source/user-guide/ir/ir-builder.rst
+++ b/docs/source/user-guide/ir/ir-builder.rst
@@ -516,7 +516,8 @@ Memory
 Function call
 ---------------
 
-.. method:: IRBuilder.call(fn, args, name='', cconv=None, tail=False, fastmath=())
+.. method:: IRBuilder.call(fn, args, name='', cconv=None, tail=False, \
+   fastmath=(), attrs=(), arg_attrs=None)
 
    Call function *fn* with arguments *args*, a sequence of values.
 
@@ -526,6 +527,18 @@ Function call
    * *fastmath* is a string or a sequence of strings of names for
      `fast-math flags
      <http://llvm.org/docs/LangRef.html#fast-math-flags>`_.
+   * *attrs* is a string or sequence of strings of function attributes to
+     attach to the call site.
+   * *arg_attrs* is a dictionary matching argument indices (as regular
+     integers, starting at zero) to strings or sequences of strings giving
+     the attributes to attach to the respective argument at this call site.
+     If an index is not present in the dictionary, or *arg_attrs* is missing
+     entirely, no attributes are emitted for the given argument.
+
+     If some attributes, such as ``sret``, are specified at the function
+     declaration, they must also be specified at each call site for
+     correctness. (As of LLVM 11, this does not seem to be explicitly
+     specified in the LLVM language reference.)
 
 
 Branches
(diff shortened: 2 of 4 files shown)


dnadlinger:
Fixed the flake failure. I'd appreciate quick feedback on the API, as I need to merge this into a downstream fork that builds against LLVM 6 too soon, and would prefer that being forward-compatible to whatever ends up upstream (i.e. this PR).

esc:
@dnadlinger thank you for submitting this. Please be advised that we are currently in a release phase and thus our resources are quite limited -- so we may not be able provide you with feedback all that quickly.

dnadlinger:
Any feedback yet now that 0.36.0 is out the door?

dnadlinger:
Thanks, that would be much appreciated in order to avoid us having to maintain an incompatible downstream fork.

gmarkall (commented):
Many thanks for proposing this addition! Call-site argument attributes are something that I've been interested in in the past, and I've had time to have a look over the PR now - I'm leaving comments to begin with as I think some additional feedback from other maintainers is appropriate here.

In general I'm positive about the PR and would like to help move it forward. I do think the proposed API needs to be discussed before deciding on whether it should be accepted as-is - to this end I propose that I'll present and discuss it a bit at the Numba public meeting tomorrow to gather comments from other developers.

In the meantime, I have a few comments and questions:

- In my local checkout, merging current master into this PR results in something that builds and passes all tests. :tada: 
- WRT removing the `tail` kwarg from `IRBuilder.invoke`: I think it's fine to remove the documentation for it. The actual removal of the kwarg could break some code though, so I'd recommend instead of removing it, ignoring it and emitting a warning about the kwarg being deprecated (and also adding a note about the deprecation to the Deprecation Notices section of the documentation). Once this is merged, then the following release may remove the kwarg.
- I'd like to see a little more variety in the testing - e.g. function calls with attributes on a subset of the arguments, use of different attributes, etc - checking a single attribute on a single argument seems a little short of assuring the continued correctness of the logic that generates argument attributes.
- Once we're aligned on the API, documentation will also need updating to explain how argument attributes are added to calls / invokes.
- What are your thoughts on the use of multiple attributes, e.g. `noalias sret`? I believe that the dict of indices approach doesn't support this - is there a way to do this?

dnadlinger:
Many thanks for the review.

I've slightly extended the test case to include multiple attributes on multiple arguments. This should answer your question regarding `noalias sret` as well. I can add more tests once the API is fixed, but this much seemed about in line with the rest of the file.

As for the `tail` argument, the reason I decided remove it straight away is that if we leave it in, it will strictly speaking be a breaking change to ever remove it later, as somebody might use positional arguments (and I want to add the new parameters for specifying the attributes now). I'll leave the judgement on this to you, but to me, it seems exceedingly unlikely that this would break the code of anybody who knows what they are doing - the C++ API doesn't have the equivalent of `tail` either, anything other than the default value `False` would have lead to an error anyway, and it's a specific named method (`IRBuilder.invoke`), not some generic API where it would have been convenient to pass `tail` for uniformity. I'm happy to just leave it in if this is at all contentious, though.

gmarkall:
Many thanks for the response and the updated test - I had forgotten that the `ArgumentAttributes` class could hold multiple attributes.

## `arg_attrs`

Feedback from the Numba meeting suggested that it is OK to accept the API design as-is - some notes on the discussions follow.

 I was concerned that this is a bit asymmetric with the API for adding attributes to arguments in declarations, which is used like:

```python
func = self.function()
func.args[0].add_attribute("zeroext")
func.args[1].attributes.dereferenceable = 5
func.args[1].attributes.dereferenceable_or_null = 10
func.args[3].attributes.align = 4
func.args[3].add_attribute("nonnull")
```

However, implementing a similar API for the call site would be a more invasive change as presently call arguments are other instructions, rather than an object representing the argument (c.f. the `Argument` class held for each argument of a declaration).

The addition of `arg_attrs` to the API is additive rather than invasive, and it appears that we could in future implement an API mirroring the one for argument attributes in declarations whilst still supporting the API proposed here.

## `tail=` kwarg removal

Given the unlikeliness of it being used in real code (and the trivial change required by anyone using it once it is removed), it was agreed that it is OK for this PR to remove it, noting that we should add an issue to the milestone for the next release reminding that the removal should be mentioned in the release notes.

## Tests

> I can add more tests once the API is fixed, but this much seemed about in line with the rest of the file.

Thanks - IMO the tests in the rest of the file are a little light, hence the request for a little more testing. I think the additional example you've added suffices for the call instruction. It would be good to add a check of `invoke` with `fastmath` and `attrs` to ensure that there are no surprises with the addition of those kwargs.

## Documentation

> I'll add a few lines to the documentation once the API is agreed on.

Many thanks - please go ahead and update the documentation for the APIs that have been added/changed here.

esc:
@dnadlinger may I enquire about the state of this PR. The next release is only about 4-6 weeks away and I would like to get this in by then, if possible.

dnadlinger:
Ack; I've put it on my list for this weekend.

dnadlinger:
On a second look, expecting instances of `ArgumentAttributes` to be passed wasn't really in line with the rest of the API, which expects all attributes as strings, or iterables thereof. I changed the API accordingly.

Also added a catch-all test case for all the optional invoke attributes. I didn't want to over-engineer this, so I just added one test to avoid all the repeated setup code.

esc:
@dnadlinger thank you for tending to this. The ball is now back in you court, @gmarkall

gmarkall:
For reference, the changes since my last review (if I've correctly managed to fetch and compare the right revisions) are presently:

```diff
diff --git a/docs/source/user-guide/ir/ir-builder.rst b/docs/source/user-guide/ir/ir-builder.rst
index ee559da..d89e9fa 100644
--- a/docs/source/user-guide/ir/ir-builder.rst
+++ b/docs/source/user-guide/ir/ir-builder.rst
@@ -516,7 +516,8 @@ Memory
 Function call
 ---------------
 
-.. method:: IRBuilder.call(fn, args, name='', cconv=None, tail=False, fastmath=())
+.. method:: IRBuilder.call(fn, args, name='', cconv=None, tail=False, \
+   fastmath=(), attrs=(), arg_attrs=None)
 
    Call function *fn* with arguments *args*, a sequence of values.
 
@@ -526,6 +527,18 @@ Function call
    * *fastmath* is a string or a sequence of strings of names for
      `fast-math flags
      <http://llvm.org/docs/LangRef.html#fast-math-flags>`_.
+   * *attrs* is a string or sequence of strings of function attributes to
+     attach to the call site.
+   * *arg_attrs* is a dictionary matching argument indices (as regular
+     integers, starting at zero) to strings or sequences of strings giving
+     the attributes to attach to the respective argument at this call site.
+     If an index is not present in the dictionary, or *arg_attrs* is missing
+     entirely, no attributes are emitted for the given argument.
+
+     If some attributes, such as ``sret``, are specified at the function
+     declaration, they must also be specified at each call site for
+     correctness. (As of LLVM 11, this does not seem to be explicitly
+     specified in the LLVM language reference.)
 
 
 Branches
@@ -576,16 +589,18 @@ The following methods are all :ref:`terminators <terminator>`:
 Exception handling
 ------------------
 
-* .. method:: IRBuilder.invoke(self, fn, args, normal_to, unwind_to, name='', cconv=None)
+* .. method:: IRBuilder.invoke(fn, args, normal_to, unwind_to, name='', \
+     cconv=None, fastmath=(), attrs=(), arg_attrs=None)
 
      Call function *fn* with arguments *args*, a sequence of values.
 
-     *cconv* is the optional calling convention.
-
      If the function *fn* returns normally, control is transferred
      to *normal_to*. Otherwise, it is transferred to *unwind_to*,
      whose first non-phi instruction must be :class:`LandingPad`.
 
+     The remaining arguments give additional attributes to specify
+     at the call site; see :meth:`call` for a description.
+
 * .. method:: IRBuilder.landingpad(typ, personality, name='', cleanup=False)
 
      Describe which exceptions this basic block can handle.
diff --git a/llvmlite/ir/instructions.py b/llvmlite/ir/instructions.py
index 4c6dea0..f337c15 100644
--- a/llvmlite/ir/instructions.py
+++ b/llvmlite/ir/instructions.py
@@ -5,7 +5,7 @@ Implementation of LLVM IR instructions.
 from llvmlite.ir import types
 from llvmlite.ir.values import (Block, Function, Value, NamedValue, Constant,
                                 MetaDataArgument, MetaDataString, AttributeSet,
-                                Undefined)
+                                Undefined, ArgumentAttributes)
 from llvmlite.ir._utils import _HasMetadata
 
 
@@ -70,7 +70,13 @@ class CallInstr(Instruction):
         self.tail = tail
         self.fastmath = FastMathFlags(fastmath)
         self.attributes = CallInstrAttributes(attrs)
-        self.arg_attributes = arg_attrs or {}
+        self.arg_attributes = {}
+        if arg_attrs:
+            for idx, attrs in arg_attrs.items():
+                if not (0 <= idx < len(args)):
+                    raise ValueError("Invalid argument index {}"
+                                     .format(idx))
+                self.arg_attributes[idx] = ArgumentAttributes(attrs)
 
         # Fix and validate arguments
         args = list(args)
diff --git a/llvmlite/tests/test_ir.py b/llvmlite/tests/test_ir.py
index 259c4f9..dfd1fa8 100644
--- a/llvmlite/tests/test_ir.py
+++ b/llvmlite/tests/test_ir.py
@@ -1194,8 +1194,8 @@ my_block:
             fun,
             (retval, ir.Constant(int32, 42), other),
             arg_attrs={
-                0: ir.ArgumentAttributes(('sret', 'noalias')),
-                2: ir.ArgumentAttributes('noalias'),
+                0: ('sret', 'noalias'),
+                2: 'noalias'
             }
         )
         self.check_block(block, """\
@@ -1220,6 +1220,39 @@ my_block:
                     to label %"normal" unwind label %"unwind"
             """)
 
+    def test_invoke_attributes(self):
+        block = self.block(name='my_block')
+        builder = ir.IRBuilder(block)
+        fun_ty = ir.FunctionType(
+            ir.VoidType(), (int32.as_pointer(), int32, int32.as_pointer()))
+        fun = ir.Function(builder.function.module, fun_ty, 'fun')
+        fun.calling_convention = "fastcc"
+        fun.args[0].add_attribute('sret')
+        retval = builder.alloca(int32, name='retval')
+        other = builder.alloca(int32, name='other')
+        bb_normal = builder.function.append_basic_block(name='normal')
+        bb_unwind = builder.function.append_basic_block(name='unwind')
+        builder.invoke(
+            fun,
+            (retval, ir.Constant(int32, 42), other),
+            bb_normal,
+            bb_unwind,
+            cconv='fastcc',
+            fastmath='fast',
+            attrs='noinline',
+            arg_attrs={
+                0: ('sret', 'noalias'),
+                2: 'noalias'
+            }
+        )
+        self.check_block(block, """\
+        my_block:
+            %"retval" = alloca i32
+            %"other" = alloca i32
+            invoke fast fastcc void @"fun"(i32* noalias sret %"retval", i32 42, i32* noalias %"other") noinline
+                to label %"normal" unwind label %"unwind"
+        """)  # noqa E501
+
     def test_landingpad(self):
         block = self.block(name='my_block')
         builder = ir.IRBuilder(block)
```

@dnadlinger For any further changes could you please append commits to the history instead of rewriting and force-pushing? The rewrite makes it very difficult to track changes throughout the review process - many thanks in advance.

llvmlite/ir/instructions.py
```diff
@@ -63,13 +63,20 @@ class FastMathFlags(AttributeSet):
...
                       if cconv is None and isinstance(func, Function)
                       else cconv)
         self.tail = tail
         self.fastmath = FastMathFlags(fastmath)
         self.attributes = CallInstrAttributes(attrs)
+        self.arg_attributes = {}
+        if arg_attrs:
+            for idx, attrs in arg_attrs.items():
+                if not (0 <= idx < len(args)):
+                    raise ValueError("Invalid argument index {}"
```
gmarkall, line 77: I don't think this is hit in the tests - could you add a test that checks the exception is raised when the index is out of bounds please?

gmarkall (changes requested):
Many thanks for the updates! This is looking good, with just one addition suggested on the diff - after that, I'd be happy to approve this PR.

> On a second look, expecting instances of `ArgumentAttributes` to be passed wasn't really in line with the rest of the API, which expects all attributes as strings, or iterables thereof. I changed the API accordingly.

Thanks for this - this makes sense and does look better.

> Also added a catch-all test case for all the optional invoke attributes. I didn't want to over-engineer this, so I just added one test to avoid all the repeated setup code.

The one test looks good.

dnadlinger:
Apologies; I had missed this, as the issue was still tagged as "waiting for reviewer".

I've added the required test now (in a separate commit 😃).

gmarkall (approved):
Many thanks for the update, and sorry about the incorrect tag!

This now looks good to me.

esc:
@dnadlinger thank you for the patch! @gmarkall thank you for the review!

PR #706, opened 2021-03-11
Update changelog for 0.36.0 final 

sklam:
- update release date

### DIFF
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1,5 +1,5 @@
 v0.36.0 (March 11, 2021)
----------------------------
+------------------------
 
 The main feature in this release is Python 3.9 support. Also, some changes to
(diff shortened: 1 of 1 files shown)


CHANGE_LOG
```diff
@@ -1,4 +1,4 @@
-v0.36.0 (February 04, 2021)
+v0.36.0 (March 11, 2021)
 ---------------------------
```
stuartarchibald, line 2: Need to alter the number of dashes?

sklam: doh

esc:
Would suggest to ignore Public CI since it is obviously glitchin'!

stuartarchibald (approved):
Thanks for the patch.

stuartarchibald (approved):
Thanks for the patch!

PR #709, opened 2021-03-16
Add flag handling to more instructions.

### LINKED ISSUE #798: `select` doesn't output its fastmath-flags
## Reporting a bug

- [X] I have tried using the latest released version of llvmlite (most recent is
 visible in the change log (https://github.com/numba/llvmlite/blob/master/CHANGE_LOG).
- [X] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

LLVM-IR `select` doesn't get any fastmath flags during IR generation, even though it is allowed. From [LangRef 11](https://releases.llvm.org/11.0.0/docs/LangRef.html#select-instruction):
```llvm
<result> = select [fast-math flags] selty <cond>, <ty> <val1>, <ty> <val2>             ; yields ty

selty is either i1 or {<N x i1>}
```

Relevant llvmlite code:
https://github.com/numba/llvmlite/blob/daf7c89764a1e565a0342ed6a90e6e1eabc58d82/llvmlite/ir/instructions.py#L309-L315
## Example
```python
from llvmlite import ir
import llvmlite.binding as llvm
from llvmlite.ir.transforms import Visitor


class FastFloatBinOpVisitor(Visitor):
    """
    A pass to add fastmath flags to float-binop instruction iff they don't have
    any flags.
    """

    float_binops = ["fadd", "fsub", "fmul", "fdiv", "frem", "fcmp", "select"]

    def __init__(self, flags):
        self.flags = flags

    def visit_Instruction(self, instr):
        if instr.opname in self.float_binops and not instr.flags:
            for flag in self.flags:
                instr.flags.append(flag)


float_t = ir.FloatType()
double_t = ir.DoubleType()
bool_t = ir.IntType(1)
double_ptr_t = ir.PointerType(double_t)
ir_module = ir.Module("mod")
func = ir.Function(ir_module, ir.FunctionType(float_t, tuple()), "switch_fn")

block = func.append_basic_block("entry")
builder = ir.IRBuilder(block)
comp = ir.Constant(bool_t, 1)
one = ir.Constant(float_t, 1)
two = ir.Constant(float_t, 2)
three = builder.fadd(one, two)
res = builder.select(comp, one, three)
builder.ret(res)

FastFloatBinOpVisitor(["fast"]).visit(ir_module)
mod = llvm.parse_assembly(str(ir_module))
print(mod)
```

Expected:
```llvm
; ModuleID = '<string>'
source_filename = "<string>"

define float @switch_fn() {
entry:
  %.2 = fadd fast float 1.000000e+00, 2.000000e+00
  %.3 = select fast i1 true, float 1.000000e+00, float %.2
  ret float %.3
}
```
Actual
```llvm
; ModuleID = '<string>'
source_filename = "<string>"

define float @switch_fn() {
entry:
  %.2 = fadd fast float 1.000000e+00, 2.000000e+00
  %.3 = select i1 true, float 1.000000e+00, float %.2
  ret float %.3
}
```

If you confirm I'll send a PR.

stuartarchibald:
As title.

Closes #798

### DIFF
--- a/llvmlite/llvmpy/core.py
+++ b/llvmlite/llvmpy/core.py
@@ -243,11 +243,13 @@ class Builder(ir.IRBuilder):
         else:
             return self.icmp_signed(_icmp_smap[pred], lhs, rhs, name=name)
 
-    def fcmp(self, pred, lhs, rhs, name=''):
+    def fcmp(self, pred, lhs, rhs, name='', flags=()):
         if pred in _fcmp_umap:
-            return self.fcmp_unordered(_fcmp_umap[pred], lhs, rhs, name=name)
+            return self.fcmp_unordered(_fcmp_umap[pred], lhs, rhs, name=name,
+                                       flags=flags)
         else:
-            return self.fcmp_ordered(_fcmp_omap[pred], lhs, rhs, name=name)
+            return self.fcmp_ordered(_fcmp_omap[pred], lhs, rhs, name=name,
+                                     flags=flags)
 
 
 class MetaDataString(ir.MetaDataString):
--- a/llvmlite/ir/builder.py
+++ b/llvmlite/ir/builder.py
@@ -626,12 +626,13 @@ class IRBuilder(object):
         self._insert(instr)
         return instr
 
-    def select(self, cond, lhs, rhs, name=''):
+    def select(self, cond, lhs, rhs, name='', flags=[]):
         """
         Ternary select operator:
             name = cond ? lhs : rhs
         """
-        instr = instructions.SelectInstr(self.block, cond, lhs, rhs, name=name)
+        instr = instructions.SelectInstr(self.block, cond, lhs, rhs, name=name,
+                                         flags=flags)
         self._insert(instr)
         return instr
 
@@ -983,8 +984,8 @@ class IRBuilder(object):
 
     # PHI APIs
 
-    def phi(self, typ, name=''):
-        inst = instructions.PhiInstr(self.block, typ, name=name)
+    def phi(self, typ, name='', flags=()):
+        inst = instructions.PhiInstr(self.block, typ, name=name, flags=flags)
         self._insert(inst)
         return inst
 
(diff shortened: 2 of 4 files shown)


stuartarchibald:
Note to self. Need to check `fneg` https://llvm.org/docs/LangRef.html#fneg-instruction, fast math handling is ok.

llvmlite/llvmpy/core.py
```diff
@@ -243,11 +243,13 @@ def icmp(self, pred, lhs, rhs, name=''):
         else:
             return self.icmp_signed(_icmp_smap[pred], lhs, rhs, name=name)
 
-    def fcmp(self, pred, lhs, rhs, name=''):
+    def fcmp(self, pred, lhs, rhs, name='', flags=()):
```
gmarkall, line 246: I wonder whether we should leave the llvmpy layer as-is for this PR rather than changing it, and instead:

- Swap all uses of `llvmpy.core.Builder` in Numba for `ir.IRBuilder` and replace the `icmp()` and `fcmp()` calls with the relevant ones (`icmp_unsigned()`, etc.)
- Remove the `Builder` class from here.

My thinking here is that llvmpy was only intended as a compatibility layer for Numba, and will eventually be removed (I did some Numba changes / removal in the past) - so we should move in that direction instead of "maintaining" it.

gmarkall (changes requested):
I think this looks good for `phi` and `select`. Related points I think we should address:

- What do you think about leaving `llvmpy.core.Builder` as-is and attempting to remove it instead? (I'm happy to put together a Numba PR and llvmlite one that removes it if you agree its a good idea).
- It looks like (from inspecting the code) that FP comparisons will output their flags, but maybe we should add a couple of flags to the tests to make sure?
- Likewise for `fneg`, it looks like it will output flags (and discussion in #798 suggests it does) but maybe we should add a test for it?

stuartarchibald:
Closing as https://github.com/numba/llvmlite/pull/825 continued this patch and is merged.

PR #710, opened 2021-03-21
Allow `gep` instructions to accept bare Python integers as indices.

trbabb:
New feature: `gep` instructions can accept Python integers as indices, instead of requiring the client code to tediously construct ir.Constants around each index.

For example:

```
int32 = ir.IntType(32)
builder.gep(my_object, [int32(5), int32(0), int32(3)])
```

becomes

```
builder.gep(my_object, [5, 0, 3])
```

The width of each converted integer will be chosen to be the smallest number of 8-bit bytes that is a power of two and can represent the value.

It is permissible to intermix Python integers with IR instructions in the index list.

### DIFF
--- a/llvmlite/ir/_utils.py
+++ b/llvmlite/ir/_utils.py
@@ -78,3 +78,17 @@ class _HasMetadata(object):
             return ', '.join(buf)
         else:
             return ''
+
+
+def bytes_to_represent(i):
+    """Take a Python integer `i` and return the power-of-two number
+    of bytes required to represent the integer."""
+    N_BITS_IN_BYTE = 8
+    def ceil_div(a, b):
+        return  (a + b - 1) // b
+    nbits  = i.bit_length()
+    nbytes = ceil_div(nbits, N_BITS_IN_BYTE)
+     # make the number of bytes a power of 2
+    return 2 ** max(nbytes - 1, 0).bit_length()
+
+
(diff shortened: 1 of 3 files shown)


esc:
@trbabb thank you for submitting this to the Numba issue tracker. I have added this to the 0.37.0 milestone. Also, it seems like there are some remaining flake8 issues to be resolved.

trbabb:
Fixed the stray whitespace, but the overlong line is inlined SSA, which that file is already full of, and which I can't really truncate. What do I need to do to prevent that from failing the test?

esc:
> Fixed the stray whitespace, but the overlong line is inlined SSA, which that file is already full of, and which I can't really truncate. What do I need to do to prevent that from failing the test?

This should help: https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors

gmarkall (changes requested):
Many thanks for the PR!

I like the idea of this change, but I think the changes to the tests need modifying. Instead of modifying an existing test case, could you add a new test case that tests accepting integers for the `gep` instruction (and a mix of integers and `ir.Constant` values) please?

gmarkall:
@trbabb Many thanks for the PR and your efforts so far!

I'm spending some time going through the llvmlite PR backlog, and I have a couple of questions on this PR:

- Do you plan to continue with this PR? If it's difficult for you to work on this PR for any reason, please let me know and we can look into whether / how to proceed further.
- If you're keen to continue - do you need any further advice on how to proceed?

PR #711, opened 2021-03-23
Use conda for readthedocs with Python 3.9

esc:
Followup from #114


sklam (approved):
Thanks for the patch!

I've have manually verified the doc build from RTD. It's working fine.

esc:
> Thanks for the patch!
> 
> I've have manually verified the doc build from RTD. It's working fine.

Splendid!

PR #716, opened 2021-04-18
Align linker flags between CMake and non-CMake builds

chrisburr:
While debugging https://github.com/conda-forge/llvmlite-feedstock/issues/54 I discovered `-flto -Wl,--exclude-libs,ALL` isn't used when `$LLVMLITE_USE_CMAKE` is set.


esc:
@chrisburr thank you for submitting this. I have added it to the queue for review.

esc:
Thank you for this, I have cherry-picked this patch into #730 and it will find it's way into master through that. Thanks for the patch!

PR #723, opened 2021-05-22
[PATCH] make LLVM 9.0.1 accept arm64-apple-darwin

hide-kawabata:
The original LLVM 9.0.1 (https://github.com/llvm/llvm-project/releases/tag/llvmorg-9.0.1) does not allow target triple of arm64-apple-darwin, which prevents llvmlite-0.36.0 working correctly on Apple Silicon M1 macs. This patch seems to be useful as a workaround for that. Applying flags of -DLLVM_HOST_TRIPLE=arm64-apple-darwin20.3.0 and -DLLVM_TARGETS_TO_BUILD=AArm64 might be required while building LLVM.

### DIFF
--- a/conda-recipes/llvm-9.0.1-arm64.patch
+++ b/conda-recipes/llvm-9.0.1-arm64.patch
@@ -1,0 +1,24 @@
+diff --git a/llvm/include/llvm/ADT/Triple.h b/llvm/include/llvm/ADT/Triple.h
+index edeb31e..faae6f9 100644
+--- a/llvm/include/llvm/ADT/Triple.h
++++ b/llvm/include/llvm/ADT/Triple.h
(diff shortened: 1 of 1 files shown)


gmarkall:
@hide-kawabata Many thanks for this PR and your efforts so far, and I'm sorry that we have not already made further progress with it. I'm spending some time going through the llvmlite PR backlog - this PR seems to be one that's been waiting for a while.

I think that this is no longer needed now that we're on LLVM 11 and it is reported that llvmlite can work on M1 (even if not fully supported). Is this the case? If so, can we close this PR?

hide-kawabata:
OK to close the PR although I haven't checked LLVM 11 version yet.
Sorry for late response.

PR #726, opened 2021-06-09
Disallow alwaysinline and noinline on functions

gmarkall:
Adding "alwaysinline" and "noinline" attributes to a function results in an error similar to the following when compiling with LLVM:

```
Attributes 'noinline and alwaysinline' are incompatible!
LLVM ERROR: Broken module found, compilation aborted!
```

and then the interpreter is killed. This PR prevents both attributes being added at the Python level, to prevent the above error message and termination.

### DIFF
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -44,8 +44,9 @@ getelementptr
 An LLVM :ref:`instruction` that lets you get the address of a
 subelement of an aggregate data structure.
 
-See :llvm:ref:`i_getelementptr` in the official LLVM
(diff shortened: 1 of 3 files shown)


esc:
I looked at the build failure, this has nothing to do with the PR but is caused by an upgrade of `sphinx` from v3.2.1 to v4.0.1 and can be triggered by including `-Wn` which treats this new warning as an error and breaks the build. Confirmed locally, I can reproduce this build failure.

esc:
Thanks for the patch!

PR #730, opened 2021-06-14
Don't export LLVM symbols when linking statically

esc:
As title.


stuartarchibald (approved):
Thanks for the patch. This has been checked on OSX via an OOB conversation, the essence of which was running `nm` on the DSO and looking for symbols with type `T`. On linux, the changes now match the flags set in `ffi/Makefile.linux`.

esc:
@stuartarchibald thanks for the approval!

PR #734, opened 2021-06-24
bump llvmdev to 11

esc:
As it is now passing everywhere else, we upgrade public CI now too.


esc:
I have moved the llvmdev 11 build 2 packages to `numba/label/main` and now everything is well.

esc:
the RTD build apparently completed:

https://readthedocs.org/projects/llvmlite/builds/14087364/

But the status was not communicated back to github, so this PR, looks like it is hanging.

stuartarchibald (approved):
Thanks for the patch.

PR #739, opened 2021-06-30
Bump to 0.38.0 development series

sklam:
as titled


esc:
@sklam thanks for the patch!

PR #747, opened 2021-07-16
Remove 3.6 support

esc:
Remove 3.6 support and include clamping the maximum Python version in the conda recipe as this had been omitted.


stuartarchibald (approved):
Thanks for the patch.

PR #750, opened 2021-07-16
More cherrypicked patches for release 0.37.0rc2

sklam:
as titled


stuartarchibald (approved):
Thanks for the patch. Cherry-pick looks good, checked by locally making the same and diffing.

PR #751, opened 2021-07-20
Fix encoding of unicode string in metadata.

sklam:
Supersede to https://github.com/numba/numba/pull/7208

Fixes unicode encoding for LLVM debug metadata string.


stuartarchibald (approved):
Thanks for the patch!

esc:
@sklam thank you for the patch! @stuartarchibald thank you for the review!

PR #752, opened 2021-07-20
Skip test if libm is not found.

sklam:
No guarantee that libm is search-able in path.

The test is failing in my OSX Big Sur.


esc:
Anyone with a BigSure system to confirm this fix?

guilhermeleobas:
I'm on BigSur and all tests pass without this patch:
https://gist.github.com/guilhermeleobas/29863546c0e766bb438a7d4aa23c8f04

esc:
> I'm on BigSur and all tests pass without this patch: https://gist.github.com/guilhermeleobas/29863546c0e766bb438a7d4aa23c8f04

OK. Thank you! Just to confirm: this patch is NOT needed for you on OSX Big Sur and all tests pass fine?

guilhermeleobas:
> 
> 
> > I'm on BigSur and all tests pass without this patch: https://gist.github.com/guilhermeleobas/29863546c0e766bb438a7d4aa23c8f04
> 
> OK. Thank you! Just to confirm: this patch is NOT needed for you on OSX Big Sur and all tests pass fine?

Yes, that's correct!

gmarkall:
@sklam is this needed? Or did it turn out to be some anomaly on your system?

jakirkham:
Think this is a real issue.

<hr>

TBH don't recall macOS shipping a `libm` per se in the past. Instead macOS provides this functionality though system libraries (like `libSystem.B.dylib`). It use to be that one wouldn't link to `libm` on macOS as it didn't exist (so one had to handle this specially in build scripts). However macOS now ships a `libm.tbd` (perhaps for ease of use?), but this is just a shim to system libraries.

To demonstrate this we can take a C file like so

```C
#include <math.h>
#include <stdio.h>

int main() {
	printf("The value of e is: %f\n", exp(1));
	return 0;
}
```

Build it (works fine whether or not `-lm` is there)...

```bash
$ clang -lm main.c
```

...and inspect the linkages

```bash
$ otool -L ./a.out 
./a.out:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
```

Note that no `libm.dylib` shows up.

<hr>

Even though Python can return a result for this, it appears to be incorrect as well.

```python
Python 3.10.5 | packaged by conda-forge | (main, Jun 14 2022, 07:09:13) [Clang 13.0.1 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from ctypes.util import find_library
>>> from os.path import exists
>>> find_library("m")
'/usr/lib/libm.dylib'
>>> exists(find_library("m"))
False
```

<hr>

Also would note this test caused us issues in conda-forge recently ( https://github.com/conda-forge/llvmlite-feedstock/pull/62#issuecomment-1209762518 ).

Given all of this would suggest dropping this test on macOS.

stuartarchibald:
xref: https://github.com/conda-forge/llvmlite-feedstock/pull/62#issuecomment-1210408912 suggest discussing a point release to coincide with Numba 0.56.1 to patch this test.

sklam:
I've cherrypicked the prefered changes from https://github.com/numba/llvmlite/pull/870 into this PR

jakirkham:
Thanks Siu! 🙏 Should add improvements on that are welcome (if others have any) 🙂

gmarkall (approved):
This looks good to me (I don't have a mac to test locally, but given that the test no longer runs on it as the only change here, I'm comfortable with it).

esc:
> This is pending `llvmlite_155`.

This passed!

jakirkham:
Thanks all! 🙏

PR #757, opened 2021-07-28
Add issue templates based on those in Numba

gmarkall:
These templates are similar to Numba's, but customised for llvmlite. The main aim in adding them is to direct questions and involved feature discussions toward the llvmlite section of the Numba discourse, rather than the creation of new issues.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,34 @@
+---
+name: Bug Report
+about: Report a bug. Not for asking general questions - see below.
+
+---
+
+<!--
+
+Thanks for opening an issue! To help the Numba team handle your information
+efficiently, please first ensure that there is no other issue present that
+already describes the issue you have
+(search at https://github.com/numba/llvmlite/issues?&q=is%3Aissue).
+
+-->
+
+## Reporting a bug
+
+<!--
+
+Before submitting a bug report please ensure that you can check off these boxes:
+
+-->
+
+- [ ] I have tried using the latest released version of llvmlite (most recent is
+ visible in the change log (https://github.com/numba/llvmlite/blob/master/CHANGE_LOG).
+- [ ] I have included a self contained code sample to reproduce the problem.
+  i.e. it's possible to run as 'python bug.py'.
+
+<!--
+
+Please include details of the bug here, including, if applicable, what you
+expected to happen!
+
+-->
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
+---
+name: Bug Report
+about: Report a bug. Not for asking general questions - see below.
+
+---
+
+<!--
+
+Thanks for opening an issue! To help the Numba team handle your information
```
stuartarchibald, line 9:
```suggestion
Thanks for opening an issue! To help the llvmlite team handle your information
```
Should this be Numba or llvmlite? In practice, at present, the Numba engineers are a superset of llvmlite engineers.

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,25 @@
...
+name: Feature Request
+about: Tell us about something you'd like llvmlite to support. Not for asking general questions - see below.
+
+---
+
+---
+
+<!--
+
+Thanks for opening an issue! To help the Numba team handle your information
```
stuartarchibald, line 11: Same Q RE Numba vs llvmlite as above

stuartarchibald: Also, this text probably needs an update to ask users to explicitly check for similar feature requests (seems like it's a copy paste error on Numba's original template?!).

stuartarchibald (commented):
Thanks for the patch, couple of minor things to resolve else looks good. I also wonder if there's URLs in the docs/README etc which would be made more specific as a result of this change?

.github/ISSUE_TEMPLATE/bug_report.md
gmarkall: I left it as Numba with my reasoning being that it's the Numba engineers who look after llvmlite - it doesn't feel to me that there's a separate team at all.

gmarkall:
@stuartarchibald Many thanks for the review - I've updated the feature request template accordingly. I left it at "Numba team" since it's same team involved in maintaining llvmlite.

.github/ISSUE_TEMPLATE/bug_report.md
stuartarchibald: Indeed, the reality is that is it the same engineers, agree with just leaving it! Thanks!

stuartarchibald (approved):

PR #760, opened 2021-08-05
add support for attaching metadata to global variables

john-tornblom:


gmarkall:
Thanks very much for the PR, I've queued it for review.

gmarkall:
Hi again @john-tornblom - many thanks for this PR! I'm spending some time going through the llvmlite PR backlog - this PR has been waiting a while so I'm just looking now.

I had to merge master and resolve conflicts locally to try it out. Since the resolution was simple, I've pushed the merge commit to your branch - I hope you don't mind. I'll have a look at the code and follow up with a review shortly.

llvmlite/tests/test_ir.py
```diff
@@ -387,6 +387,67 @@ def test_debug_info_2(self):
...
+                            'true, name: "gvar", type: !0)', strmod)
+        self.assert_ir_line('!3 = !DIGlobalVariableExpression(expr: !1, '
+                            'var: !2)', strmod)
+        self.assert_ir_line('@"gvar" = global float 0x4045000000000000, '
+                            '!dbg !3', strmod)
+
+        # The remaining debug info is not part of the automated test, but
+        # can be used to produce an object file that can be loaded into a
+        # debugger to print the value og gvar, e.g.,
+        #    printf "file test_debug_info_gvar.o \n p gvar" | gdb
```
gmarkall, line 429:
```suggestion
        # debugger to print the value of gvar. This can be done by printing the
        # module then compiling it with clang and inspecting with gdb:
        #
        #     clang test_debug_info_gvar.ll -c
        #     printf "file test_debug_info_gvar.o \n p gvar" | gdb
        #
        # Which should result in the output:
        #
        #     (gdb) $1 = 42
```

gmarkall (changes requested):
Many thanks for the PR - this looks generally good! I have a couple of suggestions:

- I've suggested some additions to the comment in the test to make it a little clearer how to explore the global variable with the debugger - I was a bit confused for a few minutes about the simplest way to go from having the IR module to an object file, so I hope these directions will be useful for others.
- Could you amend the documentation please? Presently `Function` is documented as having a `set_metadata()` function, but I think this needs hoisting up to the documentation for `GlobalValue` and the wording amending (see https://llvmlite.readthedocs.io/en/latest/user-guide/ir/values.html#llvmlite.ir.Function.set_metadata).
- I think `Function` no longer needs the `_HasMetadata` mixin because it gets it from `GlobalValue` - can you remove it from the `Function` class declaration please? (It's not on the diff but appears to be on line 920 / 924) of `values.py`.

john-tornblom:
Thanks for the feedback, I will address the comments during the holidays.

gmarkall:
Many thanks for addressing the feedback - I'll check this next Wednesday when I'm back at work.

gmarkall (approved):
Many thanks for addressing the feedback and your continuing contributions! This looks good!

esc:
@john-tornblom thank you for the review! @gmarkall thank you for the patch!

PR #758, opened 2021-07-30
downgrade conda compilers for llvmlite and llvmdev builds

esc:
As title

PR #765, opened 2021-08-17
Cherry-picks for 0.37.0

stuartarchibald:
* PR #758 continued. Ensure GCC7 is used on non-ARM platforms.


sklam (approved):
Thanks for the patch!

PR #564, opened 2020-03-28
Merge atomic into regular memory access & introduce volatile memory access

ARF1:
Closes #450.
Incorporates and closes #559.

This builds on @ssikdar1's works in #559 where he merged the public api of atomic load/store into regular load/store.

Further commits make following changes to @ssikdar1's work:

- restores the deleted public api for atomic memory access and adds deprecation warnings to `builder.load_atomic()` and `builder.store_atomic()`
- merges `LoadAtomicInstr` with `LoadInstr` and `StoreAtomicInstr` with `StoreAtomicInstr`
- enables the `volatile` flag for load/store
- documents api changes

PR #767, opened 2021-08-23
add the option to mark load and store instructions as volatile

john-tornblom:


john-tornblom:
Ahh, I just noticed this is already addressed by https://github.com/numba/llvmlite/pull/564, closing...

PR #775, opened 2021-10-12
targets.py: Add ABIName parameter for RISC-V hard float targets

occheung:
LLVM requires RISC-V targets to specify ABI if hard float is supported. The current implementation does not allow `TargetOptions.McOptions.ABIName` to be specified.

This patch exposes the ABIName to the `create_target_machine` function. So a RISC-V target that supports hard-float can be created like such:
```
        lltarget = llvm.Target.from_triple("riscv32-unknown-linux")
        llmachine = lltarget.create_target_machine(
                        features="+m,+a,+f,+d",
                        reloc="pic", codemodel="default", abiname="ilp32d")
```
Similar feature was also added to Rust's LLVM interface: [See discussion](https://github.com/rust-lang/rust/issues/65024) | [See commit](https://github.com/rust-lang/rust/commit/539de439ad6f32e9a9a8a593299072a106786890)

Alternatively, we can also process the target triple and features to deduce the appropriate ABI.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,6 +14,9 @@ jobs:
       py39:
         PYTHON: '3.9'
         CONDA_ENV: cienv
+      py310:
+        PYTHON: '3.10'
+        CONDA_ENV: cienv
 
 - template: buildscripts/azure/azure-linux-macos.yml
   parameters:
@@ -29,6 +32,10 @@ jobs:
         RUN_FLAKE8: yes
       py39:
         PYTHON: '3.9'
+        CONDA_ENV: cienv
+        RUN_FLAKE8: yes
+      py310:
+        PYTHON: '3.10'
         CONDA_ENV: cienv
         RUN_FLAKE8: yes
 # temporarily disabled
@@ -47,6 +54,15 @@ jobs:
         PYTHON: '3.9'
         CONDA_ENV: cienv
         WHEEL: 'yes'
+      py310_wheel:
+        PYTHON: '3.10'
+        CONDA_ENV: cienv
+        WHEEL: 'yes'
+
+      py39_docs:
+        PYTHON: '3.9'
+        CONDA_ENV: cienv
+        BUILD_DOCS: 'yes'
 
 - template: buildscripts/azure/azure-windows.yml
   parameters:
--- a/buildscripts/azure/azure-linux-macos.yml
+++ b/buildscripts/azure/azure-linux-macos.yml
@@ -15,6 +15,7 @@ jobs:
 
   steps:
     - script: |
+        set -e
         if [ "$(uname)" == "Linux" ] && [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" ]]; then sudo apt-get install -y libc6-dev-i386; fi
         echo "Installing Miniconda"
         buildscripts/incremental/install_miniconda.sh
@@ -24,11 +25,13 @@ jobs:
       displayName: 'Before Install'
 
     - script: |
+        set -e
         export PATH=$HOME/miniconda3/bin:$PATH
         buildscripts/incremental/build.sh
       displayName: 'Build'
 
     - script: |
+        set -e
         export PATH=$HOME/miniconda3/bin:$PATH
         conda install -y flake8
         echo "Running flake8 check"
@@ -37,6 +40,18 @@ jobs:
       condition: eq(variables['RUN_FLAKE8'], 'yes')
 
     - script: |
+        set -e
+        export PATH=$HOME/miniconda3/bin:$PATH
+        conda install -y sphinx sphinx_rtd_theme pygments
+        echo "Building docs"
+        cd docs
+        make SPHINXOPTS=-Wn clean html
+        cd ..
+      displayName: 'Build Docs'
+      condition: eq(variables['BUILD_DOCS'], 'yes')
+
+    - script: |
+        set -e
         export PATH=$HOME/miniconda3/bin:$PATH
         buildscripts/incremental/test.sh
       displayName: 'Test'
--- a/buildscripts/azure/azure-windows.yml
+++ b/buildscripts/azure/azure-windows.yml
@@ -12,6 +12,9 @@ jobs:
       py39:
         PYTHON: '3.9'
         CONDA_ENV: cienv
+      py310:
+        PYTHON: '3.10'
+        CONDA_ENV: cienv
 
 
   steps:
--- a/buildscripts/incremental/setup_conda_environment.sh
+++ b/buildscripts/incremental/setup_conda_environment.sh
@@ -35,17 +35,5 @@ if [[ $(uname) == Linux ]]; then
 $CONDA_INSTALL gcc_linux-64 gxx_linux-64
 fi
 
-# Install enum34 for Python < 3.4 and PyPy, and install dependencies for
-# building the docs.
-if [ "$PYTHON" == "pypy" ]; then
-  $CONDA_INSTALL zlib # pypy has no conda-level zlib linkage
-  python -m ensurepip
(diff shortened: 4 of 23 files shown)


sbourdeauducq:
Can we get this merged? Sounds like a simple change.

esc:
> Can we get this merged? Sounds like a simple and valid change.

@sbourdeauducq I think the problem is that access to RISC-V is limited. Have you been able to test this on some real hardware?

gmarkall:
I can test this - will post an update once I've done so.

sbourdeauducq:
> Have you been able to test this on some real hardware?

Definitely - our compiler uses llvmlite with the patch and we test it on our FPGA board, we even have hardware-in-the-loop CI.

gmarkall:
@sbourdeauducq How are you building and testing this? Are you building an llvmlite with the RISCV target enabled?

sbourdeauducq:
If you meant LLVM with the RISCV target enabled - yes.

gmarkall:
Thanks... I was originally going about testing this the wrong way - I was building on a RISC-V host and trying to find a way to get the LLVM JIT to work so the whole test suite could run - I've not completed testing this PR yet but haven't forgotten it.

sbourdeauducq:
If you don't have any hardware to test, QEMU can emulate RISC-V. Or just looking at the generated assembly should be enough anyway.

gmarkall:
In that case, is it possible to add a test that can run on the host with a RISC-V target that can check for the expected output in the generated assembly? That would make it a lot easier to check that I'm checking this PR correctly (and also lay a good foundation for future RISC-V-specific code generation functionality and tests).

seibert:
#785 is related, as we will want to add RISCV as a target in our LLVM builds in order to run any RISC-V tests in our CI systems.

occheung:
> In that case, is it possible to add a test that can run on the host with a RISC-V target that can check for the expected output in the generated assembly?

Tests regarding the impact of ABI are performed in LLVM already, such as [this](https://github.com/llvm/llvm-project/blob/main/llvm/test/CodeGen/RISCV/calling-conv-vector-float.ll) (albeit in later LLVM version).

AFAIK the ABIs should have no influence on how the LLVM IR needs be written?

sbourdeauducq:
> AFAIK the ABIs should have no influence on how the LLVM IR needs be written?

Yes.
The only thing that can be tested here is that when the ABI option is set from the high-level Python code, the assembly (machine, not IR) code generated at the end by LLVM is affected appropriately.

gmarkall:
Many thanks for adding the tests! I think that you have found the correct location to modify, but I'm not sure if there's additional places it would need modifying too - I will check.

gmarkall:
There's also https://github.com/numba/llvmlite/blob/master/conda-recipes/llvmdev/bld.bat#L10 and https://github.com/numba/llvmlite/blob/master/conda-recipes/llvmdev/build.sh#L8 to modify as well.

gmarkall:
Many thanks for adding the RISCV target - note that this will need the llvmdev / llvmlite packages to be rebuilt on our channel before the test starts to pass in CI.

esc:
@gmarkall thank you for tending to this PR. We will need to decide what llvmlite version this should be a part of.

sbourdeauducq:
From our perspective we would really appreciate a Conda Win64 package with RISC-V, this, and https://github.com/numba/llvmlite/pull/702.
It's easy to have patches on our Hydra CI server and NixOS machines, but for Windows it's a very different story.

gmarkall (changes requested):
This generally looks good to me. Some notes:

* WIth #788, the tests passed. :tada: 
* I was a little concerned about fragility of the test because it checks the assembly strings line-for-line, but on reflection I don't think they are likely to change unless we upgrade LLVM, so they will likely be stable enough in their current form.
* The documentation in `docs/source/user-guide/binding/target-information.rst` will also need updating, as it isn't generated from the docstring.

gmarkall:
@occheung Many thanks for updating the docs!

A side note: could I please ask that for future changes you merge master rather than rebasing please? The reason for this is that Github doesn't make it easy to track what happened on the reviewing side in newly-added commits. I *think* the only substantive change in your last push is the addition of the docs, but it's quite hard for me to tell for certain. Many thanks in advance!

gmarkall (approved):
This looks good to me, assuming that CI passes after new packages have been built following the merge of #788.

esc:
@gmarkall it looks like this is still bust

gmarkall:
@esc Let's discuss this a little in the dev meeting if there's time.

stuartarchibald:
Is the issue that the build scripts are looking in channel `numba` and picking up build 3:
https://github.com/numba/llvmlite/blob/master/buildscripts/incremental/setup_conda_environment.sh#L30
Log extract:
```
2021-11-23T12:23:01.4828781Z   added / updated specs:
2021-11-23T12:23:01.4830437Z     - llvmdev=11
2021-11-23T12:23:01.4830913Z 
2021-11-23T12:23:01.4831264Z 
2021-11-23T12:23:01.4831844Z The following packages will be downloaded:
2021-11-23T12:23:01.4832287Z 
2021-11-23T12:23:01.4832844Z     package                    |            build
2021-11-23T12:23:01.4833699Z     ---------------------------|-----------------
2021-11-23T12:23:01.4834640Z     llvmdev-11.1.0             |       he1b5a44_3       374.3 MB  numba
2021-11-23T12:23:01.4835592Z     ------------------------------------------------------------
2021-11-23T12:23:01.4836332Z                                            Total:       374.3 MB
2021-11-23T12:23:01.4836829Z 
2021-11-23T12:23:01.4837394Z The following NEW packages will be INSTALLED:
2021-11-23T12:23:01.4837907Z 
2021-11-23T12:23:01.4838694Z   llvmdev            numba/linux-64::llvmdev-11.1.0-he1b5a44_3
```
whereas it's build 4 that contains the additional cross targets and they are all in `numba/label/ci` ?

esc:
> Is the issue that the build scripts are looking in channel `numba` and picking up build 3: https://github.com/numba/llvmlite/blob/master/buildscripts/incremental/setup_conda_environment.sh#L30 Log extract:
> 
> ```
> 2021-11-23T12:23:01.4828781Z   added / updated specs:
> 2021-11-23T12:23:01.4830437Z     - llvmdev=11
> 2021-11-23T12:23:01.4830913Z 
> 2021-11-23T12:23:01.4831264Z 
> 2021-11-23T12:23:01.4831844Z The following packages will be downloaded:
> 2021-11-23T12:23:01.4832287Z 
> 2021-11-23T12:23:01.4832844Z     package                    |            build
> 2021-11-23T12:23:01.4833699Z     ---------------------------|-----------------
> 2021-11-23T12:23:01.4834640Z     llvmdev-11.1.0             |       he1b5a44_3       374.3 MB  numba
> 2021-11-23T12:23:01.4835592Z     ------------------------------------------------------------
> 2021-11-23T12:23:01.4836332Z                                            Total:       374.3 MB
> 2021-11-23T12:23:01.4836829Z 
> 2021-11-23T12:23:01.4837394Z The following NEW packages will be INSTALLED:
> 2021-11-23T12:23:01.4837907Z 
> 2021-11-23T12:23:01.4838694Z   llvmdev            numba/linux-64::llvmdev-11.1.0-he1b5a44_3
> ```
> 
> whereas it's build 4 that contains the additional cross targets and they are all in `numba/label/ci` ?

good catch, that may very well be the case!

gmarkall:
All tests passed now that the llvmdev packages are available on the main channel - many thanks @esc!

esc:
> All tests passed now that the llvmdev packages are available on the main channel - many thanks @esc!

👌 Excellent!

esc:
@occheung @sbourdeauducq @gmarkall thank you all!

PR #776, opened 2021-10-12
added ability to iterate over basic block's predecessors

carpall:
Like in the C api, basic blocks have a list of predecessors.
Predecessors are all the basic blocks which refer to a specific basic block.
For example:
```llvm
define dso_local i32 @main() #0 !dbg !9 {
  %1 = alloca i32, align 4
  %2 = alloca i32, align 4
  store i32 0, i32* %1, align 4
  call void @llvm.dbg.declare(metadata i32* %2, metadata !14, metadata !DIExpression()), !dbg !15
  store i32 1, i32* %2, align 4, !dbg !15
  %3 = load i32, i32* %2, align 4, !dbg !16
  %4 = icmp ne i32 %3, 0, !dbg !16
  br i1 %4, label %5, label %6, !dbg !18

5:                                                ; predecessors = %0
  store i32 0, i32* %2, align 4, !dbg !19
  br label %7, !dbg !21

6:                                                ; predecessors = %0
  store i32 1, i32* %2, align 4, !dbg !22
  br label %7

7:                                                ; predecessors = %6, %5
  %8 = load i32, i32* %1, align 4, !dbg !24
  ret i32 %8, !dbg !24
}
```

### DIFF
--- a/llvmlite/ir/builder.py
+++ b/llvmlite/ir/builder.py
@@ -823,7 +823,7 @@ class IRBuilder(object):
         return swt
 
     def add_pred(self, block):
-        if not self.basic_block in block.predecessors:
+        if self.basic_block not in block.predecessors:
             block.predecessors.append(self.basic_block)
 
     def branch(self, target):


llvmlite/ir/builder.py
```diff
@@ -837,6 +842,7 @@ def cbranch(self, cond, truebr, falsebr):
         br = instructions.ConditionalBranch(self.block, "br",
                                             [cond, truebr, falsebr])
         self._set_terminator(br)
+        self.set_pred(truebr); self.set_pred(falsebr)
```
gmarkall, line 845: Would look with the API change mentioned in the review as:

```suggestion
        truebr.add_predecessor(self)
        falsebr.add_predecessor(self)
```

gmarkall (changes requested):
Many thanks for the PR!

I'm finding the API confusing - wouldn't it be more straightforward if blocks had a method to add predecssors, rather than the `IRBuilder` having a `set_pred` method that sets the predecessor of the passed block as the current block? (Does the C API do things this way?)

Could you also include tests and documentation for the additional feature please?

llvmlite/ir/builder.py
Carpall: thank you for the note, the meaning of `set_pred` was to set self.basic_block as predecessor of the parameter, but it effectly adds to predecessors, then the problem was just the name of the function

Carpall (commented):
fixing problems for `flake8`

Carpall:
> > > > Could you also include tests and documentation for the additional feature please?
> > > 
> > > 
> > > I second that, thank you @gmarkall -- we will need some tests. I do like the feature though.
> > 
> > 
> > how can I add documentation for this little feature? I didn't find something about contributions on the readme
> 
> https://llvmlite.readthedocs.io/en/latest/contributing.html <-- maybe a bit out of date, unfortunately.

Now I have no time to edit the documentation, about the test I think it was so small as change to be tested
documentation:
.has_predecessors: returns true whether the block has predecessors
.predecessors: returns a list of predecessors of self block
precedessors are blocks which contain a branch to a specified block
ex:
```llvm
entry:
  br  label %bb0
bb0:
  br i1 %0, label %bb1, label %bb2 ; bb1.predecessors: [bb0], bb2.predecessors: [bb0], bb0.predecessors: [entry]
```

PR #779, opened 2021-10-14
pin compiler toolchain to version 10 on osx

esc:
As title

PR #769, opened 2021-09-02
Python 3.10

esc:
As title

PR #782, opened 2021-10-20
Test python3.10 and compiler pin

esc:
Testing compatabuility of #769 and #779


esc:
The Build Farm passed on this one!

PR #735, opened 2021-06-24
add support for more constant expressions

john-tornblom:
This is an initial PR to clarify preferred design decisions before I commit more time and effort, expect more commits based on feedback before a potential merge.

1. Is the use of a decorator called _binop appropriate? It is somewhat copy&paste from builder.py
2. Is the naming scheme "lhs" and "rhs" in docstrings OK, even though they do not match argument list?
3. is the testing approach appropriate, i.e., string compare of emitted llvm-ir?

PR #784, opened 2021-11-03
Adds _ConstOpMixin.ptrtoint to complement existing bitcast and inttoptr methods.

bslatkin:
See details and background in LLVM IR language reference:

https://llvm.org/docs/LangRef.html#ptrtoint-to-instruction

### DIFF
--- a/llvmlite/ir/values.py
+++ b/llvmlite/ir/values.py
@@ -52,6 +52,24 @@ class _ConstOpMixin(object):
                                                typ)
         return FormattedConstant(typ, op)
 
+    def ptrtoint(self, typ):
+        """
+        Cast this pointer constant to the given integer type.
+        """
+        if not isinstance(typ, types.IntType):
+            raise TypeError(
+                "can only ptrtoint() to integer type, not '%s'"
+                % (typ,))
+        if not isinstance(self.type, types.PointerType):
+            raise TypeError(
+                "can only call inttoptr() on pointer type, not '%s'"
+                % (self.type))
+
+        op = "ptrtoint ({0} {1} to {2})".format(self.type,
+                                                self.get_reference(),
+                                                typ)
+        return FormattedConstant(typ, op)
+
     def inttoptr(self, typ):
         """
         Cast this integer constant to the given pointer type.
--- a/llvmlite/tests/test_ir.py
+++ b/llvmlite/tests/test_ir.py
@@ -2233,14 +2233,14 @@ class TestConstant(TestBase):
 
         self.assertRaisesRegex(
             TypeError,
-            "can only ptrtoint\\(\\) to integer type, not 'i64\\*'",
+            r"can only ptrtoint\(\) to integer type, not 'i64\*'",
             gv.ptrtoint,
             int64.as_pointer())
 
         c2 = ir.Constant(int32, 0)
         self.assertRaisesRegex(
             TypeError,
-            "can only call inttoptr\\(\\) on pointer type, not 'i32'",
+            r"can only call inttoptr\(\) on pointer type, not 'i32'",
             c2.ptrtoint,
             int64)
 
--- a/llvmlite/tests/test_ir.py
+++ b/llvmlite/tests/test_ir.py
@@ -2225,6 +2225,25 @@ class TestConstant(TestBase):
         c = gv.bitcast(int64.as_pointer())
         self.assertEqual(str(c), 'bitcast (i32* @"myconstant" to i64*)')
 
+    def test_ptrtoint(self):
+        m = self.module()
+        gv = ir.GlobalVariable(m, int32, "myconstant")
+        c = gv.ptrtoint(int64)
+        self.assertEqual(str(c), 'ptrtoint (i32* @"myconstant" to i64)')
+
+        self.assertRaisesRegex(
+            TypeError,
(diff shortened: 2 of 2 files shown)


llvmlite/ir/values.py
```diff
@@ -52,6 +52,24 @@ def bitcast(self, typ):
...
+        Cast this pointer constant to the given integer type.
+        """
+        if not isinstance(typ, types.IntType):
+            raise TypeError(
+                "can only ptrtoint() to integer type, not '%s'"
+                % (typ,))
+        if not isinstance(self.type, types.PointerType):
+            raise TypeError(
+                "can only call inttoptr() on pointer type, not '%s'"
+                % (typ,))
```
stuartarchibald, line 66:
```suggestion
                % (self.type,))
```

llvmlite/tests/test_ir.py
```diff
@@ -2225,6 +2225,23 @@ def test_bitcast(self):
         c = gv.bitcast(int64.as_pointer())
         self.assertEqual(str(c), 'bitcast (i32* @"myconstant" to i64*)')
 
+    def test_ptrtoint(self):
+        m = self.module()
+        gv = ir.GlobalVariable(m, int32, "myconstant")
+        c = gv.ptrtoint(int64)
+        self.assertEqual(str(c), 'ptrtoint (i32* @"myconstant" to i64)')
+
+        self.assertRaises(
```
stuartarchibald, line 2234: Suggest using `assertRaisesRegex` to check the exception string is that intended.

llvmlite/tests/test_ir.py
```diff
@@ -2225,6 +2225,23 @@ def test_bitcast(self):
...
+            TypeError,
+            gv.ptrtoint,
+            int64.as_pointer())
+
+        c2 = ir.Constant(int32, 0)
+        self.assertRaises(
```
stuartarchibald, line 2240: Suggest using `assertRaisesRegex` to check the exception string is that intended.

stuartarchibald (commented):
Thanks for the patch. Couple of minor things to look at but else looks good. Thanks again!

llvmlite/ir/values.py
bslatkin: Thanks for catching this. Fixed in updated patch.

llvmlite/tests/test_ir.py
bslatkin: Good call this would have caught the mistake as well. Added in updated patch.

llvmlite/tests/test_ir.py
bslatkin: Done

bslatkin:
@stuartarchibald Okay it should be all set now -- please take another look. Thanks

stuartarchibald:
@bslatkin many thanks! As a small request, in future, please could you avoid doing rebases or force-pushes on reviewed branches because GitHub cannot track these sorts of updates very well which means review comments don't get closed automatically and the reviewer ends up having to re-review the whole diff again to check fixes (thankfully this one is quite short!). Thanks for your help with this!

llvmlite/ir/values.py
stuartarchibald: No problem, thanks for the fix.

llvmlite/tests/test_ir.py
stuartarchibald: Thanks!

bslatkin:
Understood -- sorry about that! I'll do that next time.

stuartarchibald (approved):
Thanks for the patch and fixes @bslatkin.

esc:
@bslatkin thank you for the patch! @stuartarchibald thank you for the review!

llvmlite/ir/values.py
```diff
@@ -52,6 +52,24 @@ def bitcast(self, typ):
...
+        """
+        Cast this pointer constant to the given integer type.
+        """
+        if not isinstance(typ, types.IntType):
+            raise TypeError(
+                "can only ptrtoint() to integer type, not '%s'"
+                % (typ,))
+        if not isinstance(self.type, types.PointerType):
+            raise TypeError(
+                "can only call inttoptr() on pointer type, not '%s'"
```
gmarkall, line 65: I know this is merged now, but in resolving conflicts for #735, I saw this and found it confusing. Maybe I'm not understanding something, but should this have been `ptrtoint` rather than `inttoptr`?

llvmlite/tests/test_ir.py
```diff
@@ -2225,6 +2225,25 @@ def test_bitcast(self):
...
+        self.assertRaisesRegex(
+            TypeError,
+            r"can only ptrtoint\(\) to integer type, not 'i64\*'",
+            gv.ptrtoint,
+            int64.as_pointer())
+
+        c2 = ir.Constant(int32, 0)
+        self.assertRaisesRegex(
+            TypeError,
+            r"can only call inttoptr\(\) on pointer type, not 'i32'",
```
gmarkall, line 2243: Similar question as above - should this have been `ptrtoint`?

llvmlite/ir/values.py
stuartarchibald: yes!

stuartarchibald: Fixed in https://github.com/numba/llvmlite/pull/794

PR #786, opened 2021-11-15
Update Windows and OSX CI images.

stuartarchibald:
macOS-10.14 and vs2017-win2016 are being deprecated by Azure CI.
This patch updates the images used to the oldest image that is
still supported.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -3,7 +3,7 @@ jobs:
 - template: buildscripts/azure/azure-linux-macos.yml
   parameters:
     name: macOS
-    vmImage: macOS-10.14
+    vmImage: macOS-10.15
     matrix:
       py37:
         PYTHON: '3.7'
@@ -51,4 +51,4 @@ jobs:
 - template: buildscripts/azure/azure-windows.yml
   parameters:
     name: Windows
-    vmImage: vs2017-win2016
+    vmImage: windows-2019


gmarkall (approved):
Given that this passes on current CI and will be necessary in future, I think this looks good.

gmarkall:
@stuartarchibald Any reason I might be missing not to just move this to RTM state?

stuartarchibald:
> @stuartarchibald Any reason I might be missing not to just move this to RTM state?

hmmm, it seems to build ok, I'm not sure why https://github.com/numba/numba/pull/7650 isn't having an impact?

gmarkall:
> hmmm, it seems to build ok, I'm not sure why [numba/numba#7650](https://github.com/numba/numba/pull/7650) isn't having an impact?

What do you mean?

stuartarchibald:
> > hmmm, it seems to build ok, I'm not sure why [numba/numba#7650](https://github.com/numba/numba/pull/7650) isn't having an impact?
> 
> What do you mean?

I'm surprised this OSX base image change didn't trigger the same SDK problems.

gmarkall:
Ah, I see - I think we're not hitting the same issues because we're not using the particular floating point functions that triggered the issue in llvmlite (e.g. `isinf`, `isnan`, etc.)

gmarkall:
Following OOB discussion:

- The compiler bootstrapping process https://github.com/numba/llvmlite/blob/master/buildscripts/incremental/build.sh#L8-L26 seems to isolate llvmlite from the potential of the issue resolved by numba/numba#7650
- There is no evidence of an issue like the one in Numba here
- These changes only relate to the CI setup, so the risks introduced are low - if errors start appearing on CI we can investigate them, this doesn't affect how we build packages for release.
- The chance of preventing some future issue by implementing a workaround like numba/numba#7650 seem pretty low in the absence of any evidence of a problem here in this PR.

Therefore, we conclude that this can / should be merged as-is.

stuartarchibald:
Thanks for recording the discussion @gmarkall .

esc:
@stuartarchibald thank you for the patch! @gmarkall thank you for the review!

PR #789, opened 2021-11-16
don't build docs on every platform, only once

esc:
This will ease the load on the testing runs. Also, it means, we can release for new versions of Python without having to wait for the availability of the doc packages.

cc: #769

### DIFF
--- a/buildscripts/azure/azure-linux-macos.yml
+++ b/buildscripts/azure/azure-linux-macos.yml
@@ -15,6 +15,7 @@ jobs:
 
   steps:
     - script: |
+        set -e
         if [ "$(uname)" == "Linux" ] && [[ "$CONDA_SUBDIR" == "linux-32" || "$BITS32" == "yes" ]]; then sudo apt-get install -y libc6-dev-i386; fi
         echo "Installing Miniconda"
         buildscripts/incremental/install_miniconda.sh
@@ -24,11 +25,13 @@ jobs:
       displayName: 'Before Install'
 
     - script: |
+        set -e
         export PATH=$HOME/miniconda3/bin:$PATH
         buildscripts/incremental/build.sh
       displayName: 'Build'
 
     - script: |
+        set -e
         export PATH=$HOME/miniconda3/bin:$PATH
         conda install -y flake8
         echo "Running flake8 check"
@@ -37,6 +40,7 @@ jobs:
       condition: eq(variables['RUN_FLAKE8'], 'yes')
 
     - script: |
+        set -e
         export PATH=$HOME/miniconda3/bin:$PATH
         conda install -y sphinx sphinx_rtd_theme pygments
         echo "Building docs"
@@ -47,6 +51,7 @@ jobs:
       condition: eq(variables['BUILD_DOCS'], 'yes')
 
     - script: |
+        set -e
         export PATH=$HOME/miniconda3/bin:$PATH
         buildscripts/incremental/test.sh
       displayName: 'Test'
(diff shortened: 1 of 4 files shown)


stuartarchibald:
Thanks for the patch. I'm not sure the AZP script is doing the right thing at present, looking at the raw log: https://dev.azure.com/numba/ff1fe4d0-ed73-4f1c-b894-1d50a27e048f/_apis/build/builds/10219/logs/36 the last few lines have:
```
2021-11-16T16:54:49.1920132Z Building docs
2021-11-16T16:54:49.1923777Z /home/vsts/work/_temp/4db2e7f6-3914-4151-9c14-f741281e04fd.sh: line 5: clean: command not found
2021-11-16T16:54:49.1983274Z ##[section]Finishing: Build Docs
```

buildscripts/azure/azure-linux-macos.yml
```diff
@@ -36,6 +36,16 @@ jobs:
       displayName: 'Flake8'
       condition: eq(variables['RUN_FLAKE8'], 'yes')
 
+    - script: |
+        export PATH=$HOME/miniconda3/bin:$PATH
+        conda install -y sphinx sphinx_rtd_theme pygments
+        echo "Building docs"
+        cd docs
+        SPHINXOPTS=-Wn clean html
```
stuartarchibald, line 44:
```suggestion
        make SPHINXOPTS=-Wn clean html
```
think this is the problem with the "missing command"

buildscripts/azure/azure-linux-macos.yml
```diff
@@ -36,6 +36,16 @@ jobs:
       displayName: 'Flake8'
       condition: eq(variables['RUN_FLAKE8'], 'yes')
 
+    - script: |
+        export PATH=$HOME/miniconda3/bin:$PATH
+        conda install -y sphinx sphinx_rtd_theme pygments
```
stuartarchibald, line 41:
```suggestion
        conda install -y sphinx sphinx_rtd_theme pygments make
```

stuartarchibald (commented):
Couple of suggestions that might fix the build. Also, do AZP scripts support `set -e` ?

buildscripts/azure/azure-linux-macos.yml
esc: yes, a `make` is missing

esc:
> Couple of suggestions that might fix the build. Also, do AZP scripts support `set -e` ?

These are shell scripts. I think they should? Or do you mean the inlined shell commands in the `.yml`?

stuartarchibald:
> > Couple of suggestions that might fix the build. Also, do AZP scripts support `set -e` ?
> 
> These are shell scripts. I think they should? Or do you mean the inlined shell commands in the `.yml`?

The inlined ones, were `set -e` present I think the build would have failed due to the issue in the original patch.

esc:
@stuartarchibald it only needed single patch `make` is already present on the system.

stuartarchibald (approved):

esc:
@stuartarchibald I made a phony commit here, to test the `set -e` via CI. Expecting the 3.9 doc build to fail

esc:
> @stuartarchibald I made a phony commit here, to test the `set -e` via CI. Expecting the 3.9 doc build to fail

I tested it, it failed fine and I removed the temporary commit again.

stuartarchibald:
> @stuartarchibald I made a phony commit here, to test the `set -e` via CI. Expecting the 3.9 doc build to fail

Noted, thanks for checking.

PR #796, opened 2021-11-24
adding changelog for 0.38.0 RC1

esc:
As title

### DIFF
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1,8 +1,10 @@
-v0.38.0 (November 24, 2021)
+v0.38.0 (November 25, 2021)
 ---------------------------
 
-The two biggest new features for this release are the support for Python 3.10
-and the ability to cross-compile for existing LLVM targets.
+This release makes llvmlite compatible with Python 3.10. It also adds an
+``abiname`` option to the target machine creation interface that mimics the same
+in LLVM. Further, a large number of functions are added to the IR API to support
+common uses of constant expressions. Finally, a number of bugs were fixed!
 
 Pull-Requests:
 
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1,7 +1,40 @@
-v0.38.0
--------
+v0.38.0 (November 24, 2021)
+---------------------------
 
-In development
+The two biggest new features for this release are the support for Python 3.10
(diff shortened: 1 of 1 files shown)


esc:
@gmarkall the changelog is now live. Do you perhaps have any suggestions for summarizing the other changes you reviewed? I would like to advertise those in the synopsis of the release. Thank you in advance.

esc:
https://llvmlite--796.org.readthedocs.build/en/796/release-notes.html#v0-38-0-november-24-2021

CHANGE_LOG
```diff
@@ -1,7 +1,40 @@
-v0.38.0
--------
+v0.38.0 (November 24, 2021)
```
stuartarchibald, line 1:
```suggestion
v0.38.0 (November 25, 2021)
```

CHANGE_LOG
```diff
@@ -1,7 +1,40 @@
...
+---------------------------
+
+The two biggest new features for this release are the support for Python 3.10
+and the ability to cross-compile for existing LLVM targets.
```
stuartarchibald, line 5: Maybe something like:
```suggestion
This release makes llvmlite compatible with Python 3.10. It also adds an
``abiname`` option to the target machine creation interface that mimics the same
in LLVM. Further, a large number of functions are added to the IR API to support
common uses of constant expressions. Finally, a number of bugs were fixed!
```
I think the cross compilation support is a feature of how LLVM is built only.

esc:
@stuartarchibald thank you for the suggestions!

gmarkall (approved):
Looks good to me!

esc:
@stuartarchibald @gmarkall I have added an additional change log entry, for #797

gmarkall (approved):
LGTM!

gmarkall:
> DOES llvmlite have riscv support currently,can anyone please tell ..i am trying to install numba so for that dependency i need llvmlite i m facing issue like llvm-config binary not found but binary is already installed in the target.

llvmlite supports RISC-V but Numba does not. If you're having difficulty with using llvmlite, please start a post on the llvmlite section of the Numba Discourse (https://numba.discourse.group/c/llvmlite/12), describing the details of your system, how you installed llvmlite, what you're trying to do, and any outputs you're getting.

PR #801, opened 2021-11-30
Update ffi.py

franzhaas:
- moved from pkg_resources to importlib

i didnt test this, i hope the CI infrastructure of the llvmlite project does it for me.

The benefit i hope to get out of this is that it works wit pyogidizer. Faster startup time if running in an environment that has the setuptools installed..

### DIFF
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -156,9 +156,11 @@ _lib_name = get_library_name()
 
 pkgname = ".".join(__name__.split(".")[0:-1])
 try:
-    with importlib.resources.path(pkgname, _lib_name) as path:
-        lib = ctypes.CDLL(str(path))
-        # outside this context manager the lib may not exist
+    _lib_handle = importlib.resources.path(pkgname, _lib_name)
+    lib = ctypes.CDLL(str(_lib_handle.__enter__()))
+    # on windows file handles to the dll file remain open after
+    # loading, therefore we can not exit the context manager
+    # which might delete the file
 except OSError as e:
     msg = f"""Could not find/load shared object file: {_lib_name}
  Error was: {e}"""
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -177,7 +177,7 @@ try:
     __handle_of_resource_path = importlib.resources.path(__name__, _lib_name)
     _path2library = __handle_of_resource_path.__enter__()
     _lib_paths.append(_path2library)
-except Exception as e:
+except Exception as _:
     pass
 
 # Try to load from all of the different paths
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -172,7 +172,7 @@ _lib_paths = [
 # use importlib.resources, path returns an context manager, in order to make sure that the file remains available, we keep the context manager alive...
 import importlib.resources
 __handle_of_resource_path = importlib.resources.path(__name__, _lib_name)
-_lib_paths.append(resource_filename(next(__handle_of_resource_path))
+_lib_paths.append(resource_filename(next(__handle_of_resource_path)))
 
 
 # Try to load from all of the different paths
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -164,11 +164,7 @@ _lib_name = get_library_name()
 
 
 # Possible CDLL loading paths
-_lib_paths = [
-    os.path.join(_lib_dir, _lib_name),  # Absolute
-    _lib_name,  # In PATH
-    os.path.join('.', _lib_name),  # Current directory
-]
+_lib_paths = []
 
 # use importlib.resources, path returns an context manager, 
 #in order to make sure that the file remains available, we 
@@ -179,6 +175,11 @@ try:
     _lib_paths.append(_path2library)
 except Exception as _:
     pass
+
+_lib_paths.extend([ _lib_name,  # In PATH
+    os.path.join('.', _lib_name),  # Current directory
+])
+
 
 # Try to load from all of the different paths
 errors = []
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -1,6 +1,7 @@
 import ctypes
 import os
 import threading
+import importlib.resources
 
 from llvmlite.binding.common import _decode_string, _is_shutting_down
 from llvmlite.utils import get_library_name
@@ -169,13 +170,14 @@ _lib_paths = [
     os.path.join('.', _lib_name),  # Current directory
 ]
 
-# use importlib.resources, path returns an context manager, in order to make sure that the file remains available, we keep the context manager alive...
-import importlib.resources
+# use importlib.resources, path returns an context manager, 
+#in order to make sure that the file remains available, we 
+#keep the context manager alive...
 try:
     __handle_of_resource_path = importlib.resources.path(__name__, _lib_name)
     _path2library = __handle_of_resource_path.__enter__()
     _lib_paths.append(_path2library)
-except:
+except Exception as e:
     pass
 
 # Try to load from all of the different paths
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -1,5 +1,6 @@
 import ctypes
 import os
+import sys
 import threading
 import importlib.resources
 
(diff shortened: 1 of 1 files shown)


esc:
@franzhaas thank you for submitting this. I have set it to "in progress" for now as there seem to be syntax errors introduced by the change. CI looks like it is picking up the errors.

franzhaas:
I am unsure about my understanding of how the library gets located.

I do believe there are 4 situations in the code.:
library at absolute path
library somewhere in the path
library in working dir 
library in resource of package

Can we reduce these cases?
I do believe that at least "library at absolute path" is redundant to case "library in resource of package".

But for the other two I am curious what they are for...

Can someone provide me some insight please?

franzhaas:
This will work if.:

1.) the ctypes.DLL call loads the library completely and does not access the lib on disk afterwards
2.) the dll is not read from disk on any other places.

Unfortunately i couldnt confirm or disprove both cases.

franzhaas:
I checked point 1, on windows dlls are (at least under some circumstances) used by memory maping the file.

there for we can not delete it after loading the dll....

esc:
@franzhaas thank you for updating this. I am not sure yet, what exactly you are trying to achieve with this PR. I would suggest for you to answer the following, so that I can better understand the intent of this pull-request:

a) What do you do?
b) What do you expect to happen?
c) What happens instead?
d) How does this patch fix it?

Thank you!

franzhaas:
Hi,

First of all, I do not try to fix something. This is not a bug fix this is about convenience, and maintenance (going from helping egg files to helping single file executables).

I encountered issues when i packaged a piece of software that uses numba in a single file executeable.
These issues where due to the presence of setuptools (a dependancy of another package) which triggered usage of pkg_resources in llvmlite, which works badly in single file executables. A very similar issue to the one that was fixed by introducing the pkg_resources usage (egg files support).

pkg_resources is a great solution, which has a long history of success. Starting from Python 3.7 the features of pkg_resources are implemented in standard python.

By using this standard python library i hoped to achieve to load the resource the same way under all circumstances, reducing diversity among installations and use cases, which is a good thing in my point of view.

As far as i can tell, this is working. But this is not a bugfix, it is not a killer feature. 

If you do not think that this change is introducing enough value to justify the effort, i can understand that.

Thanks,
Franz

esc:
@franzhaas thank you for the update. Is this related, by any chance?

https://github.com/numba/numba/pull/5209/files

?

franzhaas:
@esc 
Yes it is replacing the same package this PR tries to replace, with a different part of importlib, 

This change is expected to speedup imports. (although it was not my motivation)

franzhaas:
Is there something i can do to get this either integrated, or to get a educated decision to close it?

I have 2 ideas in my mind.:

1.) Creating a new pull request with a clean history or squash those in this pull request
2.) benchmark if this change improves import time under the following situation.:
- setup tools installed

Can someone point me to a sensible course of action?

gmarkall:
@franzhaas Thanks for pinging - as a holidays task I'm working my way through the llvmlite PR backlog - I'll prioritize this one and get back to you.

gmarkall:
Informal testing on my local machine shows that this does shave about 5% off the import time of llvmlite. I'm checking by running with `python -Ximporttime -c "import llvmlite"` and comparing between master and this branch.

Note: I have to have site packages (can't use `-S`) because I have it installed with `python setup.py develop` in general, but if I used `-S` then I'd expect the percentage to be a bit larger.

I think the two things to help get this over the line are:

- High confidence that this won't break any existing use cases. I note that CI seems fine, but I need to go back and read the discussion to understand if there were any other concerns about this.
- Being able to test the functionality it brings WRT PyOxidizer - is there a simple set of steps I can run to try packaging llvmlite with PyOxidizer with this PR?

gmarkall:
@franzhaas Forgot to say earlier also - many thanks for your efforts so far and for continuing to persist with / push on this PR! :slightly_smiling_face:

gmarkall:
> High confidence that this won't break any existing use cases. I note that CI seems fine, but I need to go back and read the discussion to understand if there were any other concerns about this.

I think to address this the way forward would be to build with this PR on the buildfarm and then test Numba with the resulting package (cc @esc - who is away until next year)

gmarkall:
> 1.) Creating a new pull request with a clean history or squash those in this pull request

I think keeping this PR as-is is fine. In general we try to persuade people not to rewrite history in Numba / llvmlite PRs, because it can confuse Github's reviewing interface.

llvmlite/binding/ffi.py
```diff
@@ -151,47 +151,19 @@ def __call__(self, *args, **kwargs):
             return self._cfn(*args, **kwargs)
 
 
-_lib_dir = os.path.dirname(__file__)
-
-if os.name == 'nt':
-    # Append DLL directory to PATH, to allow loading of bundled CRT libraries
```
gmarkall, line 157: Why / when would CRT libraries be bundled with llvmlite?

gmarkall: (note, this is mainly a question for @sklam / @stuartarchibald / @esc, not necessarily a bit of feedback about the PR - I'm finding it hard to imagine a situation where this makes sense)

gmarkall: Answering my own question - it was added in https://github.com/numba/llvmlite/commit/1f7133c5 which we no longer seem to do (presumably it was needed for the newer libraries needed by LLVM on Python 2 which would have been linked to much older runtimes)

gmarkall (approved):
@franzhaas Having looked at the code and each of the cases above, my thoughts are:

- Addition to the `PATH` variable: this is antiquated and should not need to be considered
- Absolute path / relative path / in resource of package - my understanding is that this should be handled as expected by `importlib.resources.path`

So, to the best of my judgment I think the code changes look good and I would just like to try this on the buildfarm (which will unfortunately have to wait until people are back after the holidays) - if you can let me know how to do a manual test with PyOxidizer in the meantime, that would be great! (I don't think it's essential to test this given it's not something we claim works, but I'd like to try it anwyay)

seibert:
Testing on internal build farm run `llvmlite_126`

seibert:
It passed everywhere, except the transitory issues with some of the Python 3.10 packages (the internal build processes update conda as the first step, so there must be some bad packages upstream) and our win-32 worker is down, so those tests aren't running:

Airflow

franzhaas:
> @franzhaas Having looked at the code and each of the cases above, my thoughts are:
> 
> * Addition to the `PATH` variable: this is antiquated and should not need to be considered
To my understanding, as long as python 3.7 is suported, this needs to be donne.

> * Absolute path / relative path / in resource of package - my understanding is that this should be handled as expected by `importlib.resources.path`
Yes you are right. There is the wiredness that on windows it is not a good idea to delete an open file. 

This can happen when having the dll packed into a container file, then when using a importlib context manager, it gets copied onto a file system, used and deleted on exit of the context manager. 

> 
> So, to the best of my judgment I think the code changes look good and I would just like to try this on the buildfarm (which will unfortunately have to wait until people are back after the holidays) - if you can let me know how to do a manual test with PyOxidizer in the meantime, that would be great! (I don't think it's essential to test this given it's not something we claim works, but I'd like to try it anwyay)

I put creating a pyoxidizer example onto my holiday todo list...

Thanks for this great piece of software!

gmarkall:
Many thanks for the additional info!

> * Addition to the `PATH` variable: this is antiquated and should not need to be considered
>       To my understanding, as long as python 3.7 is suported, this needs to be donne.

Sorry, this is something I've missed / misunderstood - I got the impression this was only done for bundling a MSVC runtime, which we no longer do. What is it about Python 3.7 that requires this?

> > * Absolute path / relative path / in resource of package - my understanding is that this should be handled as expected by `importlib.resources.path`
> >   Yes you are right. There is the wiredness that on windows it is not a good idea to delete an open file.
> 
> This can happen when having the dll packed into a container file, then when using a importlib context manager, it gets copied onto a file system, used and deleted on exit of the context manager.

I may have been a little unclear here - my intention was to imply that I think your code changes are fine. I think remaining in the context manager is fine as we have no way to unload llvmlite and free everything up anyway (AFAIK).

> I put creating a pyoxidizer example onto my holiday todo list...

Many thanks - if it's loads of work and you'd rather have a holiday, please have a holiday instead of fiddling with pyoxidizer - it will be nice to have the example, but not essential.

> Thanks for this great piece of software!

Many thanks for your contributions! :slightly_smiling_face:

franzhaas:
> 
> Sorry, this is something I've missed / misunderstood - I got the impression this was only done for bundling a MSVC runtime, which we no longer do. What is it about Python 3.7 that requires this?
> 

I was msitaken. The path change is only needed on python 3.7 if we need to load a dll and its dependencies when the dependencis are not in the standard dll search path.

in llvmlites situation, this is not the case.

franzhaas:
hi,

I have created an example that shows the advantage of this pull request for pyoxidizer. It is a very hacky.

As i can not build proper wheels (llvmlite wont compile due to issues with cmake on my machine) i modified a downlaoded wheel to compare it with the current std llvmlite.

https://github.com/franzhaas/llvmlite_pyoxidizer_example

in order to create the exe you need to 
first pipenv sync
and then
cd into the directory of the example you are interested in and.:

pipenv run pyoxidizer run --target install

I personaly dont think that it makes sense to place it in the example directory or add it to the CI. 

Best regards,
Franz

gmarkall:
@franzhaas Many thanks for the example - I'll check it out and see if I can run it both ways.

I think we need to fix our build setup so we can be sure this works with Python 3.10 then re-run CI, so I'll leave this in the "Pending CI" state - but other than that I still think this looks good.

gmarkall:
Also "Waiting on reviewer" for me to try out the example.

franzhaas:
I recently noted that numpy drops pyhton 3.7 suport.

https://numpy.org/neps/nep-0029-deprecation_policy.html

This builds an argument to remove the python3.7 suport from this PR, as numpy is a dependency of numba.

What is numbas policy here?

gmarkall:
I think we might have been planning to drop support for 3.7 in Numba 0.56 / llvmlite 0.39, but @esc / @stuartarchibald may recall more clearly.

stuartarchibald:
> What is numbas policy here?

This is up for discussion for Numba 0.56 and hasn't been decided. Numba does track NumPy's NEP 0.29 but has historically had a slower schedule, this with view of helping out HPC users where upgrades to Python versions can sometimes take a while.

seibert:
I would definitely lean toward being slow to drop Python 3.7 (at least for a few more NumPy releases) unless it requires a non-trivial amount of code to work around its limitations or is becoming difficult to test.

gmarkall:
I understand the buildfarm is now working with Python 3.10 - if that's the case, could this have another run please @seibert / @esc / @stuartarchibald ?

esc:
`Build llvmlite_127 has started`

esc:
> `Build llvmlite_127 has started`

The build was fine.

gmarkall:
Thanks for the build @esc!

I think we're good to merge this now - any objections?

esc:
> Thanks for the build @esc!
> 
> I think we're good to merge this now - any objections?

Absolutely, I'll mark as RTM and merge during the next merge window.

esc:
Thank you for the patch!

PR #806, opened 2022-01-10
Bump to v0.39.0dev

esc:
As title


gmarkall (approved):
This looks good to me (on the basis that it's doing the same as the previous PR for this task, https://github.com/numba/llvmlite/pull/739)

PR #807, opened 2022-01-10
Exclude ExecutionEngine tests on linux 32

esc:
Since we started building llvmdev with all backend target architectures
included (for binary wheels and binary conda packages) we noticed that
under Linux 32 we are not able to generate ppc64 code correctly. This is
viewed as a minor annoyance since it is a very unlikely use-case in
practice. Thus, we simply exclude the failing tests on Linux 32 so as to
be able to build and ship an llvmdev anyway.

For reference, the error messages looked like:

```
INFO - Failed Tests (2):
INFO - LLVM :: ExecutionEngine/RuntimeDyld/PowerPC/ppc64_elf.s
INFO - LLVM :: ExecutionEngine/RuntimeDyld/PowerPC/ppc64_reloc.s
```

### DIFF
--- a/conda-recipes/llvmdev_manylinux2014/build.sh
+++ b/conda-recipes/llvmdev_manylinux2014/build.sh
@@ -81,5 +81,10 @@ fi
 # From: https://github.com/conda-forge/llvmdev-feedstock/pull/53
 make install || exit $?
 
+# run the tests, skip some on linux-32
 cd ../test
-../build/bin/llvm-lit -vv Transforms ExecutionEngine Analysis CodeGen/X86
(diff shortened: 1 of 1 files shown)


esc:
note that this was included in `llvmdev` build number 4 and is submitted now for completeness. I.e. no `llvmdev` rebuild should be required.

esc:
EDIT: the description has been updated to indicate `linux-32` not `win-32`.

stuartarchibald (approved):
Thanks for the patch.

esc:
Thank you for the patch!

PR #812, opened 2022-01-20
Update versioneer to enable Python 3.11

oscargus:
Updated versioneer as old versions of it use `configparser.SafeConfigParser` which is removed in Python 3.11.


oscargus:
There is a hack in here to avoid updating the newly copied `_version.py` file since it is no longer copied before trying to update it.

I _think_ it is related to executing `build_ext` as part of `build`, which no longer seems possible (looking at the previous tests, it is clear that `build_ext` did not try to update this file, which the new version seems to do) but really no idea. Probably it is better to change in setup.py, but then this will really only become a problem next time versioneer has to be updated...

versioneer.py
```diff
@@ -618,429 +1125,987 @@ def git_versions_from_keywords(keywords, tag_prefix, verbose=False):
...
+                # build/lib<..> dir with no _version.py to write to.
+                # As in place builds will already have a _version.py
+                # in the module dir, we do not need to write one.
+                return
+            # now locate _version.py in the new build/ directory and replace
+            # it with an updated value
+            target_versionfile = os.path.join(self.build_lib,
+                                              cfg.versionfile_build)
+            # Hand-edited to get around build problem
+            if os.path.isfile(target_versionfile):
```
oscargus: This line is manually added to make the builds succeed.

oscargus:
Updated to latest version.

esc:
@oscargus we are starting to prep for 3.11 support slowly. We haven't noticed any failures in our initial testing. Do you have an idea where and why this may break on 3.11 w/o this patch? Thank you!

oscargus:
I do not really recall anything other than what I stated originally. It uses `configparser.SafeConfigParser` which apparently is removed in Python 3.11.

But if it works, then I guess that there is no immediate point in changing. However, from a quick check it doesn't look like you are testing with 3.11 now (nor any PR?)?

Anyway, if you are in that process, I can try to update once again since I think that there is a newer version of versioneer.

versioneer.py
oscargus: Seems like it is not required in newer versions.

esc:
> I do not really recall anything other than what I stated originally. It uses `configparser.SafeConfigParser` which apparently is removed in Python 3.11.

OK, we will have to double check.

> But if it works, then I guess that there is no immediate point in changing. However, from a quick check it doesn't look like you are testing with 3.11 now (nor any PR?)?

It was a bit tricky to test with Python 3.11 since Numba/llvmlite are a bit unique in this respect. We have been testing against the RCs for a few months however. The results and information on how we achieved this are here:

https://github.com/numba/numba/issues/8304

> Anyway, if you are in that process, I can try to update once again since I think that there is a newer version of versioneer.

Probably a good idea, thank you.

oscargus:
I checked a bit further and it may be that your version of versioneer is so old so that it doesn't even use `SafeConfigParser`... So in that case there may be no need for it. 

Unless of course you want to bump it for other reasons... :-)

(Replacing
```
* Compatible With: python2.6, 2.7, 3.2, 3.3, 3.4, and pypy
```
with 
```
* Compatible with: Python 3.7, 3.8, 3.9, 3.10 and pypy3
```
is a possible motivation as well, and I assume that it has developed. But does indeed not seem required.)


(I made an effort updating versioneer once I was informed about the issue in a number of relevant packages, although I must admit that I didn't really consider this case.)

oscargus:
Note that it doesn't state 3.11, but it is the latest available version (0.27) and it doesn't use `SafeConfigParser`.

oscargus:
Maybe. They just added formal testing for 3.11 https://github.com/python-versioneer/python-versioneer/pull/348

All I can say is that other packages have upgraded and works well. Here is the PR that started it: https://github.com/aesara-devs/aesara/pull/721 

You are on 0.12, `SafeConfigParser` was introduced in 0.15 https://github.com/python-versioneer/python-versioneer/commit/1fe079a9e0e9f4a7b7b748062ccadf08b1912544, and the bug has been fixed since 0.19(?) https://github.com/python-versioneer/python-versioneer/pull/207

esc:
> Maybe. They just added formal testing for 3.11 [python-versioneer/python-versioneer#348](https://github.com/python-versioneer/python-versioneer/pull/348)
> 
> All I can say is that other packages have upgraded and works well. Here is the PR that started it: [aesara-devs/aesara#721](https://github.com/aesara-devs/aesara/pull/721)
> 
> You are on 0.12, `SafeConfigParser` was introduced in 0.15 [python-versioneer/python-versioneer@1fe079a](https://github.com/python-versioneer/python-versioneer/commit/1fe079a9e0e9f4a7b7b748062ccadf08b1912544), and the issue has been fixed since 0.19(?) [python-versioneer/python-versioneer#207](https://github.com/python-versioneer/python-versioneer/pull/207)

OK, well, if nothing is actually broken broken, then perhaps we could just wait until smth. actually breaks?

oscargus:
Your call for sure.

(I usually update dependencies for the sake of it every now and then, unless it breaks something, but there is of course also the other way around. And this is not really a dependency in the traditional way either.)

esc:
> Your call for sure.
> 
> (I usually update dependencies for the sake of it every now and then, unless it breaks something, but there is of course also the other way around. And this is not really a dependency in the traditional way either.)

Once we get CI testing for llvmlite for Python 3.11 going I'd suggest to try to update the vendored versioneer code to a Python 3.11 compatible version. Does that make sense?

You can subscribe to https://github.com/numba/llvmlite/issues/885 for updates on when this will happen.

esc:
@oscargus llvmlite 0.40.0 was released already and supports Python 3.11 -- do you want to give this another spin.

gmarkall:
There doesn't seem to be a versioneer-related issue at present, and there's been no activity here for a couple of years. If any new versioneer changes are to be made, I would suggest a new PR.

PR #815, opened 2022-01-31
update local references

esc:
As title


stuartarchibald (approved):
Thanks for the patch. I've run the `git grep` locally and observed that the remaining references are as noted in https://github.com/numba/llvmlite/pull/815#issuecomment-1025528498.

Issue #814, opened 2022-01-31
Rename `master` to `main`.


esc:
Similar to: https://github.com/numba/numba/issues/7039

The notes on https://github.com/github/renaming seem to suggest that using the "branch renaming" feature will largely do most of the work for us.

* [x] create PR to rename any references inside the repo
* [x] rename branch on github with UI
* [x] merge #815

esc:
The local rename PR is at: https://github.com/numba/llvmlite/pull/815

esc:
<img width="367" alt="Screen Shot 2022-01-31 at 14 17 51" src="https://user-images.githubusercontent.com/41563/151800384-ff1037fd-ae0d-431e-824a-bb8cab50a23b.png">

esc:
Looks like it went through.

<img width="469" alt="Screen Shot 2022-01-31 at 14 18 47" src="https://user-images.githubusercontent.com/41563/151800447-45730bae-9e9c-4a95-89a0-f3798a74c084.png">

esc:
This task is now complete. Please followup on this ticket, in case you encounter any issues!

stuartarchibald:
Thanks @esc

esc:
oh dear, it looks like Azure CI broke and sombody with appropriate permissions must log in to fix it!

esc:
> oh dear, it looks like Azure CI broke and sombody with appropriate permissions must log in to fix it!

need to tell Azure somehow that `main` must be built.

esc:
CI is working again

PR #816, opened 2022-01-31
remove configuration landscape service as it is no longer used

esc:
As title


esc:
Followup from #814

stuartarchibald (approved):
Thanks for the patch. I don't think this is used by anything current.

PR #817, opened 2022-02-01
remove uppper limit on Python requires

esc:
As title

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -223,6 +223,5 @@ setup(name='llvmlite',
       license="BSD",
       cmdclass=cmdclass,
       long_description=long_description,
-      python_requires=">={},<{}".format(min_python_version,
-                                        max_python_version),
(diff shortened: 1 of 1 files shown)


sklam (approved):
LGTM

esc:
For reference, the discussion leading up to this change are here: https://github.com/numba/numba/pull/7783#issuecomment-1088090391

The `TL;DR` on this is: if this patch were in released version of llvmlite and if Python 3.11 was suddenly released today, then: anyone trying to install llvmllite with pip would be able to install it fine, but would receive a runtime warning that the version of Python isn't compatible (even though it probably would be, but that is for another ticket). This was deemed better than the previous behaviour, which would cause pip to look at previous versions of llvmlite until it found one that doesn't pin Python this way (some ancient version) and then try to install that, either failing hopelessly or leaving the user confused.

PR #820, opened 2022-02-02
M1 support

esc:
For testing only.


esc:
it is unlikely that any of these changes were needed.

PR #813, opened 2022-01-24
Add m1 support to conda build scripts

seibert:
Minor adjustments needed to address differences in clang version and SDK version for osx-aarch64.

PR #821, opened 2022-02-07
M1 support (using llvmdev build 5 for llvmlite) 

esc:


esc:
Was only testing for #813 relevant patch has been 🍒 so closing this.

PR #824, opened 2022-03-01
add GitHub URL for PyPi

andriyor:
Warehouse now uses the project_urls provided to display links in the sidebar on [this screen](https://pypi.org/project/requests/), as well as including them in API responses to help automation tool find the source code for Requests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -219,6 +219,9 @@ setup(name='llvmlite',
       ],
       # Include the separately-compiled shared library
       url="http://llvmlite.readthedocs.io",
+      project_urls={
+          "Source": "https://github.com/numba/llvmlite",
+      },
       packages=packages,
       license="BSD",
       cmdclass=cmdclass,


esc:
> Warehouse now uses the project_urls provided to display links in the sidebar on [this screen](https://pypi.org/project/requests/), as well as including them in API responses to help automation tool find the source code for Requests.

Thank you for submitting this. I don't see any reason not to include this. Is there documentation anywhere? Do we want to test this during the next release candidate phase?

andriyor:
Docs: https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls

esc:
> Docs: https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls

Cool, can you add `Documentation` -- via read the docs -- too? The other examples don't seem applicable. Can we just make up categories?

esc:
Screen Shot 2022-04-12 at 13 20 04

I am not sure the `url` field is actually working, see above screenshot. The link `homepage` points to:

https://numba.pydata.org/

Which isn't actually mentioned in `setup.py`. So, I am left wondering where this came from. My tendency would be to perhaps remove `url` and include it in `project_urls`. @andriyor what do you think? Also, should we do this for `llvmlite` also?

esc:
@andriyor I am so sorry, I just realized, this PR IS for llvmlite NOT Numba. The `url` is also correct. Now the question remains if we should add this Numba too?

esc (approved):
LGTM

PR #825, opened 2022-03-10
Add flag handling to more instructions.

apmasell:
As title.

Closes https://github.com/numba/llvmlite/issues/798
Modification of https://github.com/numba/llvmlite/pull/709

### DIFF
--- a/llvmlite/ir/builder.py
+++ b/llvmlite/ir/builder.py
@@ -555,7 +555,7 @@ class IRBuilder(object):
         return self.sub(values.Constant(value.type, 0), value, name=name)
 
     @_unop('fneg')
-    def fneg(self, arg, name=''):
+    def fneg(self, arg, name='', flags=()):
         """
         Floating-point negative:
             name = -arg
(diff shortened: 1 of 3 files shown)


stuartarchibald:
Thanks for the patch. Is it worth checking other instructions for flag handling and getting them all done in one go? xref https://github.com/numba/llvmlite/pull/709#issuecomment-800290020

apmasell:
I missed that comment in the original. I'll have a look.

apmasell:
Okay. I checked which LLVM instructions support fast-math and fneg is included; everything else has been covered. The fneg function is the only one to use `_unop`...is that something that other unary instructions should be transitioning to? It seems to have flags support anyway.

Also, I realised a number of the changes used a default list parameter, so I changed them to an empty tuple to avoid mutability issues.

stuartarchibald:
> Okay. I checked which LLVM instructions support fast-math and fneg is included; everything else has been covered.

Thanks for checking.

>The fneg function is the only one to use `_unop`...is that something that other unary instructions should be transitioning to? It seems to have flags support anyway.

I don't think there are any other unary operations in the builder API that have directly equivalent LLVM operations? Seems like the `_*op*('<operation name>')` decorator family rely on there being a named operation in LLVM.

> Also, I realised a number of the changes used a default list parameter, so I changed them to an empty tuple to avoid mutability issues.

Thanks.

stuartarchibald:
@sklam could you take a look at this please, whilst minor, this is a modification/continuation of a patch I authored https://github.com/numba/llvmlite/pull/709. Many thanks.

sklam (approved):
LGTM!

@apmasell, congrats on getting your first PR for merge for llvmlite!

PR #829, opened 2022-03-25
Support loading llvmlite from the main executable.

folded:
Change library loading code to also check for the presence of llvmlite
statically linked into the main executable if the dynamic library fails
to load. This allows llvmlite to be used in hermetic and embedded
build cases.

Check that the version function can be dynamically resolved before
declaring success.

Exit the resource context manager correctly so that any extracted
files are unlinked correctly.

On failure, raise OSError with the first error chained.

### DIFF
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -153,6 +153,12 @@ class _lib_fn_wrapper(object):
 
 
 _lib_load_errors = []
+# llvmlite native code may exist in one of two places, with the first taking
+# priority:
+# 1) In a shared library available as a resource of the llvmlite package.
+#    This may involve unpacking the shared library from an archive.
+# 2) Linked directly into the main binary. Symbols may be resolved from the
+#    main binary by passing None as the argument to ctypes.CDLL.
 for lib_context in (
         importlib.resources.path(
             __name__.rpartition(".")[0], get_library_name()),
@@ -164,8 +170,14 @@ for lib_context in (
             version_info = lib.LLVMPY_GetVersionInfo()
             break
     except (OSError, AttributeError) as e:
+        # OSError may be raised if the file cannot be opened, or is not
+        # a shared library.
+        # AttributeError is raised if LLVMPY_GetVersionInfo does not exist.
         _lib_load_errors.append(e)
 else:
+    # None of the expected locations contains a loadable version of llvmlite.
+    # raise OSError with the first error seen during the load event as the
+    # cause of this exception.
     raise OSError(
         "Could not find/load shared object file") from _lib_load_errors[0]
 
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -159,7 +159,7 @@ for lib_context in (
         contextlib.nullcontext(None)):
     try:
         with lib_context as lib_path:
-            lib = ctypes.CDLL(lib_path)
+            lib = ctypes.CDLL(lib_path and str(lib_path))
             # Check that we can look up expected symbols.
             version_info = lib.LLVMPY_GetVersionInfo()
             break
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -79,6 +79,19 @@ class _LLVMLock:
         self._lock.release()
 
 
+class _suppress_cleanup_errors:
+    def __init__(self, context):
+        self._context = context
+
+    def __enter__(self):
+        return self._context.__enter__()
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        try:
+            return self._context.__exit__(exc_type, exc_value, traceback)
+        except PermissionError:
+            pass  # Resource dylibs can't be deleted on Windows.
+
 class _lib_wrapper(object):
     """Wrap libllvmlite with a lock such that only one thread may access it at
     a time.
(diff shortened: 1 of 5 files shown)


esc:
> Change library loading code to also check for the presence of a statically linked version of llvmlite if the dynamic library fails to load. This allows llvmlite to be used in hermetic and embedded build cases.

Thank you for adding this. There are a few ways to build llvmlite. By default the Numba team will statically link LLVM in both wheels and conda packages. See also:

https://llvmlite.readthedocs.io/en/latest/admin-guide/install.html#pre-built-binaries

Hence I am a little confused as to what benefit this PR brings. May I kindly request, that you describe in slightly more detail: what problems did you encounter and how does this help with the build? Thank you.

folded:
I modified the PR description slightly to make the intention hopefully more clear.

This CL addresses the case where you link llvmlite (and llvm) directly into the main executable. You might do this, for example, if you're using an embedded python runtime and don't want to distribute lots of additional shared libraries. It's admittedly a niche use-case (although an important one for us).

I'm happy to maintain this patch downstream if you feel like it's not sufficiently general,. but I do think there are some other small quality of life improvements in this code and it would be pretty easy to just drop the `contextlib.nullcontext(None)` that supports looking up the symbols in the main executable.

esc:
> I modified the PR description slightly to make the intention hopefully more clear.
> 
> This CL addresses the case where you link llvmlite (and llvm) directly into the main executable. You might do this, for example, if you're using an embedded python runtime and don't want to distribute lots of additional shared libraries. It's admittedly a niche use-case (although an important one for us).
> 
> I'm happy to maintain this patch downstream if you feel like it's not sufficiently general,. but I do think there are some other small quality of life improvements in this code and it would be pretty easy to just drop the `contextlib.nullcontext(None)` that supports looking up the symbols in the main executable.

OK, if this patch has value for you I don't mind considering it. It's just important that this does not impact what is currently working. Also, would you be open to adding more comments to the source code? There are a few lines where it isn't immediately obvious to me what is going on and they could IMHO do with some additional comments.

folded:
Done. LMK if there's anything that needs more clarification.

esc:
> Done. LMK if there's anything that needs more clarification.

This looks better, thank you.

llvmlite/binding/ffi.py
```diff
@@ -151,20 +152,34 @@ def __call__(self, *args, **kwargs):
...
 
 
-_lib_name = get_library_name()
-
-
-pkgname = ".".join(__name__.split(".")[0:-1])
-try:
-    _lib_handle = importlib.resources.path(pkgname, _lib_name)
-    lib = ctypes.CDLL(str(_lib_handle.__enter__()))
-    # on windows file handles to the dll file remain open after
```
esc, line 161: I'm not sure about this change yet. The comment seems to indicate that the context manager can not be used on Windows.

llvmlite/binding/ffi.py
```diff
@@ -151,20 +152,34 @@ def __call__(self, *args, **kwargs):
...
-    msg = f"""Could not find/load shared object file: {_lib_name}
- Error was: {e}"""
-    raise OSError(msg)
+_lib_load_errors = []
+# llvmlite native code may exist in one of two places, with the first taking
+# priority:
+# 1) In a shared library available as a resource of the llvmlite package.
+#    This may involve unpacking the shared library from an archive.
+# 2) Linked directly into the main binary. Symbols may be resolved from the
+#    main binary by passing None as the argument to ctypes.CDLL.
```
esc, line 161: Under what circumstances do we pass `None`?

llvmlite/binding/ffi.py
```diff
@@ -151,20 +152,34 @@ def __call__(self, *args, **kwargs):
...
+for lib_context in (
+        importlib.resources.path(
+            __name__.rpartition(".")[0], get_library_name()),
+        contextlib.nullcontext(None)):
+    try:
+        with lib_context as lib_path:
+            lib = ctypes.CDLL(lib_path and str(lib_path))
```
esc, line 168: What is the intent of using `and` here. Is this something special about the type of `lib_context?

llvmlite/binding/ffi.py
folded: Isn't it ok to delete a file that has open file handles on Windows? This is definitely the way you would want this to behave on posix systems, as it's totally fine to keep a file descriptor open and unlink the file. This has the advantage that the filesystem is cleaned up.

I'm investigating this to check the windows behaviour.

esc: @folded thank you for looking into this, if you can work out if or if not this comment is/was relevant and can explain it too, that would be most helpful!

llvmlite/binding/ffi.py
```diff
@@ -77,28 +79,96 @@ def __exit__(self, *exc_details):
...
+        #    from the main binary by passing None as the argument to
+        #    ctypes.CDLL.
+        for lib_context in (
+                _suppress_cleanup_errors(importlib.resources.path(
+                    __name__.rpartition(".")[0], get_library_name())),
+                contextlib.nullcontext(None)):
+            try:
+                with lib_context as lib_path:
+                    if os.name == 'nt' and not lib_path:
+                        # The Windows implementation of ctypes.CDLL does not
```
esc, line 124: Question here: does this mean that Windows doesn't support the kind of static linking?

esc:
@folded it's been a while since I last heard from you, how do you want to proceed with this one?

folded:
My work for this week is to do the internal upgrade to 0.39.0, and as a result I'll be going through these patches again and cleaning them up. I'll get back to you soon.

esc:
> My work for this week is to do the internal upgrade to 0.39.0, and as a result I'll be going through these patches again and cleaning them up. I'll get back to you soon.

Thank you for the update!

folded:
I've split this into 3 PRs, starting with folded:refactor-native-library-loading

PR #599, opened 2020-06-17
LLVM 10

esc:
As title, upgrade to LLVM 10.

PR #830, opened 2022-04-05
Add LLVM 14 support

apmasell:


stuartarchibald:
Thanks for starting work on LLVM 14 support. In thinking about how to make it easier to review the changes needed, perhaps the following approach might work:
* Creating a separate patch for the C++ code formatting changes, this gets merged first so as to minimise future diffs and to make it easy to spot code changes that are real vs. formatting.
* Subsequently adding support for LLVM 14 including build and patch updates as a separate PR, then merging this next.
* Finally adding support for the addition LLVM pass bindings/wrappers and merging this last.

What do you think?

apmasell:
Sure. I've created #831 for the code formatting changes. I'm going to leave this PR open so I can use it for the build farm.

stuartarchibald:
> Sure. I've created #831 for the code formatting changes. I'm going to leave this PR open so I can use it for the build farm.

Thanks, keeping this open for testing purposes sounds like a good plan.

Hardcode84:
Hi,

Here is svml patch for llvm 14 https://github.com/Hardcode84/llvm-project/commit/bc2dcd190b7148d04772fa7fcd18b5200b758d4a

conda-recipes/llvmdev/meta.yaml
```diff
@@ -1,32 +1,19 @@
-{% set shortversion = "11.1" %}
-{% set version = "11.1.0" %}
-{% set sha256_llvm = "ce8508e318a01a63d4e8b3090ab2ded3c598a50258cc49e2625b9120d4c03ea5" %}
-{% set sha256_lld = "017a788cbe1ecc4a949abf10755870519086d058a2e99f438829aef24f0c66ce" %}
+{% set shortversion = "14.0" %}
+{% set version = "14.0.1" %}
+{% set sha256_llvm = "1a3c2e57916c5a70153aaf0a0e6f1230d6368b9e0f4d04dcb9e039a31b1cd4e6" %}
 {% set build_number = "5" %}
```
sklam, line 4: rmbr to reset this before merge

conda-recipes/llvmlite/meta.yaml
```diff
@@ -30,9 +30,9 @@ requirements:
   host:
     - python
     # On channel https://anaconda.org/numba/
-    - llvmdev 11.1.0 *5 # [(osx and arm64)]
-    - llvmdev 11.1.0 *4 # [not ((osx and arm64) or win)]
-    - llvmdev 11.1.0 4 # [win]
+    - llvmdev 14.0.1 *5 # [(osx and arm64)]
+    - llvmdev 14.0.1 *4 # [not ((osx and arm64) or win)]
+    - llvmdev 14.0.1 4 # [win]
```
sklam, line 35:
```suggestion
    - llvmdev 14.0.1
```

detrout:
Thanks for pointing me here from the llvmlite patch.

How important are the patches to llvm itself? There seems to be a bunch of changes in the conda recipes, are those needed outside of a conda environment?

apmasell:
The patches to LLVM 14 that are included with the recipe are pretty important. Numba uses SVML and that is not in LLVM mainline and needs to be patched in. Some of the other changes are for building on different environments (MacOS and Windows can be uncooperative).

MehdiChinoune:
Could you update the PR to support LLVM-15, please.

apmasell:
That's not on our roadmap yet. LLVM 15 has a number of breaking changes in the way it handles pointers, and we aren't prepared to handle those yet.

conda-recipes/llvmdev/bld.bat
```diff
@@ -24,31 +34,18 @@ REM the 64bit linker anyway. This must be passed in to certain generators as
...
-REM correct architecture. Also, using Visual Studio 16 2019 we use toolset
-REM v141, which basically means use a Visual Studio 15 2017 type compiler from
-REM Visual Studio 16 2019. See also:
-REM https://stackoverflow.com/questions/55708600/whats-the-cmake-generator-for-visual-studio-2019
+set MAX_INDEX_CMAKE_GENERATOR=0
 
-set "CMAKE_GENERATOR[0]=Visual Studio 14 2015%ARCH_POSTFIX%"
-set "CMAKE_GENERATOR[1]=Visual Studio 15 2017%ARCH_POSTFIX%"
-set "CMAKE_GENERATOR[2]=Visual Studio 16 2019"
+set "CMAKE_GENERATOR[0]=Visual Studio 17 2022"
```
sklam, line 39:
```
set "CMAKE_GENERATOR[0]=Visual Studio 16 2019"
```

conda-recipes/llvmdev/bld.bat
```diff
@@ -24,31 +34,18 @@ REM the 64bit linker anyway. This must be passed in to certain generators as
...
 
-set "CMAKE_GENERATOR_ARCHITECTURE[0]="
-set "CMAKE_GENERATOR_ARCHITECTURE[1]="
-set "CMAKE_GENERATOR_ARCHITECTURE[2]=%GEN_ARCH%"
+set "CMAKE_GENERATOR_ARCHITECTURE[0]=%GEN_ARCH%"
 
-set "CMAKE_GENERATOR_TOOLSET[0]=host %PreferredToolArchitecture%"
-set "CMAKE_GENERATOR_TOOLSET[1]=host  %PreferredToolArchitecture%"
-set "CMAKE_GENERATOR_TOOLSET[2]=v141"
+set "CMAKE_GENERATOR_TOOLSET[0]="
```
sklam, line 43:
```
set "CMAKE_GENERATOR_TOOLSET[0]=v142"
```

sklam (commented):
@apmasell, try vs2019 and pinning to 14.2 toolset:

ffi/custom_passes.cpp
```diff
@@ -920,27 +920,63 @@ struct RefPrunePass : public FunctionPass {
...
+            if (int_operand && !int_operand->isZeroValue()) {
+                return true;
+            }
+            // If the operand is a PHI node, check if the path we're on will
+            // yield a value indicating an exception
+            auto phi_operand = dyn_cast<PHINode>(operand);
+            if (phi_operand) {
+                auto arg_value = dyn_cast<ConstantInt>(
+                    phi_operand->getIncomingValueForBlock(bb));
+                return arg_value && !arg_value->isZeroValue();
```
sklam, line 974: same as earlier, not all non-zero value means exception

ffi/custom_passes.cpp
```diff
@@ -920,27 +920,63 @@ struct RefPrunePass : public FunctionPass {
...
+            auto int_operand = dyn_cast<ConstantInt>(operand);
+            if (int_operand && !int_operand->isZeroValue()) {
+                return true;
+            }
+            // If the operand is a PHI node, check if the path we're on will
+            // yield a value indicating an exception
+            auto phi_operand = dyn_cast<PHINode>(operand);
+            if (phi_operand) {
+                auto arg_value = dyn_cast<ConstantInt>(
+                    phi_operand->getIncomingValueForBlock(bb));
```
sklam, line 973: Look at PHI value for the return code is clever. That may save us needing the metadata all together.

sklam (commented):
Before proceeding, we should get more cases. Particular, we can to see if there are other patterns if multiple exceptions of different exception type are raised.

e.g.

```python
if cond_a:
    raise ValueError
elif cond_b:
    raise TypeError
else:
    return ok
```

ffi/custom_passes.cpp
```diff
@@ -920,27 +920,63 @@ struct RefPrunePass : public FunctionPass {
...
+                return false;
+            }
+            // With one operand
+            if (first->getNumOperands() != 1) {
+                return false;
+            }
+            auto operand = first->getOperand(0);
+            // If the operand is a constant, check if it indicates an exception
+            auto int_operand = dyn_cast<ConstantInt>(operand);
+            if (int_operand && !int_operand->isZeroValue()) {
```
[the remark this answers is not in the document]
sklam: See https://github.com/numba/numba/blob/9d4f82df7f0426d89b189eafc7acdc6bf719c716/numba/core/callconv.py#L473-L479. It's Numba handling the return code. For this refprune case, we used to only care about user exception. The existing Numba code seems to only ever emit `1`. Numba callconv may need to clean up.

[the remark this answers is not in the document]
apmasell: The case that is failing in the Numba repository emits a `1`, so I'm not sure what's going on with the calling convetions.

ffi/custom_passes.cpp
apmasell: The way the LLVM constants are set up, I can only check for 0, 1, and -1. https://llvm.org/doxygen/classllvm_1_1Constant.html

ffi/custom_passes.cpp
apmasell: It is restricted to the cases where the PHI values contain constants. It would be a lot more effort to untangle if they were other results.

gmarkall:
CI was messed up because of the recent issue with it, so I'm just running the pipelines because I'm curious what the tests look like right now.

gmarkall:
@apmasell This is marked as RFR but it also contains a commit with the message "DO NOT MERGE" and has some CI fails - can you clarify its status please?

gmarkall:
> @apmasell This is marked as RFR but it also contains a commit with the message "DO NOT MERGE" and has some CI fails - can you clarify its status please?

I've now noticed that CI needs #903.

apmasell:
The DO NOT MERGE commit contains version changes that we need to have parallel LLVM 11 and LLVM 14 in CI. So, once review is done, I will remove that change before we merge. There's a similar one in Numba to make CI work.

ffi/custom_passes.cpp
```diff
@@ -920,27 +918,68 @@ struct RefPrunePass : public FunctionPass {
...
+                    phi_operand->getIncomingValueForBlock(bb));
+                return arg_value && arg_value->isOneValue();
+            }
+            // This path doesn't raise
             return false;
-        // get the value of the casted metadata and then return bool on whether
-        // it is the number one.
-        return data->getValue()->isOneValue();
+        }
+        return false;
```
sklam, line 982: I ran the entire Numba testsuite with an asserting that number of `fanout_raise` to be zero and the testsuite passed. That means nothing is ever trigger the `fanout_raise` case any more. I also haven't found any Numba jit code that will trigger this case.

sklam: I'll try to confirm this again tomorrow

sklam: ah, i just found `numba::typed::typedlist::_append[abi:v3][abi:c8tJTIcFHzwl2ILiXkcBV0KBSgP9CGZpAgA_3d](ListType_5bint32_5d, int)` is triggering `fanout_raise`. i might have an older commit of this PR installed previously.

llvmlite/tests/test_refprune.py
```diff
@@ -495,6 +496,28 @@ def test_fanout_raise_4(self):
...
+    call void @NRT_decref(i8* %ptr)
+    br label %common.ret
+bb_C:
+    br label %common.ret       ; pretend we throw an exception
+common.ret:
+    %common.ret.op = phi i32 [ 0, %bb_B ], [ 1, %bb_C ]
+    ret i32 %common.ret.op
+}
+
+!0 = !{i32 1}
```
sklam, line 514: unused metadata

ffi/custom_passes.cpp
sklam, line 982: Missing check to `!numba_exception_output` defined in https://github.com/numba/numba/pull/8535/files#diff-077fdcd3241710dcfffe590b74201d8156c1a21f19032e49165e873d5de4d7e7R400.

sklam (changes requested):

conda-recipes/llvmdev/meta.yaml
```diff
@@ -13,20 +14,16 @@ source:
...
-    # Fixes for aarch64 on LLVM 11 from https://reviews.llvm.org/D104123
-    - ../llvm_11_consecutive_registers.patch
-
+    - ../llvm14-remove-use-of-clonefile.patch
+    - ../llvm14-svml.patch
   - url: https://github.com/llvm/llvm-project/releases/download/llvmorg-{{ version }}/lld-{{ version }}.src.tar.xz
     fn: lld-{{ version }}.src.tar.xz
     sha256: {{ sha256_lld }}
-    folder: tools/lld
+
```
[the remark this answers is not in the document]
apmasell: `lld` requires to to build properly. There are a bunch of MachO structures that are in the unwind headers and the system unwind headers do not have the right ones.

conda-recipes/llvmlite/meta.yaml
```diff
@@ -23,19 +23,19 @@ requirements:
     # build.sh deals with it!
     - {{ compiler('c') }}    # [not (osx or armv6l or armv7l or win)]
     - {{ compiler('cxx') }}  # [not (osx or armv6l or armv7l or win)]
-    - vs2017_{{ target_platform  }}    # [win]
+    - vs2015_{{ target_platform  }}    # [win]
```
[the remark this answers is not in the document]
apmasell: I had problems with generating cmakes files on Windows. Things seem to have changed on the LLVM side. I can try going to 2017 again, but I think 2019 isn't supported yet.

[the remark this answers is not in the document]
h-vetinari: > but I think 2019 isn't supported yet.

It's the other way around. As of LLVM 14 (more precisely: https://reviews.llvm.org/D114639), VS2019 is the minimally required version

ffi/custom_passes.cpp
apmasell: Sorry, I pushed the wrong branch. This was the previous design that got abandoned. I've replaced the history with the correct branch.

conda-recipes/llvmlite/meta.yaml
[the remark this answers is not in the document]
apmasell: VS2019 doesn't seem to be available in conda yet. I'm pretty much at the "🤷  this builds" point with Windows.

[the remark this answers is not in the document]
h-vetinari: It is, but you get it through the regular `{{ compiler('cxx') }}`. Conda actually switched to VS2019 by default some time ago, but I haven't checked out how you hook into that infrastructure. Otherwise those are specified via `conda_build_config.yaml`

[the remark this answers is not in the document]
h-vetinari: FWIW, [here](https://conda-forge.org/docs/maintainer/knowledge_base.html#using-vs2019) are the instructions for using vs2019 within conda-forge (from before the switch to the new default, but describe the necessary todos).

llvmlite/tests/test_refprune.py
```diff
@@ -440,34 +440,37 @@ def test_fanout_raise_1(self):
...
+define i32 @main(i8* %ptr, i1 %cond, i8** %excinfo) {
 bb_A:
     call void @NRT_incref(i8* %ptr)
     br i1 %cond, label %bb_B, label %bb_C
 bb_B:
     call void @NRT_decref(i8* %ptr)
     ret i32 0
 bb_C:
-    ret i32 1, !ret_is_a_raise !0           ; bad metadata
+    store i8* null, i8** %excinfo, !numba_exception_output !0           ; bad metadata
```
sklam, line 451: The metadata does not look bad. Am I misreading it?

llvmlite/tests/test_refprune.py
```diff
@@ -478,23 +481,46 @@ def test_fanout_raise_3(self):
...
+define i32 @main(i8* %ptr, i1 %cond, i8** %excinfo) {
 bb_A:
     call void @NRT_incref(i8* %ptr)
     br i1 %cond, label %bb_B, label %bb_C
 bb_B:
-    ret i32 1, !ret_is_raise !0    ; BAD; all tails are raising without decref
+    ret i32 1    ; BAD; all tails are raising without decref
 bb_C:
-    ret i32 1, !ret_is_raise !0    ; BAD; all tails are raising without decref
+    ret i32 1    ; BAD; all tails are raising without decref
```
sklam, line 491: The original intention of the test is to verify that no pruning is done when all branches are raising. I'd suggest adding the store to excinfo for these branches so `isRaising` will return True.

conda-recipes/llvmdev/meta.yaml
```diff
@@ -1,8 +1,9 @@
-{% set shortversion = "11.1" %}
-{% set version = "11.1.0" %}
-{% set sha256_llvm = "ce8508e318a01a63d4e8b3090ab2ded3c598a50258cc49e2625b9120d4c03ea5" %}
-{% set sha256_lld = "017a788cbe1ecc4a949abf10755870519086d058a2e99f438829aef24f0c66ce" %}
-{% set build_number = "5" %}
+{% set shortversion = "14.0" %}
```
sklam, line 1: We'll need split build for linux-aarch64 to use llvm11 as discussed in https://github.com/numba/numba/issues/8756.

See 
https://github.com/numba/llvmlite/pull/599/files#diff-a97126465bc884942a676793bbd8f9443dc410f5aa05c57d9bbf1d0ce5a6c437R2-R18 for reference.

sklam (changes requested):
Changes needed:
- The build recipe needs to be changed to build llvm11 for linux-aarch64 per https://github.com/numba/numba/issues/8756
- Code changes in the tests are loosing the original intention. I've made suggestions to retain the original intention.

conda-recipes/llvmdev/build.sh
```diff
@@ -93,10 +100,3 @@ if [[ $ARCH == 'x86_64' ]]; then
    bin/opt -S -vector-library=SVML -mcpu=haswell -O3 $RECIPE_DIR/numba-3016.ll | bin/FileCheck $RECIPE_DIR/numba-3016.ll || exit $?
 fi
 
-# run the tests, skip some on linux-32
-cd ../test
-if [[ $ARCH == 'i686' ]]; then
-    ../build/bin/llvm-lit -vv Transforms Analysis CodeGen/X86
-else
-    ../build/bin/llvm-lit -vv Transforms ExecutionEngine Analysis CodeGen/X86
-fi
```
stuartarchibald, line 102: What is the reason for removing these tests? If it's because they are in `build.sh` perhaps they should be moved to `run_test.sh`?

llvmlite/tests/test_refprune.py
sklam: Now that the test changed. The metadata should not be removed in https://github.com/numba/llvmlite/pull/830/commits/9523fadecfc638a1b6abf0494ff45c2aea17757d

llvmlite/tests/test_refprune.py
sklam:
```suggestion
    store i8* null, i8** %excinfo, !numba_exeption_output !0           ; bad metadata
```

llvmlite/tests/test_refprune.py
```diff
@@ -440,34 +440,37 @@ def test_fanout_raise_1(self):
...
 }
 
 !0 = !{i1 true}
 """
 
     def test_fanout_raise_2(self):
         mod, stats = self.check(self.fanout_raise_2)
-        self.assertEqual(stats.fanout_raise, 0)
+        # Change in behaviour: ignore bad metadata
+        self.assertEqual(stats.fanout_raise, 2)
```
sklam, line 461: The original intention is to check that pruning does not occur when there's a bad metadata; e.g. misspelled metadata name. See suggestion above.

llvmlite/tests/test_refprune.py
```diff
@@ -478,23 +481,45 @@ def test_fanout_raise_3(self):
...
+    call void @NRT_decref(i8* %ptr)
+    br label %common.ret
+bb_C:
+    store i8* null, i8** %excinfo, !numba_exception_output !0
+    br label %common.ret
+common.ret:
+    %common.ret.op = phi i32 [ 0, %bb_B ], [ 1, %bb_C ]
+    ret i32 %common.ret.op
+}
+
```
sklam, line 516: restore missing metadata
```suggestion
!0 = !{i1 1}
```

sklam: https://github.com/numba/llvmlite/pull/830#discussion_r1128313228

llvmlite/tests/test_refprune.py
sklam: Also, this line is failing flake8 for being longer than 80cols.

llvmlite/tests/test_refprune.py
```diff
@@ -440,34 +440,37 @@ def test_fanout_raise_1(self):
...
 }
 
 !0 = !{i1 true}
 """
 
     def test_fanout_raise_2(self):
         mod, stats = self.check(self.fanout_raise_2)
-        self.assertEqual(stats.fanout_raise, 0)
+        # Change in behaviour: ignore bad metadata
+        self.assertEqual(stats.fanout_raise, 2)
```
sklam, line 461: CI is failing. The change should be reverted back to:

```suggestion
        self.assertEqual(stats.fanout_raise, 0)
```

sklam (approved):

conda-recipes/llvmdev/build.sh
sklam: We are having trouble getting any test to work. Deferring adding the test back to another PR given that this PR has passed buildfarm multiple times.

stuartarchibald: Please could a new issue be opened to track this so it's visible as to what is failing/has been tried already. I also think it is ok on this occasion to defer this work to another patch given how many runs through the build farm this has had. Many thanks.

CC @apmasell.

stuartarchibald:
The installation instructions need updating from here onward:
https://github.com/numba/llvmlite/blob/dffe582d6080494ba8e39689d09aacde1952214c/docs/source/admin-guide/install.rst?plain=1#L178
suggest deferring this to a subsequent PR?

ffi/Makefile.osx
```diff
@@ -1,6 +1,6 @@
 
-CXX = clang++ -std=c++11 -stdlib=libc++
-CXXFLAGS = $(LLVM_CXXFLAGS)
+CXX = clang++
+CXXFLAGS = $(LLVM_CXXFLAGS) -O3
```
sklam, line 3: @apmasell, is the `-O3` needed?

sklam: i am removing it in my PR

sklam (approved):
Approving for merge.
Installation docs defer to separate PR

PR #831, opened 2022-04-06
Format C++ code

apmasell:
Add formatting Azure pipeline and reformat all existing C++ code

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -38,6 +38,7 @@ jobs:
         PYTHON: '3.10'
         CONDA_ENV: cienv
         RUN_FLAKE8: yes
+        RUN_CLANG_FORMAT: yes
 # temporarily disabled
 #       pypy:
 #         PYTHON: pypy
@@ -68,13 +69,3 @@ jobs:
   parameters:
     name: Windows
     vmImage: windows-2019
-
-- job: check_formatting
-  pool:
-    vmImage: 'ubuntu-latest'
-  steps:
-  - script: |
-      set -e
-      sudo apt install -y clang-format
-      clang-format -n -Werror ffi/*.cpp ffi/*.h
-    displayName: Check C++ formatting
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -68,3 +68,13 @@ jobs:
   parameters:
     name: Windows
     vmImage: windows-2019
+
+- job: check_formatting
+  pool:
+    vmImage: 'ubuntu-latest'
+  steps:
+  - script: |
+      set -e
+      sudo apt install -y clang-format
+      clang-format -n -Werror ffi/*.cpp ffi/*.h
+    displayName: Check C++ formatting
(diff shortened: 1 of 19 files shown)


azure-pipelines.yml
```diff
@@ -68,3 +68,13 @@ jobs:
...
+
+- job: check_formatting
+  pool:
+    vmImage: 'ubuntu-latest'
+  steps:
+  - script: |
+      set -e
+      sudo apt install -y clang-format
+      clang-format -n -Werror ffi/*.cpp ffi/*.h
+    displayName: Check C++ formatting
```
stuartarchibald, line 80: Does `clang-format` produce the same formatting invariant of the `clang-format` version?

Also, I think it'd perhaps good to document that this is the formatting standard in the developer docs and issue some instructions about how to obtain and use it. I wonder if it would help to get the `clang-format` binary from the `clang-tools` conda package so as to provide a cross platform method of obtaining this tool too, it might make it easier in terms of development setup instructions?

Finally, is this adding a whole new build job to check formatting? Could this check be made as part of one of the existing build just by just adding an env var and the build script checking for that/adding it as a scripted step to an existing job? xref flake8 handling here: https://github.com/numba/numba/blob/61ec1fd0f69aeadece218dccf4c39ebc5c7dfbc4/buildscripts/azure/azure-linux-macos.yml#L30-L35

stuartarchibald (commented):
Thanks for the patch, great to see the C++ formatting standardised. Prior to checking the changes, I have some queries about `clang-format` that I've left inline. Thanks again!

azure-pipelines.yml
apmasell: Per private discussion, switching to the `clang-format-13` package from `conda-forge`.

docs/source/contributing.rst
```diff
@@ -72,7 +72,7 @@ Coding conventions
 ------------------
 
 * All Python code should follow `PEP 8 <https://www.python.org/dev/peps/pep-0008/>`_.
-* Our C++ code does not have a well-defined coding style.
+* Our C++ code is formatted with `clang-format-13` from the `conda-forge` channel
```
stuartarchibald, line 75:
```suggestion
* All C++ code is formatted using ``clang-format-13`` from the
  ``clang-format-13`` package available in the ``conda-forge`` conda channel.
```

buildscripts/azure/azure-linux-macos.yml
```diff
@@ -55,3 +55,11 @@ jobs:
         export PATH=$HOME/miniconda3/bin:$PATH
         buildscripts/incremental/test.sh
       displayName: 'Test'
+
+    - script: |
```
stuartarchibald, line 59: Perhaps add a comment here about why this is last (to stop the env under which testing takes place accidentally containing conda-forge packages)?

stuartarchibald (commented):
Thanks for the patch, great to see the formatting standardised. There are a couple of minor comments inline to look at, the only other comment I have is that I find the 2 column indent makes the code less readable than when there's 4 (maybe due to looking at a lot of Python!), if others find the same, could it perhaps be changed?

sklam (approved):
Looks good. I've manually checked by running clang-format locally and it recreates the same changes.

PR #835, opened 2022-04-27
Add pre-commit hooks for clang-format

apmasell:
Include a clang-format pre-commit hook to ensure the code is formatted
automatically.


sklam (approved):
Tested manually. Works as expected.

PR #841, opened 2022-05-19
Merge pull request #813 from seibert/m1_support

esc:
Add m1 support to conda build scripts

Backport of: https://github.com/numba/llvmlite/pull/813

### DIFF
--- a/conda-recipes/llvmdev/build.sh
+++ b/conda-recipes/llvmdev/build.sh
@@ -9,7 +9,11 @@ set -x
 LLVM_TARGETS_TO_BUILD=${LLVM_TARGETS_TO_BUILD:-"host;AArch64;AMDGPU;ARM;BPF;Hexagon;Mips;MSP430;NVPTX;PowerPC;Sparc;SystemZ;X86;XCore;RISCV"}
 
 # This is the clang compiler prefix
-DARWIN_TARGET=x86_64-apple-darwin13.4.0
+if [[ $build_platform == osx-arm64 ]]; then
+    DARWIN_TARGET=arm64-apple-darwin20.0.0
+else
+    DARWIN_TARGET=x86_64-apple-darwin13.4.0
+fi
 
 
 declare -a _cmake_config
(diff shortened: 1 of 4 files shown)


esc:
Screen Shot 2022-05-19 at 11 33 18

Looks like the Public CI failures are potentially transient.

conda-recipes/llvmdev/build.sh
```diff
@@ -9,7 +9,11 @@ set -x
 LLVM_TARGETS_TO_BUILD=${LLVM_TARGETS_TO_BUILD:-"host;AArch64;AMDGPU;ARM;BPF;Hexagon;Mips;MSP430;NVPTX;PowerPC;Sparc;SystemZ;X86;XCore;RISCV"}
 
 # This is the clang compiler prefix
-DARWIN_TARGET=x86_64-apple-darwin13.4.0
+if [[ $build_platform == osx-arm64 ]]; then
```
gmarkall, line 12: I don't know whether it matters in our setup, but it seems like this has the potential to do the wrong thing when cross-compiling on an M1 system for another system - it seems like the target is being set based on the build platform and not the target platform.

Note: the above is what I can tell from the PR that added this variable to conda-build: https://github.com/conda/conda-build/pull/4005. The docs for the variable at https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#env-vars say "the native subdir of the conda executable", which I can't make sense of.

gmarkall: I didn't review #813 that added this, but I wonder if this is potentially an issue on `main` too.

conda-recipes/llvmdev/meta.yaml
```diff
@@ -54,7 +54,7 @@ requirements:
   host:
     # needed for llc at runtime
     - zlib # [not win]
-    - xar # [osx]
+    - xar # [osx and x86_64]
```
gmarkall, line 57: xref to previous discussion establishing that this is not needed: https://github.com/numba/llvmlite/pull/813/files#r826859213

esc: Was this missed by the original PR and do we need an additional PR for `main` to make this change?

gmarkall (approved):
I think this is OK to merge, especially given that #813 is already in `main` and no issues have been observed with nightly builds, which I understand we're doing.

Some notes on what I did observe as noted on the diff:

- I wonder if cross-compiling won't behave correctly from an M1 because we select the target based on the build platform, but as we're not cross-compiling from M1 to x86 I don't think this will be an issue for the 0.55.2 release
- `xar` is not needed.

stuartarchibald (approved):
Thanks for doing the back port. I cherry picked the merge of https://github.com/numba/llvmlite/pull/813 onto release0.38 locally and the diff against this PR branch was empty so think it's good.

PR #845, opened 2022-05-19
Backport #786 for 0.38.1

esc:
Update Windows and OSX CI images.


stuartarchibald (approved):
Thanks for the patch. Visual inspection of two lines confirms this is the same as #786.

PR #851, opened 2022-06-01
adding the llvm_11_consecutive_registers.patch

### LINKED ISSUE #839: Building from source: should llvm_11_consecutive_registers.patch be used?
I am building llvmlite from source, and there seems to be a discrepancy/bug in the docs.  The docs only mention five patches to apply, while the `conda-recipes` directory has six patches in it, the unmnentioned patch being [`llvm_11_consecutive_registers.patch`](https://github.com/numba/llvmlite/blob/v0.38.0/conda-recipes/llvm_11_consecutive_registers.patch).

Should I apply this patch? I know it is only relevant to Aarch64, but that is the architecture I am building for.

esc:
Fixes #839


stuartarchibald (approved):
Thanks for the patch.

PR #857, opened 2022-06-22
Delegate passmanager remarks methods

apmasell:
Use `self.run` rather than the C binding for the `run_with_remarks` methods in
the passmanagers.


sklam (approved):
LGTM

PR #860, opened 2022-06-28
the git tag for the RC needs an rc1 suffix

esc:
As title


stuartarchibald (approved):
Thanks for the patch.

PR #871, opened 2022-08-22
Refactor native library loading

folded:
This simplifies the process of supporting multiple paths for native
code loading and allows cleanup to occur on non-windows platforms when
the libllvmlite shared library is extracted to a temporary directory.

### DIFF
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -77,19 +77,53 @@ class _LLVMLock:
         self._lock.release()
 
 
+class _suppress_cleanup_errors:
+    def __init__(self, context):
+        self._context = context
+
+    def __enter__(self):
+        return self._context.__enter__()
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        try:
+            return self._context.__exit__(exc_type, exc_value, traceback)
+        except PermissionError:
+            pass  # Resource dylibs can't be deleted on Windows.
+
+
 class _lib_wrapper(object):
     """Wrap libllvmlite with a lock such that only one thread may access it at
     a time.
 
     This class duck-types a CDLL.
     """
-    __slots__ = ['_lib', '_fntab', '_lock']
+    __slots__ = ['_lib_handle', '_fntab', '_lock']
 
-    def __init__(self, lib):
-        self._lib = lib
+    def __init__(self):
+        self._lib_handle = None
         self._fntab = {}
         self._lock = _LLVMLock()
 
+    def _load_lib(self):
+        try:
+            with _suppress_cleanup_errors(importlib.resources.path(
+                __name__.rpartition(".")[0], get_library_name())) as lib_path:
+                self._lib_handle = ctypes.CDLL(str(lib_path))
+                # Check that we can look up expected symbols.
+                _ = self._lib_handle.LLVMPY_GetVersionInfo()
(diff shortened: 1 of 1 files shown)


gmarkall:

llvmlite/binding/ffi.py
```diff
@@ -77,19 +77,55 @@ def __exit__(self, *exc_details):
...
-        self._lib = lib
+    def __init__(self):
+        self._lib_handle = None
         self._fntab = {}
         self._lock = _LLVMLock()
 
+    def _load_lib(self):
+        try:
+            with _suppress_cleanup_errors(importlib.resources.path(
+                    __name__.rpartition(".")[0],
```
sklam, line 110: FYI, #906 will cause conflict here.

sklam:
I've merged main and fixed the conflict in https://github.com/numba/llvmlite/pull/871#discussion_r1107347452

llvmlite/binding/ffi.py
```diff
@@ -80,19 +80,55 @@ def __exit__(self, *exc_details):
...
+        except (OSError, AttributeError) as e:
+            # OSError may be raised if the file cannot be opened, or is not
+            # a shared library.
+            # AttributeError is raised if LLVMPY_GetVersionInfo does not
+            # exist.
+            raise OSError("Could not find/load shared object file") from e
+
+    @property
+    def _lib(self):
+        # Not threadsafe.
```
sklam, line 127: Even though this is not threadsafe, in practice the function will be invoked during module import time because of the global references to `ffi.lib` in many of the modules under `llvmlite.binding.*`.

llvmlite/binding/ffi.py
```diff
@@ -80,19 +80,55 @@ def __exit__(self, *exc_details):
...
-    def __init__(self, lib):
-        self._lib = lib
+    def __init__(self):
+        self._lib_handle = None
         self._fntab = {}
         self._lock = _LLVMLock()
 
+    def _load_lib(self):
+        try:
+            with _suppress_cleanup_errors(_importlib_resources_path(
```
sklam, line 112: I changed this line to use `_importlib_resources_path` instead of directly using `importlib.resources` to deal with the use of deprecated API.

llvmlite/binding/ffi.py
```diff
@@ -80,19 +80,55 @@ def __exit__(self, *exc_details):
...
+    __slots__ = ['_lib_handle', '_fntab', '_lock']
 
-    def __init__(self, lib):
-        self._lib = lib
+    def __init__(self):
+        self._lib_handle = None
         self._fntab = {}
         self._lock = _LLVMLock()
 
+    def _load_lib(self):
```
sklam, line 110: I believe the introduction of this method is to allow easier patching to deal with the usecases the OP described. Originally the logic here is done directly on the bare module scope outside of any function.

sklam (approved):
My review up to https://github.com/numba/llvmlite/pull/871/commits/6bdc95ddb8d4117c935093a086ea1e80cfc67d76 which is just before I merged main to this branch.

The changes look good. The changes do not poses any observable effect to most of our user. It is done mainly to allow repacking of llvmlite for use-cases outside of the official support. If and when these use-cases become more common, we can consider adding a hook for alternative distribution of libllvmlite's content.

esc:
@sklam thank you for the review. This will be merged now.

PR #872, opened 2022-08-22
Fix symbol resolution race

folded:


gmarkall:

PR #873, opened 2022-08-22
Make symbol resolution lazy

folded:
This delays the lookup of a symbol until it is first called, which
in turn allows us to write a test for extraction of native code from
a package.


gmarkall:

PR #876, opened 2022-08-25
Remove `llvmlite.llvmpy` after deprecation

apmasell:
This module was deprecated and scheduled for deletion in 0.40.


bench.py
```diff
@@ -100,26 +100,21 @@ def run_bench(verbose):
 
 if __name__ == "__main__":
     if len(sys.argv) != 2:
-        print("Usage: %s (llvmlite|llvmpy)"
+        print("Usage: %s llvmlite"
```
sklam, line 103: This file was for comparing llvmpy (https://github.com/llvmpy/llvmpy) with llvmlite. I think we can remove this file all together now.

sklam:
oops, i accidentally closed the PR

bench.py
apmasell: Done.

sklam (approved):

PR #883, opened 2022-10-18
Adds support for calling functions with 'tail', 'notail', or 'musttail' markers.

bslatkin:
More background available in LLVM IR manual: https://llvm.org/docs/LangRef.html#call-instruction

API addition is backwards compatible with previous usage, so this doesn't break anyone.

### DIFF
--- a/docs/source/user-guide/ir/ir-builder.rst
+++ b/docs/source/user-guide/ir/ir-builder.rst
@@ -516,7 +516,7 @@ Memory
 Function call
 ---------------
 
-.. method:: IRBuilder.call(fn, args, name='', cconv=None, tail=False, \
+.. method:: IRBuilder.call(fn, args, name='', cconv=None, tail=None, \
    fastmath=(), attrs=(), arg_attrs=None)
 
    Call function *fn* with arguments *args*, a sequence of values.
--- a/docs/source/user-guide/ir/ir-builder.rst
+++ b/docs/source/user-guide/ir/ir-builder.rst
@@ -522,8 +522,8 @@ Function call
    Call function *fn* with arguments *args*, a sequence of values.
 
    * *cconv* is the optional calling convention.
-   * *tail*, if ``"tail"``, is a hint for the optimizer to perform
-     tail-call optimization. Use ``"musttail"`` to indicate the specified the call must be tail call optimized in order for the program to be correct. Use ``"notail"`` to indicate that the call should never be tail call optimized. Alternatively, may provide ``False`` to indicate no specific tail call optimization behavior, or ``True`` to request `"tail"`` hint behavior (this is backwards compatible with the previous API).
+   * *tail*, if ``"tail"`` it's a hint for the optimizer to perform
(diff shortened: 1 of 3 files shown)


gmarkall:
Many thanks for the PR - it's on my to-do list to review.

bslatkin:
Thank you for looking! Also I wanted to mention that I've compiled actual applications using the "musttail" flag (via this llvmlite PR). I can confirm that it causes stack frames to be reused for tail and sibling calls, as expected.

docs/source/user-guide/ir/ir-builder.rst
```diff
@@ -522,8 +522,8 @@ Function call
    Call function *fn* with arguments *args*, a sequence of values.
 
    * *cconv* is the optional calling convention.
-   * *tail*, if ``True``, is a hint for the optimizer to perform
-     tail-call optimization.
+   * *tail*, if ``"tail"`` it's a hint for the optimizer to perform
+     tail-call optimization. Use ``"musttail"`` to indicate the specified call must be tail-call optimized in order for the program to be correct. Use ``"notail"`` to indicate that the call should never be tail-call optimized. Alternatively, may provide ``False`` to indicate no specific tail-call optimization behavior, or ``True`` to request ``"tail"`` hint behavior (support for a boolean value is for backwards compatibility with the previous API).
```
gmarkall, line 526:
```suggestion
   * *tail* controls tail-call optimization behavior. It may be one of:

     * ``None`` (the default): indicates no specific tail-call optimization
       behavior.
     * ``"tail"``: a hint that indicates that the call should be tail-call
       optimized, but may be ignored.
     * ``"musttail"``: indicates that the call must be tail-call optimized for
       program correctness.
     * ``"notail"``: indicate thats the call should never be tail-call
       optimized.

     For backwards compatibility with previous versions, the following values
     are also accepted:

     * ``False`` is equivalent to ``None``, indicating no specific behavior.
     * ``True`` is equivalent to ``"tail"``, suggesting tail-call optimization.
```

gmarkall (changes requested):
Many thanks for the PR! I think the code changes and test additions look good in general. A couple of requests:

- Now that the tail options as `tail`, `notail`, and `musttail`, it feels more intuitive for the default value to be `None` rather than `False`, because it feels like `False` is closer in meaning to `notail` than not providing a hint at all. Could you make the default `None` instead, with the effect of `tail=None` being to provide no specific hint please? (I don't *think* this needs a code change other than changing the default value of the kwarg from `False` to `None`).
- The documentation changes don't wrap consistently with the rest of the file, and read a little confusingly. I've made a suggestion on the diff that I feel improves the way they read, based on the suggested change to make `None` the default value. What do you think of these changes? Would you be happy to accept them?

Other than the above two points I think this is looking good, so once you've responded I'll take another look and hopefully we can wrap this up soon.

gmarkall:
Thanks for fixing the method signature in the docs, I missed that!

bslatkin:
All my changes are in and the checks have passed. Thank you for the review and edits!

gmarkall (approved):
Thanks for the changes - this looks good!

PR #869, opened 2022-08-02
bump max Python version to 3.11

esc:
As title

PR #884, opened 2022-10-25
add python 3.11 support

spidertyler2005:
This likely needs testing, but python 3.11 just released and it doesn't appear to change much. llvmlite for 3.10 should likely be compatible for 3.11

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ except ImportError:
 
 
 min_python_version = "3.7"
-max_python_version = "3.11"  # exclusive
+max_python_version = "3.12"  # exclusive
 
 
 def _guard_py_ver():


esc:
@spidertyler2005 thank you for opening this. When contributing to open source projects, it is generally a good to check the issue tracker for duplicates, like this one:

https://github.com/numba/llvmlite/pull/869

spidertyler2005:
> @spidertyler2005 thank you for opening this. When contributing to open source projects, it is generally a good idea to check the issue tracker for duplicates, like this one:
> 
> #869

oh, my bad. Thanks for explaining

esc:
@spidertyler2005 we have now also pinned the 3.11 issue to the top of the issue tracker to make it easier for folks to find. Hope that helps.

PR #886, opened 2022-10-26
Simplify setup.py Python version guard

mbargull:
Over at the conda-forge feedstock we [noticed](https://github.com/conda-forge/llvmlite-feedstock/pull/66#issuecomment-1292357510) that the build ignored the version guard since we didn't have `setuptools`/`packaging` installed.
Since we're only comparing "simple" versions and `sys.version_info` is a tuple already, we can simplify the logic by omitting version string parsing overall and just compare the tuples.

cc @jakirkham

### DIFF
--- a/llvmlite/tests/test_setup.py
+++ b/llvmlite/tests/test_setup.py
@@ -1,0 +1,71 @@
+"""
+Tests for setup.py behavior
+"""
+
+import distutils.core
+try:
+    import setuptools
+except ImportError:
+    setuptools = None
+import sys
+import unittest
+from collections import namedtuple
+from importlib.util import spec_from_file_location, module_from_spec
+from pathlib import Path
+
+from . import TestCase
+
+setup_path = Path(__file__).parent.parent.parent / "setup.py"
+
+
+class TestSetup(TestCase):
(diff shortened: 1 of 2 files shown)


jakirkham:
Thanks Marcel! 🙏

cc @stuartarchibald

esc:
@mbargull @jakirkham thank you! I've added this to my queue!

esc:
I glanced over the code, it seems fine and straightforward. What are everyones ideas on how to test this?

mbargull:
Added a test case and function `_version_info_str` (forgot to turn tuples into dotted version strings in the error message before).

esc (approved):
I reviewed the code, ran this through the wheel build of the Anaconda build farm and checked this locally with a Python 3.6 environment:

```
  File "setup.py", line 45, in <module>
    _guard_py_ver()
  File "setup.py", line 42, in _guard_py_ver
    raise RuntimeError(msg.format(cur_py, min_py, max_py))
RuntimeError: Cannot install on Python version 3.6.13; only versions >=3.7,<3.11 are supported.
```

LGTM

PR #895, opened 2022-12-06
Add support for self referential metadata

aseyboldt:
Metadata definitions that refer to themselves like `!0 = { !0 }` are allowed by llvm, and used to make some metadata entries globally unique (see for instance https://llvm.org/docs/LangRef.html#noalias-and-alias-scope-metadata). It seems this isn't supported by llvmlite currently.

In this PR I add support for this by adding a new `self_ref` argument to `Model.add_metadata`, that will always return a new metadata entry, but prepending self as the first operand. So this only supports references to self as the first entry, I haven't seen any usage in llvm at other locations however.

*Update* I probably should also mention that I'm still pretty new to llvm, and this proposed API might not actually be the best as a result...

### DIFF
--- a/llvmlite/ir/module.py
+++ b/llvmlite/ir/module.py
@@ -42,22 +42,32 @@ class Module(object):
             fixed_ops.append((name, op))
         return fixed_ops
 
-    def add_metadata(self, operands):
+    def add_metadata(self, operands, *, self_ref=False):
         """
         Add an unnamed metadata to the module with the given *operands*
         (a sequence of values) or return a previous equivalent metadata.
         A MDValue instance is returned, it can then be associated to
         e.g. an instruction.
+
+        A self referential metadata entry has itself as the first
+        operand to ensure uniquencess. These references are never
+        cached.
         """
         if not isinstance(operands, (list, tuple)):
             raise TypeError("expected a list or tuple of metadata values, "
                             "got %r" % (operands,))
         operands = self._fix_metadata_operands(operands)
         key = tuple(operands)
-        if key not in self._metadatacache:
+        if self_ref or key not in self._metadatacache:
             n = len(self.metadata)
-            md = values.MDValue(self, operands, name=str(n))
-            self._metadatacache[key] = md
+            md = values.MDValue(
+                self,
+                operands,
+                name=str(n),
+                self_ref=self_ref,
+            )
+            if not self_ref:
+                self._metadatacache[key] = md
         else:
             md = self._metadatacache[key]
         return md
--- a/llvmlite/ir/module.py
+++ b/llvmlite/ir/module.py
@@ -50,8 +50,11 @@ class Module(object):
         e.g. an instruction.
 
         A self referential metadata entry has itself as the first
-        operand to ensure uniquencess. These references are never
-        cached.
+        operand to ensure uniqueness. These references are never
+        cached. If `self_ref` is set, create a self referential
+        metadata entry by first creating a new metadata value
+        and then add itself as first argument. `operands` are then
+        added as additional operands.
         """
         if not isinstance(operands, (list, tuple)):
             raise TypeError("expected a list or tuple of metadata values, "
--- a/llvmlite/tests/test_ir.py
+++ b/llvmlite/tests/test_ir.py
@@ -323,6 +323,17 @@ class TestIR(TestBase):
         self.assert_ir_line('!1 = !{ i32 789 }', mod)
         self.assert_ir_line('!2 = !{ i32 123, !0, !1, !0 }', mod)
 
+    def test_self_relf_metadata(self):
+        mod = self.module()
+        value = mod.add_metadata((), self_ref=True)
+        self.assert_ir_line('!0 = !{ !0 }', mod)
+        value2 = mod.add_metadata((value,))
+        assert value is not value2
+        self.assert_ir_line('!1 = !{ !0 }', mod)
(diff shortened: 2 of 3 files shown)


esc:
@aseyboldt thank you for submitting this PR. I have labelled it as Ready for Review for now. It will hopefully be assigned a reviewer and milestone during the next issue triage meeting on Tuesdays.

twiecki:
@esc Just checking in if a reviewer has been assigned?

llvmlite/ir/module.py
```diff
@@ -42,22 +42,32 @@ def _fix_di_operands(self, operands):
...
+    def add_metadata(self, operands, *, self_ref=False):
         """
         Add an unnamed metadata to the module with the given *operands*
         (a sequence of values) or return a previous equivalent metadata.
         A MDValue instance is returned, it can then be associated to
         e.g. an instruction.
+
+        A self referential metadata entry has itself as the first
+        operand to ensure uniquencess. These references are never
+        cached.
```
sklam, line 54: Can you add few words to relate this to the `self_ref` argument? e.g. the `self_ref` creates a self referential metadata by prepending itself to the operands?.

llvmlite/ir/module.py
```diff
@@ -42,22 +42,32 @@ def _fix_di_operands(self, operands):
...
-    def add_metadata(self, operands):
+    def add_metadata(self, operands, *, self_ref=False):
         """
         Add an unnamed metadata to the module with the given *operands*
         (a sequence of values) or return a previous equivalent metadata.
         A MDValue instance is returned, it can then be associated to
         e.g. an instruction.
+
+        A self referential metadata entry has itself as the first
+        operand to ensure uniquencess. These references are never
```
sklam, line 53:
```suggestion
        operand to ensure uniqueness. These references are never
```

llvmlite/tests/test_ir.py
```diff
@@ -323,6 +323,17 @@ def test_unnamed_metadata_3(self):
...
+    def test_self_relf_metadata(self):
+        mod = self.module()
+        value = mod.add_metadata((), self_ref=True)
+        self.assert_ir_line('!0 = !{ !0 }', mod)
+        value2 = mod.add_metadata((value,))
+        assert value is not value2
+        self.assert_ir_line('!1 = !{ !0 }', mod)
+        value3 = mod.add_metadata((), self_ref=True)
+        self.assert_ir_line('!2 = !{ !2 }', mod)
+        assert value is not value3
```
sklam, line 335: Can you add test for metadata that has other operands?

sklam (commented):
The situation with self-referential metadata is a bit unclear in LLVM. There are discussion about removing them https://lists.llvm.org/pipermail/llvm-dev/2021-February/148674.html. I'm not sure if the `distinct` keyword has make this self-referential trick outdated. For now, LLVM still contains code to create self-referential MD. It is done so by first building a metadata node without self-reference then replace the first operand with the self-reference; e.g. https://github.com/llvm/llvm-project/blob/080a07199e8c146a58ef3506cfcbf554eced21fc/llvm/lib/Analysis/LoopInfo.cpp#L1159-L1161. 

The patch here adds a `self_ref` keyword argument to prepend a self-reference to the operands. I think this logic is okay given any self-referential MD node seems to always be just a bag of stuff. I don't think we should overengineer it too much.

Given the code changes here are small, I am ok to accept this patch once the following suggestions are addressed.

llvmlite/ir/module.py
aseyboldt: I extended the description.

llvmlite/ir/module.py
aseyboldt: fixed

llvmlite/tests/test_ir.py
aseyboldt: done

aseyboldt:
@sklam Thanks, I added a test and extended the docs a little as you suggested.
I think the test failure is unrelated?
```
Warning, treated as error:
failed to reach any of the inventories with the following issues:
intersphinx inventory 'https://docs.python.org/3/objects.inv' not fetchable due to <class 'requests.exceptions.ConnectionError'>: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
make: *** [Makefile:55: html] Error 2
```

ricardoV94:
@aseyboldt any interest in reviving this PR?
@esc @sklam still interest in getting this merged?

esc:
> @aseyboldt any interest in reviving this PR? @esc @sklam still interest in getting this merged?

I would volunteer to review it, yes.

ricardoV94:
Suggestion for docs tweak:

Replacing the `add_metadata` entry (around line 51) with:

```rst
   * .. method:: add_metadata(operands, self_ref=False)

        Add an unnamed :ref:`metadata` node to the module with
        the given *operands*---a list of metadata-compatible
        values. If another metadata node with equal operands
        already exists in the module, it is reused instead.
        Returns an :class:`MDValue`.

        If *self_ref* is true, the node is given itself as its
        first operand, followed by *operands*. This is how LLVM
        gives a metadata node a globally unique identity, as
        used by ``!alias.scope`` and ``!noalias`` domains and
        scopes. Self-referential nodes are never reused, so
        every call returns a distinct node.

        EXAMPLE::

           domain = module.add_metadata([], self_ref=True)
           scope = module.add_metadata([domain], self_ref=True)
```

PR #896, opened 2022-12-11
drop upper limit on Python for conda recipe

esc:
`conda-build` will override this setting anyway, when building for Python 3.10 using `--python` like we do for llvmlite. It is effectively a no-op under these circumstances.

### DIFF
--- a/conda-recipes/llvmlite/meta.yaml
+++ b/conda-recipes/llvmlite/meta.yaml
@@ -37,7 +37,7 @@ requirements:
     # llvmdev is built with libz compression support
     - zlib           # [unix and not (armv6l or armv7l)]
   run:
-    - python >=3.8,<3.10
+    - python >=3.8
     - vs2015_runtime # [win]
     # osx has dynamically linked libstdc++
     - libcxx >=4.0.1 # [osx]


esc:
Following up from https://github.com/numba/llvmlite/pull/892#pullrequestreview-1204681843

esc:
Looking at the Anaconda based feedstock/recipe, I see that a different trick is being used to constrain the Python version:

https://github.com/AnacondaRecipes/llvmlite-feedstock/blob/9d4f646641c2e71d750dd01e55ac9b97f7246053/recipe/meta.yaml#L16

esc:
> @esc In the triage meeting I understood this is superseded / should be closed - can this be closed?

I am not sure, it's a no-op, maybe for cleanliness purposes it should be removed anyway? I think we agreed to no longer clamp the upper limit for Python for llvmlite, since it is mostly compatible with new Python minor versions. That would also indicate that this should be merged.

esc:
Sorry, wrong button.

esc:
@gmarkall do we want to push this over the line?

apmasell (approved):

PR #899, opened 2023-01-16
Run Minconda install with bash

gmarkall:
The Minconda installer seems to expect that `[[` can be used in all sh implementations, which is not necessarily the case (there are some discussions around this issue e.g. in https://github.com/conda/conda/issues/10431).

To attempt to work around this, we run the installer with bash instead.


stuartarchibald (approved):
Thanks for the patch!

PR #902, opened 2023-02-03
Testhound/add archive

testhound:
This pull request adds support for importing symbols from a static library into the execution engine.

### DIFF
--- a/docs/source/user-guide/binding/execution-engine.rst
+++ b/docs/source/user-guide/binding/execution-engine.rst
@@ -79,6 +79,13 @@ The ExecutionEngine class
             or a object file instance. Object file instance is not usable after this
             call.
 
+   * .. method:: add_archive(archive_file)
+
+        Add the symbols from the specified static archive file to the execution
+        engine. It is a fatal error in LLVM if the *archive_file* does not exist.
+
+        * *archive* str: a path to the static object file
+
    * .. method:: set_object_cache(notify_func=None, getbuffer_func=None)
 
         Set the object cache callbacks for this engine.
--- a/ffi/executionengine.cpp
+++ b/ffi/executionengine.cpp
@@ -10,7 +10,9 @@
 #include "llvm/Object/Binary.h"
 #include "llvm/Object/ObjectFile.h"
 #include "llvm/Support/Memory.h"
-
+#include "llvm/Support/MemoryBuffer.h"
+#include "llvm/Object/Archive.h"
+#include "llvm/Support/Errc.h"
 #include <cstdio>
 #include <memory>
 
@@ -165,6 +167,40 @@ LLVMPY_MCJITAddObjectFile(LLVMExecutionEngineRef EE, LLVMObjectFileRef ObjF) {
 
     engine->addObjectFile(
         {std::move(binary_tuple.first), std::move(binary_tuple.second)});
+}
+
+API_EXPORT(int)
+LLVMPY_MCJITAddArchive(LLVMExecutionEngineRef EE, const char *ArchiveName,
+                       const char **OutError) {
+    using namespace llvm;
+    using namespace llvm::object;
+    auto engine = unwrap(EE);
+
+    ErrorOr<std::unique_ptr<MemoryBuffer>> ArBufOrErr = MemoryBuffer::getFile(ArchiveName);
+
+    std::error_code EC = ArBufOrErr.getError();
+    if (EC){
+        *OutError = LLVMPY_CreateString(EC.message().c_str());
+        return 1;
+    }
+
+    Expected<std::unique_ptr<object::Archive>> ArchiveOrError =
+        object::Archive::create(ArBufOrErr.get()->getMemBufferRef());
+
+    if (!ArchiveOrError) {
+        auto takeErr = ArchiveOrError.takeError();
+        std::string archiveErrStr = "Unable to load archive: " + std::string(ArchiveName);
+        *OutError = LLVMPY_CreateString(archiveErrStr.c_str());
+        return 1;
+    }
+
+    std::unique_ptr<MemoryBuffer> &ArBuf = ArBufOrErr.get();
+    std::unique_ptr<object::Archive> &Ar = ArchiveOrError.get();
+    object::OwningBinary<object::Archive> owningBinaryArchive(std::move(Ar),
+                                                              std::move(ArBuf));
+    engine->addArchive(std::move(owningBinaryArchive));
+    *OutError = LLVMPY_CreateString("");
+    return 0;
 }
 
 //
--- a/llvmlite/tests/test_binding.py
+++ b/llvmlite/tests/test_binding.py
@@ -6,6 +6,7 @@ import locale
 import os
 import platform
 import re
+import shutil
 import subprocess
 import sys
 import unittest
@@ -16,6 +17,8 @@ from llvmlite import ir
 from llvmlite import binding as llvm
 from llvmlite.binding import ffi
 from llvmlite.tests import TestCase
+import llvmlite.tests
+from distutils.ccompiler import new_compiler
 
 
 # arvm7l needs extra ABI symbols to link successfully
@@ -1934,6 +1937,54 @@ class TestObjectFile(BaseTest):
                 self.assertEqual(len(s.data()), 31)
                 self.assertEqual(s.data().hex(), issue_632_text)
 
+
+
+class TestArchiveFile(BaseTest):
+    mod_archive_asm = """
+        ;ModuleID = <string>
+        target triple = "{triple}"
+
+        declare i32 @__multiply_accumulate(i32 %0, i32 %1, i32 %2)
+        declare i32 @__multiply_subtract(i32 %0, i32 %1, i32 %2)
+    """
+
+    @unittest.skipUnless(sys.platform.startswith('linux'),
+                         "Linux-specific test")
+    def test_add_archive(self):
+        target_machine = self.target_machine(jit=False)
+
+        jit = llvm.create_mcjit_compiler(self.module(self.mod_archive_asm),
+                                         target_machine)
+
+        # Create compiler with default options
+        c = new_compiler()
+        workdir = os.path.dirname(llvmlite.tests.__file__) + "/"
+
+        # Compile into .o files
+        file1 = workdir + "a.c"
+        file2 = workdir + "b.c"
+
+        objects = c.compile([file1, file2])
+        library_name = "foo"
+        # Create static or shared library
+
+        c.create_static_lib(objects, library_name, output_dir=workdir)
+
+        static_library_name = workdir + "lib" + library_name + ".a"
+        jit.add_archive(static_library_name)
+
+        mac_func = CFUNCTYPE(c_int, c_int, c_int, c_int)(
+        jit.get_function_address("__multiply_accumulate"))
+
+        msub_func = CFUNCTYPE(c_int, c_int, c_int, c_int)(
+        jit.get_function_address("__multiply_subtract"))
+
+        self.assertEqual(mac_func(10, 10, 20), 120)
+        self.assertEqual(msub_func(10, 10, 20), 80)
+
+        os.unlink(objects[0])
+        os.unlink(objects[1])
+        os.unlink(static_library_name)
 
 class TestTimePasses(BaseTest):
     def test_reporting(self):
--- a/llvmlite/tests/test_binding.py
+++ b/llvmlite/tests/test_binding.py
@@ -6,7 +6,6 @@ import locale
 import os
 import platform
 import re
-import shutil
 import subprocess
 import sys
 import unittest
@@ -1938,7 +1937,6 @@ class TestObjectFile(BaseTest):
                 self.assertEqual(s.data().hex(), issue_632_text)
 
 
-
 class TestArchiveFile(BaseTest):
     mod_archive_asm = """
         ;ModuleID = <string>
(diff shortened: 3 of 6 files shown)


llvmlite/tests/test_binding.py
```diff
@@ -1935,6 +1937,54 @@ def test_get_section_content(self):
...
+                         "Linux-specific test")
+    def test_add_archive(self):
+        target_machine = self.target_machine(jit=False)
+
+        jit = llvm.create_mcjit_compiler(self.module(self.mod_archive_asm),
+                                         target_machine)
+
+        # Create compiler with default options
+        c = new_compiler()
+        workdir = os.path.dirname(llvmlite.tests.__file__) + "/"
```
gmarkall, line 1959: It can't be assumed that the llvmlite test dir is writable (and I don't think it's good form to drop files into a package dir as part of running the tests anyway). If the test dir isn't writable then the test errors:

```
ERROR: test_add_archive (llvmlite.tests.test_binding.TestArchiveFile)
  File "/home/gmarkall/numbadev/llvmlite/llvmlite/tests/test_binding.py", line 1969, in test_add_archive
    c.create_static_lib(objects, library_name, output_dir=workdir)
  File "/home/gmarkall/mambaforge/envs/numbadev/lib/python3.10/site-packages/setuptools/_distutils/unixccompiler.py", line 131, in create_static_lib
    self.spawn(self.archiver +
  File "/home/gmarkall/mambaforge/envs/numbadev/lib/python3.10/site-packages/setuptools/_distutils/ccompiler.py", line 917, in spawn
    spawn(cmd, dry_run=self.dry_run, **kwargs)
  File "/home/gmarkall/mambaforge/envs/numbadev/lib/python3.10/site-packages/setuptools/_distutils/spawn.py", line 68, in spawn
    raise DistutilsExecError(
distutils.errors.DistutilsExecError: command '/opt/binutils/2.37/bin/ar' failed with exit code 1

```

llvmlite/tests/test_binding.py
```diff
@@ -1935,6 +1937,54 @@ def test_get_section_content(self):
...
+        mac_func = CFUNCTYPE(c_int, c_int, c_int, c_int)(
+            jit.get_function_address("__multiply_accumulate"))
+
+        msub_func = CFUNCTYPE(c_int, c_int, c_int, c_int)(
+            jit.get_function_address("__multiply_subtract"))
+
+        self.assertEqual(mac_func(10, 10, 20), 120)
+        self.assertEqual(msub_func(10, 10, 20), 80)
+
+        os.unlink(objects[0])
```
gmarkall, line 1983: If the test fails then this leaves files on disk. The `tempfile` module has context managers for ensuring cleanup: https://docs.python.org/3/library/tempfile.html

docs/source/user-guide/binding/execution-engine.rst
```diff
@@ -79,6 +79,13 @@ The ExecutionEngine class
             or a object file instance. Object file instance is not usable after this
             call.
 
+   * .. method:: add_archive(archive_file)
+
+        Add the symbols from the specified static archive file to the execution
+        engine. It is a fatal error in LLVM if the *archive_file* does not exist.
+
+        * *archive* str: a path to the static object file
```
gmarkall, line 87: The parameter name doesn't match that from above.  Is "static object file" appropriate here, or is "archive file" more consistent?

```suggestion
        * *archive_file* str: path to the archive file
```

docs/source/user-guide/binding/execution-engine.rst
```diff
@@ -79,6 +79,13 @@ The ExecutionEngine class
             or a object file instance. Object file instance is not usable after this
             call.
 
+   * .. method:: add_archive(archive_file)
+
+        Add the symbols from the specified static archive file to the execution
```
gmarkall, line 84: What does the word "static" mean here? My understanding is that in general, an archive contains unlinked objects, and those objects might eventually take part in a static or dynamic linking process (depending on how they were compiled, of course).

According to the LLVM docs, it looks like adding an archive uses

```suggestion
        Add the symbols from the specified archive file to the execution
```

gmarkall (commented):
Just had a quick skim (not a proper review) and added some comments on the diff. I haven't yet considered whether the strategy of compiling and archiving during the test run is the best path, but I've added a couple of pointers on the test code anyway.

docs/source/user-guide/binding/execution-engine.rst
testhound: @gmarkall static here means static archive (i.e libc.a) vs shared library (libc.so). I double checked the code on the LLVM C++ side of things and it is only expecting a static archive.

sklam:
```
cc1: fatal error: /home/vsts/miniconda3/envs/cienv/lib/python3.8/site-packages/llvmlite/tests/a.c: No such file or directory
```
see error log at https://dev.azure.com/numba/numba/_build/results?buildId=14315&view=logs&j=fd18f1e2-1ae3-5f4d-9319-996752c1f67c&t=85c24074-1b78-5ee7-1f6d-6e0b210c5911&l=27

You'll need to add the .c files as package data. They are not being installed.

testhound:
@sklam and @gmarkall all errors resolved.

ffi/executionengine.cpp
```diff
@@ -167,6 +169,42 @@ LLVMPY_MCJITAddObjectFile(LLVMExecutionEngineRef EE, LLVMObjectFileRef ObjF) {
...
+        MemoryBuffer::getFile(ArchiveName);
+
+    std::error_code EC = ArBufOrErr.getError();
+    if (EC) {
+        *OutError = LLVMPY_CreateString(EC.message().c_str());
+        return 1;
+    }
+
+    Expected<std::unique_ptr<object::Archive>> ArchiveOrError =
+        object::Archive::create(ArBufOrErr.get()->getMemBufferRef());
```
gmarkall, line 189: The `llvm::object` namespace is already visible in this scope:
```suggestion
        Archive::create(ArBufOrErr.get()->getMemBufferRef());
```

ffi/executionengine.cpp
```diff
@@ -167,6 +169,42 @@ LLVMPY_MCJITAddObjectFile(LLVMExecutionEngineRef EE, LLVMObjectFileRef ObjF) {
...
+
+    if (!ArchiveOrError) {
+        auto takeErr = ArchiveOrError.takeError();
+        std::string archiveErrStr =
+            "Unable to load archive: " + std::string(ArchiveName);
+        *OutError = LLVMPY_CreateString(archiveErrStr.c_str());
```
gmarkall, line 195: Rather than providing a generic error message, passing along the error provided by LLVM will probably be more helpful to the user. I'm not sure of the cleanest way to do it (I often have some trouble picking and choosing between LLVM C++ and C APIs, for example, but here's one way to do it:

```suggestion
        LLVMErrorRef errorRef = wrap(std::move(takeErr));
        *OutError = LLVMPY_CreateString(LLVMGetErrorMessage(errorRef));
```

gmarkall (commented):
Just had a look at the C++ side of things, which seems mostly OK but I've added some comments on there.

ffi/executionengine.cpp
```diff
@@ -167,6 +169,42 @@ LLVMPY_MCJITAddObjectFile(LLVMExecutionEngineRef EE, LLVMObjectFileRef ObjF) {
...
 }
 
+API_EXPORT(int)
+LLVMPY_MCJITAddArchive(LLVMExecutionEngineRef EE, const char *ArchiveName,
+                       const char **OutError) {
+    using namespace llvm;
+    using namespace llvm::object;
+    auto engine = unwrap(EE);
+
+    ErrorOr<std::unique_ptr<MemoryBuffer>> ArBufOrErr =
```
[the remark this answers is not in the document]
testhound: @gmarkall it's up to you, I find auto can really hurt in a situation like this especially when you need to determine when to unpack such as getting the error message.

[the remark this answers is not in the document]
testhound: @gmarkall to your comment "... which will probably be linked statically", a static archive can only be linked statically.

testhound:
@gmarkall all feedback incorporated.

testhound:
@gmarkall ready for review.

ffi/executionengine.cpp
```diff
@@ -167,6 +169,40 @@ LLVMPY_MCJITAddObjectFile(LLVMExecutionEngineRef EE, LLVMObjectFileRef ObjF) {
...
+        Archive::create(ArBufOrErr.get()->getMemBufferRef());
+
+    if (!ArchiveOrError) {
+        auto takeErr = ArchiveOrError.takeError();
+        LLVMErrorRef errorRef = wrap(std::move(takeErr));
+        *OutError = LLVMPY_CreateString(LLVMGetErrorMessage(errorRef));
+        return 1;
+    }
+
+    std::unique_ptr<MemoryBuffer> &ArBuf = ArBufOrErr.get();
```
apmasell, line 198: These can also be simplified to:
```
    OwningBinary<object::Archive> owningBinaryArchive(std::move(*ArchiveOrError),
                                                      std::move(*ArBufOrErr));
```

llvmlite/tests/test_binding.py
```diff
@@ -1935,6 +1981,60 @@ def test_get_section_content(self):
...
+            target_machine = self.target_machine(jit=False)
+
+            jit = llvm.create_mcjit_compiler(self.module(self.mod_archive_asm),
+                                             target_machine)
+
+            # Create compiler with default options
+            c = new_compiler()
+            customize_compiler(c)
+            workdir = tmpdir.name
+            srcdir = os.path.dirname(llvmlite.tests.__file__) + "/"
```
apmasell, line 2010: Please use `os.path.join` to make this portable and same for the path manipulation below and don't add the `.c` files to the repository. Store the code as inline strings and write them to a temporary directory during the test.

setup.py
```diff
@@ -100,6 +100,7 @@ def run(self):
         from llvmlite.utils import get_library_files
         self.distribution.package_data = {
             "llvmlite.binding": get_library_files(),
+            "llvmlite.tests": ["*.c"]
```
apmasell, line 103: When you inline the `.c` files, revert these changes.

testhound:
@apmasell all your feedback has been incorporated and the PR is ready for further review.

llvmlite/tests/test_binding.py
```diff
@@ -1941,6 +1986,60 @@ def test_get_section_content(self):
...
+class TestArchiveFile(BaseTest):
+    mod_archive_asm = """
+        ;ModuleID = <string>
+        target triple = "{triple}"
+
+        declare i32 @__multiply_accumulate(i32 %0, i32 %1, i32 %2)
+        declare i32 @__multiply_subtract(i32 %0, i32 %1, i32 %2)
+    """
+
+    @unittest.skipUnless(sys.platform.startswith('linux'),
```
apmasell, line 1998: Why is this Linux specific? It needs to be tested on all platforms.

testhound: @apmasell thanks I modified the code to test on all platforms. I am now getting errors on all versions of Windows and a single version of macOS. Do you have any feedback you can offer?

gmarkall: One thing that stands out to me is that the library is called `foo` but you're looking for `libfoo.lib` on Windows. It's not usual for Windows library names to be prefixed with `lib` - maybe the library is called `foo.lib`?

testhound: @gmarkall thanks that got me further on Windows, still more Windows errors to debug.

gmarkall: On Windows you can't delete an open file. I wonder if the file is still open when you attempt to clean up the temp dir, causing the issue you see - perhaps deleting `jit` will help the cleanup to succeed.

testhound:
@apmasell and @gmarkall I have updated the code. I am skipping the test on Windows and provided a detailed comment at the top of the test class. There is a error on macOS/Python 3.9 that appears related to a missing llvm-ar. All other versions of macOS succeed.

sklam:
Note for https://github.com/numba/llvmlite/pull/902/commits/269cd6d713d00be8cf77d856d8c967029bf3e2ea
The Anaconda build of Python 3.9 is using a hardcoded path for `llvm-ar` from the build environment. I tested it on conda-forge build of Python 3.9 and it works without problem.

testhound:
@apmasell and @gmarkall ready for review.

apmasell:
If there's a way to get the binary to load on Windows using a non-`distutils` tool chain, that should be done rather than skipping the test.

llvmlite/tests/test_binding.py
```diff
@@ -1941,6 +2006,100 @@ def test_get_section_content(self):
                 self.assertEqual(s.data().hex(), issue_632_text)
 
 
+# Skip this test on Windows due to the fact that
+# the /GL cross-module optimization has the side-effect
+# of removing external symbols in this test case.
```
sklam, line 2011: Maybe adding a `DEF` file can fix this; see https://devblogs.microsoft.com/oldnewthing/20140321-00/?p=1433

testhound:
> If there's a way to get the binary to load on Windows using a non-`distutils` tool chain, that should be done rather than skipping the test.

Are there other 



> If there's a way to get the binary to load on Windows using a non-`distutils` tool chain, that should be done rather than skipping the test.

@apmasell unless you are aware of another solution it appears I will need to invoke the visual studio c++ compiler directly from the Python code for Windows plaltforms.

testhound:
@apmasell, @sklam and @gmarkall Windows error resolved and ready for review.

apmasell (approved):

PR #903, opened 2023-02-06
Fix flake8 config and style for flake8 6

gmarkall:
Similar to numba/numba#8680, the config file needs updating for Flake8 6. One new error was detected, which is also corrected.


stuartarchibald (approved):
Thanks for the patch.

PR #904, opened 2023-02-06
Create GitHub Action for llvmlite release

apmasell:
Build a new GitHub Action to build wheels for llvmlite

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,124 @@
+name: Release
+
+permissions:
+  contents: write
+
+on:
+  push:
+    tags:
+      - v[0-9]+.*
+
+jobs:
+  # First we are going to create a task that generates a new release in GitHub
+  # as a draft. All the wheels will end up being uploaded here at the end.
+  create-release:
+    runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.create-gh-release.outputs.computed-prefix }}${{ steps.create-gh-release.outputs.version }}
+    steps:
+      - uses: actions/checkout@v3
+      - id: create-gh-release
+        uses: taiki-e/create-gh-release-action@v1
+        with:
+          draft: true
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+  build-wheels:
+    # Build binary wheels for the platforms we care about using cibuildwheel.
+    # Now, cibuildwheel has an excellent set of features to try to build a
+    # hermetic wheel, but we, dear reader, need to absolutely violate that
+    # because LLVM is too painful to compile as part of the build process, but
+    # we want to statically link it into the final wheel. So, we take a conda
+    # package and inject it into the build environment, a state of affairs that
+    # cibuildwheel isn't going to cooprate much on. The
+    # `buildscripts/github/config.toml` defines a bunch of paths to where in
+    # the build environment all that conda stuff will be placed.
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-22.04, macOS-11, windows-2019 ]
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.9'
+
+      - name: Install cibuildwheel
+        run: python -m pip install cibuildwheel==2.12.0
+
+      - name: Build wheels
+        run: "python -m cibuildwheel --output-dir wheelhouse --config-file buildscripts/github/config.toml"
+        env:
+          # On a 64-bit machine, cibuildwheel will try to build 32-bit
+          # packages. Since LLVM is only build for 64-bit, linking will go
+          # badly.
+          CIBW_ARCHS: auto64
+          # I cannot tell you how much this isn't going to work on musllinux,
+          # so don't try to compile against it.
+          CIBW_BUILD: cp39-*
+          CIBW_SKIP: cp39-musllinux*
+          # Run our little script to inject LLVM into the build environment. See script for details.
+          CIBW_BEFORE_BUILD: "{project}/buildscripts/github/install_llvmdev.sh"
+          CIBW_BEFORE_BUILD_WINDOWS: "pwsh {project}/buildscripts/github/install_llvmdev.ps1"
+          # cibuildwheel is going to assume we are going for maximum
+          # compatibility and that's can't happen with LLVM because it needs a
+          # relatively new C++ compiler. So, instead, we tell it to require a
+          # GLIBC of 2.31. If you get inscrutable compatibility errors, try
+          # `auditwheel -v repair` to get _too much_ output.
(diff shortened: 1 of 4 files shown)


buildscripts/github/release-environment.yml
```diff
@@ -0,0 +1,6 @@
+name: llvmlite-release-env
+channels:
+  - numba
+  - defaults
+dependencies:
+  - llvmdev=11.*
```
gmarkall, line 6: Should it be this?

```suggestion
  - llvmdev=14.*
```

apmasell: Yes, it should. I was working this before LLVM14 got merged.

esc:
I have completed a visual inspection of the proposed changes, and they look fine to me. The comments are a bit sparse, 

 But, the big question is how do we test this?  @apmasell how did you test this when you developed it? For PyPi, I think there is a pypi test server where we can upload stuff? For the GitHub releases,  do we run this in a fork and then check the artifacts?

Potential additions:

* Probe PyPi after the artifacts have been uploaded to ensure that they did arrive?
* Ensure that the correct llvmdev was installed, we explicitly require the one from the `numba` channel? I know that conda can be a bit picky sometimes during resolution.

I appreciate the use of the `™️` character.

apmasell:
You can see the artifacts that are built on my fork: https://github.com/apmasell/llvmlite/releases/tag/v99.99.99
As for testing the PyPI part, I haven't found a good way to do that. I think that's a non-critical failure. If it fails during the release, we can manually upload the wheels it builds as we do currently and adjust the action

How do we currently verify the llvmdev version?

esc:
> You can see the artifacts that are built on my fork: https://github.com/apmasell/llvmlite/releases/tag/v99.99.99 As for testing the PyPI part, I haven't found a good way to do that. I think that's a non-critical failure. If it fails during the release, we can manually upload the wheels it builds as we do currently and adjust the action
> 
> How do we currently verify the llvmdev version?

We install using `numba::llvmdev=14` ...

esc:
I think it'll be fine like this, actually, when would be a good time to merge this? after the current release?

apmasell:
I'm confused. Should `numba::llvmdev=14` be different than the environment file?

This should get merged after the release, I think.

esc:
> I'm confused. Should `numba::llvmdev=14` be different than the environment file?

I don't know exactly, but the following two have sometimes produced different results in the past:

```
conda install -c numba llvmdev`
```

and

```
conda install numba::llvmdev
```

Hence I am advocating for caution here, that we don't run into any conda bugs.

> This should get merged after the release, I think.

I concur.

apmasell:
Okay. Unfortunately, the environment files don't allow the `channel::package` syntax.

apmasell:
Sigh. This is the wrong fork. I've pushed a different method.

esc:
@apmasell what do I look at next to continue the review?

apmasell:
I've replaced the total work, so please pretend it's a fresh PR.

.github/workflows/release.yml
```diff
@@ -0,0 +1,124 @@
...
+  contents: write
+
+on:
+  push:
+    tags:
+      - v[0-9]+.*
+
+jobs:
+  # First we are going to create a task that generates a new release in GitHub
+  # as a draft. All the wheels will end up being uploaded here at the end.
```
esc, line 13:
```suggestion
  # as a draft. All the wheels will end up being uploaded here when complete.
```

`end` was being used twice in this sentence.

.github/workflows/release.yml
```diff
@@ -0,0 +1,124 @@
...
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+  build-wheels:
+    # Build binary wheels for the platforms we care about using cibuildwheel.
+    # Now, cibuildwheel has an excellent set of features to try to build a
+    # hermetic wheel, but we, dear reader, need to absolutely violate that
+    # because LLVM is too painful to compile as part of the build process, but
+    # we want to statically link it into the final wheel. So, we take a conda
+    # package and inject it into the build environment, a state of affairs that
+    # cibuildwheel isn't going to cooprate much on. The
```
esc, line 33:
```suggestion
    # cibuildwheel isn't going to cooperate much on. The
```

esc, line 33:
```suggestion
    # cibuildwheel isn't going to cooperate much on. The
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,124 @@
...
+          CIBW_ARCHS: auto64
+          # I cannot tell you how much this isn't going to work on musllinux,
+          # so don't try to compile against it.
+          CIBW_BUILD: cp39-*
+          CIBW_SKIP: cp39-musllinux*
+          # Run our little script to inject LLVM into the build environment. See script for details.
+          CIBW_BEFORE_BUILD: "{project}/buildscripts/github/install_llvmdev.sh"
+          CIBW_BEFORE_BUILD_WINDOWS: "pwsh {project}/buildscripts/github/install_llvmdev.ps1"
+          # cibuildwheel is going to assume we are going for maximum
+          # compatibility and that's can't happen with LLVM because it needs a
```
esc, line 66:
```suggestion
          # compatibility and that can't happen with LLVM because it needs a
```

buildscripts/github/install_llvmdev.ps1
```diff
@@ -0,0 +1,43 @@
...
+    {
+        throw 'Native Failure'
+    }
+}
+
+Invoke-WebRequest -Uri "https://repo.continuum.io/miniconda/Miniconda3-latest-Windows-x86_64.exe" -OutFile "miniconda.exe"
+
+mkdir miniconda3
+Start-Process miniconda.exe -Wait -ArgumentList @('/S',"/D=$(Get-Location)\miniconda3")
+./miniconda3/condabin/conda.bat install numba::llvmdev=14
```
esc, line 40: Can you comment this line too to mention that we are explicitly asking for the `llvmdev` from the `numba` channel.

esc (changes requested):
First pass: Typo fix suggestions and request for additional comments.

esc:
@apmasell thank you for the updates, I have given this a first pass to familiarize myself with the code and found a couple of typos in the process. I mostly understand the setup, as I have been doing llvmlite releases in the past. 

Since this now produces artifacts, we need to devise a strategy to test them accordingly. Ideally, I would like to test each python x platform combination of wheel in combination with an appropriate Numba. The real litmus test for llvmlite is always the Numba test suit. Do you have any ideas as to how we might go about this?

esc:
> I've made the changes requested. For testing, I'm not sure. We have a manual validation process for the existing packages, so I think the right reasonable thing to do is merge this as-is, allow it to build packages during the next, compare those packages to the packages we've build using existing tooling and the same validation. We can then make a better informed decision about what validation we need.

OK

> It is my understand that the release process checks that the build tag is good and then assumes that the package built from that tag needs minimal validation. I think that's something we might want to re-evaluate. Doing anything heavy-weight will require us to get more GitHub Action compute resources, so, that will have to wait until later.

Understood.

Fire in the hole then 🔥 🕳️

esc (approved):

PR #905, opened 2023-02-14
Add YouCompleteMe configuration file and ignore vim swap files

gmarkall:
I've been working with this config locally for a long time, and I think it might be helpful to make more broadly available, as vim and YCM are a popular development combination. This PR has no effect on the code, the tests, or the build, but helps make llvmlite development on vim a smoother experience out-of-the-box for anyone using Vim + YCM, and working with Conda.

YouCompleteMe (https://github.com/ycm-core/YouCompleteMe) is a code-completion engine for vim. This PR adds a configuration file for it that enables it to understand the C++ FFI code in llvmlite by pointing to the correct include dirs (as long as one is working within a Conda environment), setting the relevant C++ standard, and ensuring that all header files are treated as C++.

Additionally, Vim swap files are added to the gitignore, so that they don't show up in `git status`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,4 @@ coverage.xml
 MANIFEST
 docs/_build/
 docs/gh-pages/
-
+.*.swp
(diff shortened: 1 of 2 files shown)


apmasell (approved):

gmarkall:
Thanks for the review!

PR #906, opened 2023-02-14
Replace importlib-resources legacy API use

sklam:
Suppress deprecation warnings due to use of legacy API in importlib-resources

The replacement implementation follows advice from https://importlib-resources.readthedocs.io/en/latest/using.html#migrating-from-legacy and adapts code from https://github.com/python/importlib_resources/blob/66ea2dc7eb12b1be2322b7ad002cefb12d364dff/importlib_resources/_legacy.py#L72-L84.

The DeprecationWarning was first noticed in numba test `numba.tests.test_import.TestNumbaImport.test_no_accidental_warnings`

### DIFF
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -1,6 +1,6 @@
 import ctypes
 import threading
-import importlib.resources
+import importlib.resources as _impres
 
 from llvmlite.binding.common import _decode_string, _is_shutting_down
 from llvmlite.utils import get_library_name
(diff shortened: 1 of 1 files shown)


llvmlite/binding/ffi.py
```diff
@@ -151,12 +152,31 @@ def __call__(self, *args, **kwargs):
...
+                             else _impres.path)
+
+
 _lib_name = get_library_name()
 
 
 pkgname = ".".join(__name__.split(".")[0:-1])
 try:
-    _lib_handle = importlib.resources.path(pkgname, _lib_name)
+    _lib_handle = _importlib_resources_path(pkgname, _lib_name)
```
guilhermeleobas, line 179: Once llvmlite bumps Python min version to 3.9, the code above can be safely replaced by:

```python
from importlib.resources import as_file, files
_lib_handle = as_file(files(pkgname).joinpath(_lib_name))
```

guilhermeleobas (approved):
LGTM

Issue #900, opened 2023-01-19
llvmlite opaque pointer support


folded:
The plan of record is that non-opaque pointer support remains in LLVM 16 (with decreasing test coverage) and will be removed in LLVM 17: https://reviews.llvm.org/D140487

The direct breakages I saw when opaque pointers are enabled by default are caused by llvmlite's method of getting ABI size and alignment for values, which it does by creating a module, declaring a global value of the right type, and then interrogating that.

Someone more knowledgeable than me pointed out "A GlobalValue's type is a pointer, but instead of GV->getType()->getElementType(), you can do GV->getValueType().", and implementing this succeeded in getting tests passing again.

I'll prepare a patch for review to close this bug, but I wanted to kick off a discussion here to make sure that what I'm proposing doesn't have other issues that I'm not aware of and maybe haven't been picked up by test cases.

Changes, in summary are:

* `ValueRef` gets `@property global_value_type`, which returns the value type if `self` is a global value.
* `Type._get_ll_pointer_type` becomes `Type._get_ll_global_value_type` and usages change like so:
```
-        llty = self._get_ll_pointer_type(target_data, context)
-        return target_data.get_pointee_abi_size(llty)
+        llty = self._get_ll_global_value_type(target_data, context)
+        return target_data.get_abi_size(llty)
```
* expose `LLVMABIAlignmentOfType` binding to replace `get_pointee_abi_alignment`
* (with deprecation schedule) remove `LLVMPY_GetElementType`, `LLVMPY_ABISizeOfElementType` and `LLVMPY_ABIAlignmentOfElementType` and associated python code.

j2kun:
Should this issue be closed due to https://github.com/numba/llvmlite/pull/1064 being merged? Or is it waiting to be released? (I ran into this issue when trying to use llvmlite 0.43.0 with clang-18-generated llvm IR, not realizing the version mattered)

PR #908, opened 2023-02-17
[llvmlite] Support opaque pointers.

folded:
Fix for issue #900 

LLVM 14 introduces optional opaque pointers. If opaque pointers are enabled then there is no way to determine the pointee type. LLVM 16 makes this the default.

In llvmlite and numba this is used to determine ABI size and alignment of types by creating a global variable and then asking for its pointee type in order to get a TypeRef. For global variables specifically this can be shortcut via gv->getValueType(),
and doing so removes the current issues when opaque pointers are enabled.

This change has been tested with LLVM 11-14 and so also includes some other changes required to make llvmlite compile on those versions.

### DIFF
--- a/ffi/core.cpp
+++ b/ffi/core.cpp
@@ -1,6 +1,7 @@
 #include "core.h"
 
 #include "llvm-c/Support.h"
+#include "llvm/IR/LLVMContext.h"
 
 extern "C" {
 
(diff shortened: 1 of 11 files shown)


dlee992:
For a record: I guess this PR could be surpassed by #1042?

gmarkall:
> For a record: I guess this PR could be surpassed by #1042?

Maybe technically, but we won't be able to merge #1042 because it contains the whole LLVM 17 transition in one go, and we need to adopt a more gradual strategy.

dlee992:
Yeah, totally understandable.

gmarkall:
Opaque pointers were implemented in #1064 - closing.

Many thanks for this initial work / reference, @folded !

PR #910, opened 2023-02-28
Aarch64 split build for LLVM14

sklam:


sklam:
https://github.com/numba/llvmlite/pull/910/commits/0a300974bdfdc182a5de02a61728aee116650eb8: I made sure that git history from current llvm11 files are retained

sklam:
Merged in #830 to fix CI

sklam:
pass BFID: `llvmdev_73`

stuartarchibald (approved):
Thanks for the patch. On inspection, this seems to split out the `llvmdev` builds such that on `linux-aarch64` `conda build` will build `llvmdev` based on version 11.x and on other OS/architectures it will build based on version 14.x. This to help work around issues such as https://github.com/numba/numba/issues/8738.

I've checked the logs for the referenced buildfarm run of this patch and the correct LLVM versions appear to have been built on the expected platforms as described above.

Thanks also for making the minor changes to `test_refprune.py` and `ffi/Makefile.osx` as discussed OOB. That the public CI testing of the former passed suggests it is fine, all that remains is a rebuild of the OSX packages to pick up the change from `-O3` to `-O2` following merge.

Thanks again for working on this.

PR #917, opened 2023-03-14
Remove maximum Python version limit

### LINKED ISSUE #912: “Escape hatch” for maximum Python version check
---

## Feature request

I understand that `llvmlite` is developed together with Numba, and new Python interpreter versions will probably continue to be officially unsupported here until both packages have fully-validated support ready.

In Fedora Linux, we are currently unable to package Numba because we must build all Python packages with the current Python interpreter version. Because Fedora is a fast-moving distribution, a new CPython release happens almost at the same time as the first Fedora Linux stable release that contains it; incompatibilities are sorted out using CPython pre-releases. To be successfully included in Fedora, Numba would need to have support for a new Python version around the time of its final release—at least in some kind of development branch.

Despite the obstacles to packaging Numba, `llvmlite` is [usefully packaged in Fedora Linux](https://src.fedoraproject.org/rpms/python-llvmlite), where it is used by Sympy and Mathics3. We have been able to build it for Python 3.11 and 3.12 by simply loosening the bound on the maximum Python version in `setup.py`. Because we run the full test suite in the RPM build, we have pretty good confidence in the resulting package even without official upstream support for the current Python release.

So that we, and other Linux distributions, do not have to carry a downstream patch, would you consider a PR that allowed explicitly bypassing the maximum Python version check—perhaps by setting an environment variable like `LLVMLITE_NO_MAX_PYTHON_CHECK`?

Alternatively, would you consider supporting new Python releases in `llvmlite` once its tests pass on them, rather than waiting for Numba to catch up?

apmasell:
Allow any new version of Python for llvmlite. Closes #912


gmarkall (approved):
LGTM.

stuartarchibald (changes requested):
As noted at the Numba triage meeting. The documentation needs updating to reflect that whilst the `llvmlite` source distribution has no upper bound on the Python version, it is only supported and tested on certain Python versions.

apmasell:
The documentation already makes the guarantee that llvmlite will work with 3.8 or later and makes no mention of an upper limit.

setup.py
```diff
@@ -33,13 +32,12 @@ def _version_info_str(int_tuple):
 def _guard_py_ver():
     currrent_python_version = sys.version_info[:3]
     min_py = _version_info_str(min_python_version)
-    max_py = _version_info_str(max_python_version)
     cur_py = _version_info_str(currrent_python_version)
 
-    if not min_python_version <= currrent_python_version < max_python_version:
-        msg = ('Cannot install on Python version {}; only versions >={},<{} '
+    if not min_python_version <= currrent_python_version:
```
stuartarchibald, line 37:
```suggestion
    if not min_python_version <= current_python_version:
```
and fix up. Thanks.

README.rst
```diff
@@ -57,7 +57,8 @@ Key Benefits
 Compatibility
 =============
 
-llvmlite works with Python 3.8 and greater.
+llvmlite works with Python 3.8 and greater. We attempt to test with the lastest
+Python version that is available in CI.
```
stuartarchibald, line 61:
```suggestion
llvmlite works with Python 3.8 and greater. We attempt to test with the latest
Python version, this can be checked by looking at the public CI builds.
```
Perhaps this?

apmasell: Sure.

stuartarchibald (approved):
Thanks for the updates @apmasell, I approve 79b8aff and 2c767a6 following @gmarkall's approval of 9ea5668.

azure-pipelines.yml
```diff
@@ -35,6 +35,10 @@ jobs:
         PYTHON: '3.10'
         CONDA_ENV: cienv
         RUN_FLAKE8: yes
+      py311:
+        PYTHON: '3.11'
+        CONDA_ENV: cienv
+        RUN_FLAKE8: yes
```
sklam, line 41: This will duplicate 3.11 testing

sklam: merged code looks like:

```
      py311:
        PYTHON: '3.11'
        CONDA_ENV: cienv
        RUN_FLAKE8: yes
        RUN_CLANG_FORMAT: yes
      py311:
        PYTHON: '3.11'
        CONDA_ENV: cienv
        RUN_FLAKE8: yes
```

sklam (changes requested):
I suggest a merge main to resolve the below

apmasell:
Closed in favour of #931

PR #918, opened 2023-03-15
Enable ppc64le wheels

mgiessing:
Minimal patch to enable ppc64le wheels.
This is no change content-wise :-)

### DIFF
--- a/buildscripts/manylinux_2_17/build_llvmdev.sh
+++ b/buildscripts/manylinux_2_17/build_llvmdev.sh
@@ -3,7 +3,7 @@
 cd $(dirname $0)
 source ./prepare_miniconda.sh $1
 source /root/miniconda3/bin/activate buildenv
-if [[ $ARCH == "aarch64" ]] ; then
+if [[ $ARCH == "aarch64" || $ARCH == "ppc64le" ]] ; then
     export BUILD_CHANNELS="-c conda-forge"
 fi
 echo "BUILD_CHANNELS:  $BUILD_CHANNELS"
--- a/buildscripts/manylinux_2_17/build_llvmlite.sh
+++ b/buildscripts/manylinux_2_17/build_llvmlite.sh
@@ -20,7 +20,7 @@ conda create -y -n $envname
 source activate $envname
 # Install llvmdev
 
-if [[ $ARCH == "aarch64" ]] ; then
+if [[ $ARCH == "aarch64" || $ARCH == "ppc64le" ]] ; then
     conda install -y numba/label/manylinux2014::llvmdev --no-deps
 else
     conda install -y -c numba/label/manylinux2014 llvmdev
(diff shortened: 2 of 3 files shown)


mgiessing:
I'm not entirely sure about the wheel CI/CD since I don't see any github actions, cibuildwheel.
It seems you're using azure pipelines, but how is the aarch64 wheel produced? 
I don't see any CPU specification in the yaml and doesn't it need any kind of emulation (prob. similar to ppc64le?)

Thanks! :)

gmarkall:
Thanks for the PR! I'm not entirely sure how the wheels are built, but I think  @apmasell might be able to comment on what we're doing now and how we might plan to change it in the future.

gmarkall:
.. or perhaps @esc knows. A question for both about the practicality of doing it on the available resources and time / admin constraints, rather than the technical nature of this PR itself: if this PR were merged, would it be straightforward to build the wheels on a ppc machine and upload them?

mgiessing:
Yeah, I already build and upload ppc64le wheels (for llvmlite & numba) to my gemfury repo, however having it upstreamed would simplify other projects/processes that rely on this. (I'm currently using NVIDIA triton-inference-server that uses numba/llvmlite)

I also understand the additional effort that might come with more CPU architectures, but I'd be totally happy with having something like a "Tier 3 support" where the wheel is just build but not necessarily tested upstream during development process (which is probably done for Linux x86 & MacOS x86/aarch64).

mgiessing:
Something similar is done for the qiskit project[1] where architectures like ppc64le or s390x got T3 support - that already helps a lot!

[1] https://qiskit.org/documentation/getting_started.html#platform-support

apmasell:
The patch looks fine. Right now, we're building wheels in a private environment, so these changes won't trigger building PPC wheel, though they are necessary preconditions. We do test on PPC, so everything should work even though there are no wheels.

I am moving the wheel builds to GitHub Actions (see #904), but there are no PPC machines available for GHA. We plan to supply our own, but that's work we haven't started. It's on the road map, but the plan is to get x86-64 wheel and conda builds deployed for all Numba projects, then ARM, then PPC.

esc:
@mgiessing Hi and sorry for the long radio silence. Quick update here: we managed to establish a Tier based support policy framework for the Numba project:

PR: https://github.com/numba/numba/pull/10283

Rendered doc: https://numba.readthedocs.io/en/latest/reference/support_tiers.html

Currently ppc64le is in Tier 2. In order for it to be promoted to Tier 1, the following pre-conditions would need to be met:


```
    A conda based Python distribution for the operating system and hardware target is required. This is because the public CI
    CD for the Numba stack is driven by conda and conda packages (even for wheel builds). The purpose of this condition is to 
    reduce the overall CI/CD maintenance burden.

    The operating system and hardware target must be supported by GitHub Actions on the free-tier.

    The release package type must be conda or wheel.

    The distribution system must be PyPI or Anaconda repositories.
```

Do you know anything about the state of those pre-conditions?

PR #919, opened 2023-03-15
Experimental OrcJITv2 support

gmarkall:
For working on OrcJITv2 support.

Current status:

- Passes tests with LLVM 14.0.6
- Numba support needs work.
- MCJIT example ported to LLJIT works.

Current known to-do list:

- [ ] Add documentation.
- [ ] Invalidate modules passed to `add_ir_module()` since they're not usable after that.
- [X] Remove the hack to keep a global list of thread safe modules to keep them alive after compilation.
- [ ] Remove abort statements.
- [ ] Provide the resource tracker back to the caller of `add_ir_module()`, wrapped in something that removes it when it is garbage-collected.

Not todo:

- Implement an object cache. This will come later.


gmarkall:
This is no longer relevant, closing.

PR #922, opened 2023-03-16
Fix AzureCI not using llvm14 on windows

sklam:
we missed this config in the last PR.

### DIFF
--- a/buildscripts/azure/azure-windows.yml
+++ b/buildscripts/azure/azure-windows.yml
@@ -34,7 +34,7 @@ jobs:
         call conda remove --all -q -y -n %CONDA_ENV%
         call conda create -n %CONDA_ENV% -q -y python=%PYTHON% cmake
         call activate %CONDA_ENV%
-        call conda install -y -q -c numba/label/dev llvmdev="14.*"
+        call conda install -y -q -c numba/label/dev llvmdev="14.*" libxml2
       displayName: 'Before Install'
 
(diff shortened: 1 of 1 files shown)


sklam:
This is minimal work to get what we have working but there are two issues that need to be fixed:

1. `setup_conda_environment.cmd` is unused. shouldn't azure-windows.yml use that instead of inlined scripts? (CC @esc)
2. i don't think we need libxml2. We should turn it off for the windows build. Is there a reason to enable libxml2? (CC @apmasell)

gmarkall (approved):
I think this is good to merge based on the fact that it enables Windows to build with LLVM 14 (I've manually verified that LLVM 14 is picked up in the Windows build for each Python version.

Regarding @sklam's comments:

1. It does indeed not use `setup_conda_environment.cmd` - looks like during initial setup some weirdness was identified with it that would need working out: https://github.com/numba/llvmlite/pull/413/commits/276d2023bb2dc215f75a976b631bb88013c2c427#r239559309
2. I notice that on Linux / macOS we explicitly disable libxml2 in the llvmdev build (https://github.com/numba/llvmlite/blob/main/conda-recipes/llvmdev/build_llvm14.sh#L31) but we're not doing that for Windows. I guess that might be why libxml2 seems to be required on Windows.

apmasell:
I think we had it for the remarks output.

PR #932, opened 2023-03-31
Fix wheel builds

sklam:


sklam:
BFID `llvmlite_wheel_39`

esc (commented):
I checked the 
* permission changes (`+x`) -- fine.
* path changes (`manylinux_2_17`) -- fine.
* changes to `build.sh` and `meta.yaml` -- see below.

```
zsh» diff llvmdev/build.sh llvmdev_manylinux2014/build.sh
97,101d96
<
< # SVML tests on x86_64 arch only
< if [[ $ARCH == 'x86_64' ]]; then
<    bin/opt -S -vector-library=SVML -mcpu=haswell -O3 $RECIPE_DIR/numba-3016.ll | bin/FileCheck $RECIPE_DIR/numba-3016.ll || exit $?
< fi
```

and

```
zsh» diff llvmdev/meta.yaml llvmdev_manylinux2014/meta.yaml
31a32,33
>     - CFLAGS
>     - CXXFLAGS
39,45d40
<     # We cannot do this on macOS or windows
<     # OSX already has llvm so has to be handled
<     # at build.sh time
<     # Windows needs to build using vs2015_runtime
<     # irrespective of python version
<     - {{ compiler('c') }} # [unix and not (armv6l or armv7l)]
<     - {{ compiler('cxx') }} # [unix and not (armv6l or armv7l)]
47c42
<     - make # [unix and not (armv6l or armv7l or aarch64)]
---
>     - make
60,61d54
<   files:
<     - numba-3016.ll
```

Since it passed on the build farm, I think these minor difference in recipe may be inconsequential to producing a working binary wheel artifact or may be specific to the manylinux build.

The only thing I was worried about is the removal of:

```
  string: "manylinux2014h{{ PKG_HASH }}"
```

IIRC this was used somehow to distinguish the non-manylinux2014 `llvmdev` packages from the manylinux2014 ones. But if this is no longer an issue, that's fine too.

sklam:
> * changes to `build.sh` and `meta.yaml` -- see below.
> 
> ```
> zsh» diff llvmdev/build.sh llvmdev_manylinux2014/build.sh
> 97,101d96
> <
> < # SVML tests on x86_64 arch only
> < if [[ $ARCH == 'x86_64' ]]; then
> <    bin/opt -S -vector-library=SVML -mcpu=haswell -O3 $RECIPE_DIR/numba-3016.ll | bin/FileCheck $RECIPE_DIR/numba-3016.ll || exit $?
> < fi
> ```
> 
> and
> 
> ```
> zsh» diff llvmdev/meta.yaml llvmdev_manylinux2014/meta.yaml
> 31a32,33
> >     - CFLAGS
> >     - CXXFLAGS
> 39,45d40
> <     # We cannot do this on macOS or windows
> <     # OSX already has llvm so has to be handled
> <     # at build.sh time
> <     # Windows needs to build using vs2015_runtime
> <     # irrespective of python version
> <     - {{ compiler('c') }} # [unix and not (armv6l or armv7l)]
> <     - {{ compiler('cxx') }} # [unix and not (armv6l or armv7l)]
> 47c42
> <     - make # [unix and not (armv6l or armv7l or aarch64)]
> ---
> >     - make
> 60,61d54
> <   files:
> <     - numba-3016.ll
> ```
> 
> Since it passed on the build farm, I think these minor difference in recipe may be inconsequential to producing a working binary wheel artifact or may be specific to the manylinux build.

All these are changes to match up to the old manylinux scripts:
- `build.sh` and `meta.yaml` skips on `numba-3016.ll` 
- compiler flags changes are inherited from: https://github.com/numba/llvmlite/blob/6f95465cbea8036a6bd9a450a4655b484521d842/conda-recipes/llvmdev_manylinux2014/meta.yaml#L35-L36
- and we should use the compiler from the manylinux docker image

sklam:
> The only thing I was worried about is the removal of:
> 
> ```
>   string: "manylinux2014h{{ PKG_HASH }}"
> ```
> 
> IIRC this was used somehow to distinguish the non-manylinux2014 `llvmdev` packages from the manylinux2014 ones. But if this is no longer an issue, that's fine too.

Good catch, I'll need to add it back it

sklam:
> > The only thing I was worried about is the removal of:
> > ```
> >   string: "manylinux2014h{{ PKG_HASH }}"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > IIRC this was used somehow to distinguish the non-manylinux2014 `llvmdev` packages from the manylinux2014 ones. But if this is no longer an issue, that's fine too.
> 
> Good catch, I'll need to add it back it

done in  https://github.com/numba/llvmlite/pull/932/commits/84dfec2173783863a805a7957a042bc475c92967

sklam:
Note: despite missing the buildstring, the `llvmlite_wheel_39` build is unaffected because the use of the `numba/label/manylinux2014` label https://anaconda.org/numba/llvmdev/files?channel=manylinux2014

Issue #727, opened 2021-06-11
How can I extract the value of a Constant?


akcheung:
Is there a way I can retrieve the integer value from a `ValueRef`? I wrote this code:
```
from llvmlite import binding as llvm

asm_sum = r"""
define i32 @foo(i32 %arg) #0 {
  %tmp = alloca i32, align 4
  store i32 1, i32* %tmp, align 4
  ret i32 %arg
}
"""

ref = llvm.parse_assembly(asm_sum, None)
fn = ref.get_function("foo")

for b in fn.blocks:
  for i in b.instructions:
    if i.opcode == "store":
      ops = list(i.operands)
      x = ops[0] + 1   # this generates an error
```

and would like to get the int `1` from it. In llvm's C++ bindings I'd call functions like `getZExtValue`. Is there a way to do so here? I've also tried `ops[0].name` but that didn't work either.

gmarkall:
Unfortunately I don't think there's a way to do that - a wrapper for `getZExtValue` would need to be added to the binding layer (in the `ffi` subdir then the Python layer in `llvmlite/binding` augmented to be able to access it.

(Marking this as a feature request - is this something you'd potentially be interested in contributing?)

PR #936, opened 2023-04-19
Enable querying constants and value kinds

tbennun:
This PR adds bindings for obtaining whether a value is constant, what its `ValueKind` is, and its exact constant value in certain cases (integer, floating point; in other cases the IR string is returned instead).

This is useful for inspecting constant values in Python code.


tbennun:
This PR also fixes #727

apmasell (commented):
Thanks! This is a useful feature. Please add unit tests.

ffi/value.cpp
```diff
@@ -309,6 +309,34 @@ LLVMPY_DisposeOperandsIter(LLVMOperandsIteratorRef GI) {
     delete llvm::unwrap(GI);
 }
 
+API_EXPORT(LLVMBool)
```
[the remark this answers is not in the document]
tbennun: Fixed, thank you!

tbennun:
@apmasell Removed LLVMBool from Python-facing API, added tests, and fixed formatting.

apmasell (approved):
Looks good. Thank you!

llvmlite/binding/value.py
```diff
@@ -140,6 +177,14 @@ def is_instruction(self):
     def is_operand(self):
         return self._kind == 'operand'
 
+    @property
+    def is_constant(self):
```
sklam, line 181: Needs doc update in docs/source/user-guide/binding/type-references.rst

llvmlite/binding/value.py
```diff
@@ -140,6 +177,14 @@ def is_instruction(self):
...
+        return bool(ffi.lib.LLVMPY_IsConstant(self))
+
+    @property
+    def value_kind(self):
```
sklam, line 185: Needs doc update in docs/source/user-guide/binding/type-references.rst

llvmlite/binding/value.py
```diff
@@ -299,6 +344,43 @@ def opcode(self):
                              % (self._kind,))
         return ffi.ret_string(ffi.lib.LLVMPY_GetOpcodeName(self))
 
+    def get_constant_value(self):
```
sklam, line 347: Needs doc update in docs/source/user-guide/binding/type-references.rst

llvmlite/binding/value.py
```diff
@@ -299,6 +344,43 @@ def opcode(self):
...
+            if accuracy_loss.value:
+                warnings.warn(
+                    'Accuracy loss encountered in conversion of constant '
+                    f'value {str(self)}'
+                )
+
+            return value
+
+        # Otherwise, return the IR string
+        return str(self)
```
sklam, line 382: Needs to add a test for the str return

llvmlite/binding/value.py
```diff
@@ -299,6 +344,43 @@ def opcode(self):
...
+                ('little' if little_endian.value else 'big'),
+                signed=False,
+            )
+        elif self.value_kind == ValueKind.constant_fp:
+            # Convert floating-point values to double (Python float)
+            accuracy_loss = c_bool(False)
+            value = ffi.lib.LLVMPY_GetConstantFPValue(self,
+                                                      byref(accuracy_loss))
+            if accuracy_loss.value:
+                warnings.warn(
```
sklam, line 374: Needs to add test for the warning

llvmlite/binding/value.py
```diff
@@ -299,6 +344,43 @@ def opcode(self):
                              % (self._kind,))
         return ffi.ret_string(ffi.lib.LLVMPY_GetOpcodeName(self))
 
+    def get_constant_value(self):
+        """
+        Return the constant value, either as a literal (when supported)
+        or as a string.
+        """
+        if not self.is_constant:
+            raise ValueError('expected constant value, got %s'
```
sklam, line 353: Needs to add a test for the raise

sklam (changes requested):
I have some additional concerns mainly the lack of docs for the new code, and the lack of test coverage for some cases.

llvmlite/binding/value.py
tbennun: @sklam For some reason, despite how much I try, I can't seem to trigger the `losesInfo` boolean with a single constant (maybe this appears during passes). Not sure how to proceed here, any suggestions?

tbennun:
@sklam Added documentation and more tests (apart from testing the warning, which I couldn't reproduce but is part of the LLVM interface, see above comment). I also noticed that adding one argument to `get_constant_value` significantly improves its usage for signed integers, which was apparent in one of the tests.

tbennun:
Any updates on this @sklam ? This is not waiting for me as of two months ago.

esc:
@tbennun sorry, seems like we have dropped the ball on this one. I am changing the label now and will ping @sklam again. It is now scheduled for inclusion in the 0.41 release to be tagged during the week starting 7th of August.

llvmlite/binding/value.py
sklam: Sorry for the delay. 

After some digging into the LLVM source, the floating-point constants are encoded as double-precision so there will not be any information loss here. To force info loss, we will need to use bigger types like the long-double. For example: `%const = fadd fp128 0xLF3CB1CCF26FBC178452FB4EC7F91DEAD, 0xLF3CB1CCF26FBC178452FB4EC7F91973F`. This will require converting the 128-bit float to 64-bit float thus triggering the warning. 

However, this brings to my attention that a warning is not appropriate here. This API is not giving user a choice to read float bigger than double-precision.  In the LLVM OCAML binding, the similar function will return `None` (https://github.com/llvm/llvm-project/blob/1defa781243f9d0bc66719465e4de33e9fb7a243/llvm/bindings/ocaml/llvm/llvm_ocaml.c#L1030-L1032). I think the Python API should raise an exception.

Also, maybe it is better to have separate functions for int constant and float constant, so that the float version can provide more control on what to do if information is lost. For instance, if user is okay with the info loss, they can force the downcast.

sklam (changes requested):
Thanks for adding to the docs. Those looks great.

However, I have followed the rabbit hole and discovered more details to be considered for `get_constant_value` and the info loss issue. See below.

llvmlite/binding/value.py
tbennun: That's great, I agree and already modified the constant float test. Only thing I'm not sure about is the separate functions for int and float constants. One of the main benefits of Python is that you can return multiple types. Maybe this should be another keyword argument, and keyword args should act like preferences?

I am already extending the behavior to support other types of constants (e.g. structs), and it was beneficial to be able to call the same method on the nested data: https://github.com/tbennun/llvmlite/commit/3992ea175c164c9deedb8644b12b138c2e0c0dfc#diff-27ab06b9fa73e854c37e5ccf261269588cd8e02fadbd71c788102bc46c4a2f1bR390

tbennun: @sklam all changes are complete and the PR is ready for re(re)review.

sklam (approved):

PR #942, opened 2023-05-08
Add ORCJITv2 support

apmasell:
This mainly uses LLJIT through the C++ API avaiable in LLVM 14.

Create an API that allows creation of individual JITDylib as garbage-tracked Python objects so that life-cycle management of JITted code can be easily performed in Python.

Based-on: Graham Markall <gmarkall@nvidia.com>

### DIFF
--- a/ffi/orcjit.cpp
+++ b/ffi/orcjit.cpp
@@ -1,0 +1,337 @@
+#include "core.h"
+#include "llvm-c/LLJIT.h"
+#include "llvm-c/Orc.h"
+#include <sstream>
+
+#include "llvm/AsmParser/Parser.h"
+#include "llvm/ExecutionEngine/Orc/Core.h"
+#include "llvm/ExecutionEngine/Orc/ExecutionUtils.h"
+#include "llvm/ExecutionEngine/Orc/LLJIT.h"
+#include "llvm/ExecutionEngine/Orc/RTDyldObjectLinkingLayer.h"
+#include "llvm/ExecutionEngine/SectionMemoryManager.h"
+#include "llvm/IRReader/IRReader.h"
+#include "llvm/Support/MemoryBuffer.h"
+#include "llvm/Support/SourceMgr.h"
+using namespace llvm;
+using namespace llvm::orc;
+
+inline LLJIT *unwrap(LLVMOrcLLJITRef P) { return reinterpret_cast<LLJIT *>(P); }
+
+inline TargetMachine *unwrap(LLVMTargetMachineRef TM) {
+    return reinterpret_cast<TargetMachine *>(TM);
+}
+
+inline LLVMOrcJITTargetMachineBuilderRef wrap(JITTargetMachineBuilder *JTMB) {
+    return reinterpret_cast<LLVMOrcJITTargetMachineBuilderRef>(JTMB);
+}
+
+class JITDylibTracker {
+  public:
+    std::shared_ptr<LLJIT> lljit;
+    IntrusiveRefCntPtr<llvm::orc::ResourceTracker> tracker;
+    JITDylib &dylib;
+    JITDylibTracker(std::shared_ptr<LLJIT> &lljit_, JITDylib &dylib_,
+                    IntrusiveRefCntPtr<llvm::orc::ResourceTracker> &&tracker_)
+        : lljit(lljit_), dylib(dylib_), tracker(tracker_) {}
+};
+
+typedef struct {
+    uint8_t element_kind;
+    char *value;
+    size_t value_len;
+} LinkElement;
+typedef struct {
+    char *name;
+    uint64_t address;
+} SymbolAddress;
+extern "C" {
+
+API_EXPORT(std::shared_ptr<LLJIT> *)
+LLVMPY_CreateLLJITCompiler(LLVMTargetMachineRef tm, const char **OutError) {
+    LLJITBuilder builder;
+
+    if (tm) {
+        // The following is based on
+        // LLVMOrcJITTargetMachineBuilderCreateFromTargetMachine. However,
+        // we can't use that directly because it destroys the target
+        // machine, but we need to keep it alive because it is referenced by
+        // / shared with other objects on the Python side.
+        auto *template_tm = unwrap(tm);
+
+        builder.setJITTargetMachineBuilder(
+            JITTargetMachineBuilder(template_tm->getTargetTriple())
+                .setCPU(template_tm->getTargetCPU().str())
+                .setRelocationModel(template_tm->getRelocationModel())
+                .setCodeModel(template_tm->getCodeModel())
+                .setCodeGenOptLevel(template_tm->getOptLevel())
+                .setFeatures(template_tm->getTargetFeatureString())
+                .setOptions(template_tm->Options));
+    }
+    builder.setObjectLinkingLayerCreator(
+        [&](llvm::orc::ExecutionSession &session, const llvm::Triple &triple)
+            -> std::unique_ptr<llvm::orc::ObjectLayer> {
+            auto linkingLayer =
+                std::make_unique<llvm::orc::RTDyldObjectLinkingLayer>(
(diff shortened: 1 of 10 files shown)


ffi/orcjit.cpp
```diff
@@ -0,0 +1,337 @@
...
+            return linkingLayer;
+        });
+
+    auto jit = builder.create();
+
+    if (!jit) {
+        char *message = LLVMGetErrorMessage(wrap(jit.takeError()));
+        *OutError = LLVMPY_CreateString(message);
+        LLVMDisposeErrorMessage(message);
+        return nullptr;
```
sklam, line 94: Can we get to `llvm::orc::ExecutionSession::setErrorReporter` and define a reporter that doesn't print to stderr?

ffi/orcjit.cpp
```diff
@@ -0,0 +1,337 @@
...
+                .setCodeModel(template_tm->getCodeModel())
+                .setCodeGenOptLevel(template_tm->getOptLevel())
+                .setFeatures(template_tm->getTargetFeatureString())
+                .setOptions(template_tm->Options));
+    }
+    builder.setObjectLinkingLayerCreator(
+        [&](llvm::orc::ExecutionSession &session, const llvm::Triple &triple)
+            -> std::unique_ptr<llvm::orc::ObjectLayer> {
+            auto linkingLayer =
+                std::make_unique<llvm::orc::RTDyldObjectLinkingLayer>(
```
sklam, line 74: This is not using JITLink, which we will want an option to use.

See https://github.com/JuliaLang/julia/pull/43664/files#diff-b858d0b6a127703fdaa4091f393fff5033565d010102e00b4426363ee7dfa91dR15-R17

ffi/orcjit.cpp
```diff
@@ -0,0 +1,337 @@
...
+        auto *template_tm = unwrap(tm);
+
+        builder.setJITTargetMachineBuilder(
+            JITTargetMachineBuilder(template_tm->getTargetTriple())
+                .setCPU(template_tm->getTargetCPU().str())
+                .setRelocationModel(template_tm->getRelocationModel())
+                .setCodeModel(template_tm->getCodeModel())
+                .setCodeGenOptLevel(template_tm->getOptLevel())
+                .setFeatures(template_tm->getTargetFeatureString())
+                .setOptions(template_tm->Options));
```
sklam, line 68: Numba might need to override these settings. 
- relocmodel and codemodel might need to be change because the default might not always work with JIT.
- optlevel depends on user selected opt-level which can be changed for debugging.
- target features are changed to disable certain CPU features; useful for making generic reusable code cache.

llvmlite's TargetMachine class does not expose these options yet. We might want to add the setters there or just let user pass these settings directly as argument in the current function.

sklam (commented):
The following review is from my first-pass look on the C++ code. It's just a fast pass over and I reported anything that I think might be a problem for Numba. I will need to do another pass of review with a fresh mind later.

ffi/orcjit.cpp
apmasell: Looking at `LLVMPY_CreateTargetMachine` already allows setting most of these and that's bound through `Target.create_target_machine`. It doesn't allow setting all `TargetOptions`, but I think we should extend that as required. I don't see the need for a separate API here.

llvmlite/binding/orcjit.py
```diff
@@ -0,0 +1,329 @@
...
+
+    def add_ir(self, llvmir):
+        """
+        Adds a compilation unit to the library using LLVM IR as the input
+        format.
+
+        This takes a string or an object that can be converted to a string,
+        including IRBuilder, that contains LLVM IR.
+        """
+        self.__entries.append((0, str(llvmir).encode('utf-8')))
```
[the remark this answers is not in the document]
apmasell: Why? It doesn't ensure that the enum values are the same between the Python and C++ code.

ffi/orcjit.cpp
apmasell: JITLink makes enabling debugging very difficult. I haven't been able to get it to work on Linux. Julia's solution to the debugging is to do it manually, and they don't support MacOS. It seems like this gets easier in future LLVM version as there's just a Boolean flag to enable debugging.

llvmlite/tests/test_binding.py
```diff
@@ -1264,7 +1274,7 @@ def test_lookup_current_process_symbol_fails(self):
         # enabled searching the current process for symbols.
         msg = 'Failed to materialize symbols:.*getversion'
         with self.assertRaisesRegex(RuntimeError, msg):
-            self.jit(asm_getversion, "getversion")
+            self.jit(asm_getversion, "getversion", suppress_errors=True)
```
sklam, line 1277: Q: Why doesn't all raised error print to stderr? I thought we'll need `supress_errors=True` everywhere.

llvmlite/binding/orcjit.py
```diff
@@ -272,12 +272,17 @@ def _dispose(self):
         self._capi.LLVMPY_LLJITDispose(self)
 
 
-def create_lljit_compiler(target_machine=None):
+def create_lljit_compiler(target_machine=None,
+                          use_jit_link=False,
+                          suppress_errors=False):
```
sklam, line 277: Probably better to make `use_jit_link` and `suppress_errors` keyword-only arguments to anticipate for future changes. e.g. if we drop `use_jit_link` in the future, `create_lljit_compiler(tm, True)` would mean something else; but the keyword-only will force user to write `create_lljit_compiler(tm, use_jit_link=True)` and the API changes will not be silent.
```suggestion
def create_lljit_compiler(target_machine=None, *,
                          use_jit_link=False,
                          suppress_errors=False):
```

ffi/orcjit.cpp
sklam: ok

ffi/orcjit.cpp
sklam: done in https://github.com/numba/llvmlite/pull/942/commits/d7a0d6c1759a6dce021b1332afcfa62565bdce4a

llvmlite/tests/test_binding.py
apmasell: LLVM's error handling is now designed like this: there are two important classes `Error` and `Expected<T>`.  `Expected<T>` is basically Rust's `Result<T, Error>`. Some LLVM functions return these values and any that I get I turn into Python exceptions. Internally, LLVM doesn't always propagate them though. For instance, the object layer initializer has to return `std::unique_ptr<ObjectLayer>`, so you have to eat the errors. If you let an error be destroyed without "checking" it (printing it somewhere), then it will trigger a panic. There's `cantFail` that does the equivalent of Rust's `Result::unwrap`. It's also possible to push an error into an `ExecutionSession` and let it propagate through an error handler there. That's what I'm doing with `suppress_errors`; it's hiding errors that have been pushed into the `ExecutionSession` rather than printing them to console. Only the errors I can't explicitly handle end up there though. This only needs to be set at the time the execution session is created.

ffi/orcjit.cpp
```diff
@@ -80,10 +119,9 @@ LLVMPY_CreateLLJITCompiler(LLVMTargetMachineRef tm, const char **OutError) {
                     true);
                 linkingLayer->setAutoClaimResponsibilityForObjectSymbols(true);
             }
-            linkingLayer->registerJITEventListener(
-                *llvm::JITEventListener::createGDBRegistrationListener());
```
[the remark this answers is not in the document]
apmasell: Fixed.

llvmlite/binding/orcjit.py
apmasell: Fixed.

sklam (approved):
This is the first of the OrcJIT PR and their will be subsequent PR for it once we started to use it in Numba. The API is not yet documented in this PR since we may need to change it.

PR #950, opened 2023-05-17
Fix incorrect `byval` and other attributes on LLVM 14

### LINKED ISSUE #946: Wrong syntax for LLVM 14 argument attributes
As of 3c2163ada3c6ce1acf78e6c06e8a5e38c5c24ced , llvmlite still outputs wrong syntax for LLVM 14. 

At least for `byval`, `sret` and `inalloca`  (others can be affected as well, but I couldn't find difference between LLVM 11 and LLVM 14 regarding attributes listed in https://github.com/numba/llvmlite/blob/3c2163ada3c6ce1acf78e6c06e8a5e38c5c24ced/llvmlite/ir/values.py#L1021), llvmlite outputs syntax for LLVM 11, which actually has been changed in LLVM 14. 
Please check language reference https://releases.llvm.org/14.0.0/docs/LangRef.html. 
In example below I get such error:
```
<string>:5:30: error: expected '('
define i32 @"sum"(i32* byval %".1", i32 %".2")
                             ^
```
And LLVM actually expects it to be `define i32 @"sum"(i32* byval({i32}) %".1", i32 %".2")`

Take this example (I took it from examples/sum.py and just added attributes):
```py
from __future__ import print_function

from ctypes import CFUNCTYPE, c_int, POINTER
import sys
try:
    from time import perf_counter as time
except ImportError:
    from time import time

import numpy as np

try:
    import faulthandler; faulthandler.enable()
except ImportError:
    pass

import llvmlite.ir as ll
import llvmlite.binding as llvm


llvm.initialize()
llvm.initialize_native_target()
llvm.initialize_native_asmprinter()


t1 = time()

fnty = ll.FunctionType(ll.IntType(32), [ll.IntType(32).as_pointer(),
                                        ll.IntType(32)])
module = ll.Module()

func = ll.Function(module, fnty, name="sum")

bb_entry = func.append_basic_block()
bb_loop = func.append_basic_block()
bb_exit = func.append_basic_block()

builder = ll.IRBuilder()
builder.position_at_end(bb_entry)

builder.branch(bb_loop)
builder.position_at_end(bb_loop)

index = builder.phi(ll.IntType(32))
index.add_incoming(ll.Constant(index.type, 0), bb_entry)
accum = builder.phi(ll.IntType(32))
accum.add_incoming(ll.Constant(accum.type, 0), bb_entry)

#func.args[0].add_attribute('sret')
func.args[0].add_attribute('byval')
ptr = builder.gep(func.args[0], [index])
value = builder.load(ptr)

added = builder.add(accum, value)
accum.add_incoming(added, bb_loop)

indexp1 = builder.add(index, ll.Constant(index.type, 1))
index.add_incoming(indexp1, bb_loop)

cond = builder.icmp_unsigned('<', indexp1, func.args[1])
builder.cbranch(cond, bb_loop, bb_exit)

builder.position_at_end(bb_exit)
builder.ret(added)

strmod = str(module)

t2 = time()

print("-- generate IR:", t2-t1)

t3 = time()

llmod = llvm.parse_assembly(strmod)

t4 = time()

print("-- parse assembly:", t4-t3)

print(llmod)

pmb = llvm.create_pass_manager_builder()
pmb.opt_level = 2
pm = llvm.create_module_pass_manager()
pmb.populate(pm)

t5 = time()

pm.run(llmod)

t6 = time()

print("-- optimize:", t6-t5)

t7 = time()

target_machine = llvm.Target.from_default_triple().create_target_machine()

with llvm.create_mcjit_compiler(llmod, target_machine) as ee:
    ee.finalize_object()
    cfptr = ee.get_function_address("sum")

    t8 = time()
    print("-- JIT compile:", t8 - t7)

    print(target_machine.emit_assembly(llmod))

    cfunc = CFUNCTYPE(c_int, POINTER(c_int), c_int)(cfptr)
    A = np.arange(10, dtype=np.int32)
    res = cfunc(A.ctypes.data_as(POINTER(c_int)), A.size)

    print(res, A.sum())
```

Affects https://github.com/msys2/MINGW-packages/issues/15467 and https://github.com/m-labs/artiq/issues/2050

apmasell:
Some attributes now need type information. This preserves it.

Closes #946

### DIFF
--- a/llvmlite/ir/values.py
+++ b/llvmlite/ir/values.py
@@ -1020,18 +1020,30 @@ class Function(GlobalValue):
 
 
 class ArgumentAttributes(AttributeSet):
-    _known = MappingProxyType({'byval': True,
-                               'inalloca': False,
-                               'inreg': False,
-                               'nest': False,
-                               'noalias': False,
-                               'nocapture': False,
-                               'nonnull': False,
-                               'returned': False,
-                               'signext': False,
-                               'sret': True,
-                               'zeroext': False
-                               })
+    # List from
+    # https://releases.llvm.org/14.0.0/docs/LangRef.html#parameter-attributes
+    _known = MappingProxyType({
+        'byref': True,
+        'byval': True,
+        'elementtype': True,
+        'immarg': False,
+        'inalloca': True,
+        'inreg': False,
+        'nest': False,
+        'noalias': False,
+        'nocapture': False,
+        'nofree': False,
+        'nonnull': False,
+        'noundef': False,
+        'preallocated': True,
+        'returned': False,
+        'signext': False,
+        'sret': True,
+        'swiftasync': False,
+        'swifterror': False,
+        'swiftself': False,
+        'zeroext': False,
+    })
 
     def __init__(self, args=()):
         self._align = 0
--- a/llvmlite/ir/values.py
+++ b/llvmlite/ir/values.py
@@ -872,7 +872,7 @@ class AttributeSet(set):
         return super(AttributeSet, self).add(name)
 
     def _to_list(self, typ):
-        return list(map(lambda a: self._expand(a, typ), iter(sorted(self))))
+        return [self._expand(i, typ) for i in sorted(self)]
 
 
 class FunctionAttributes(AttributeSet):
--- a/llvmlite/ir/values.py
+++ b/llvmlite/ir/values.py
@@ -858,6 +858,7 @@ class AttributeSet(set):
     _known = ()
 
     def __init__(self, args=()):
+        super().__init__()
         if isinstance(args, str):
             args = [args]
         for name in args:
@@ -869,7 +870,11 @@ class AttributeSet(set):
     def add(self, name):
         if name not in self._known:
             raise ValueError('unknown attr {!r} for {}'.format(name, self))
+        self._validate_add(name)
         return super(AttributeSet, self).add(name)
+
+    def _validate_add(self, name):
+        pass
 
     def _to_list(self, typ):
         return [self._expand(i, typ) for i in sorted(self)]
@@ -1023,26 +1028,30 @@ class ArgumentAttributes(AttributeSet):
     # List from
     # https://releases.llvm.org/14.0.0/docs/LangRef.html#parameter-attributes
     _known = MappingProxyType({
-        'byref': True,
-        'byval': True,
-        'elementtype': True,
-        'immarg': False,
-        'inalloca': True,
-        'inreg': False,
-        'nest': False,
-        'noalias': False,
-        'nocapture': False,
-        'nofree': False,
-        'nonnull': False,
-        'noundef': False,
-        'preallocated': True,
-        'returned': False,
-        'signext': False,
-        'sret': True,
-        'swiftasync': False,
-        'swifterror': False,
-        'swiftself': False,
-        'zeroext': False,
+        # Each tuple is LLVM 11 vs 14 behaviour:
+        # None (unsupported),
+        # True (emit type),
+        # False (emit name only)
+        'byref': (None, True),
+        'byval': (True, True),
+        'elementtype': (None, True),
+        'immarg': (False, False),
+        'inalloca': (False, True),
+        'inreg': (False, False),
+        'nest': (False, False),
+        'noalias': (False, False),
+        'nocapture': (False, False),
+        'nofree': (False, False),
+        'nonnull': (False, False),
+        'noundef': (False, False),
+        'preallocated': (True, True),
+        'returned': (False, False),
+        'signext': (False, False),
+        'sret': (False, True),
+        'swiftasync': (None, False),
+        'swifterror': (False, False),
+        'swiftself': (False, False),
+        'zeroext': (False, False),
     })
 
     def __init__(self, args=()):
@@ -1051,10 +1060,19 @@ class ArgumentAttributes(AttributeSet):
         self._dereferenceable_or_null = 0
         super(ArgumentAttributes, self).__init__(args)
 
+    def _validate_add(self, name):
+        import llvmlite.binding
+        llvm_major = llvmlite.binding.llvm_version_info[0]
+        requires_type = self._known.get(name)[llvm_major >= 14]
+        if requires_type is None:
+            raise ValueError(
+                f"Attribute {name} is not supported on current LLVM version")
+
     def _expand(self, name, typ):
         import llvmlite.binding
-        if (llvmlite.binding.llvm_version_info[0] >= 14 and
-                self._known.get(name)):
+        llvm_major = llvmlite.binding.llvm_version_info[0]
+        requires_type = self._known.get(name)[llvm_major >= 14]
+        if requires_type:
             return f"{name}({typ.pointee})"
         else:
             return name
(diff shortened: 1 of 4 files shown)


esc:
Two high-level questions from my initial look:

* Can the tests be made to invoke LLVM and check against the produced output, in addition to checking for the "canned" result we have now?

* What happens to the additional parameter attributes that were added in LLVM 14 -- like `byref`? Will you add support for these too?

apmasell:
For the first, I don't know how to do that given the way I can get the plumbing to work with pickle. If you have a design, I'm happy to hear it.

For the second, until we have a true way to control LLVM backend version, we should only do that if we raise the minimum LLVM version. Are we prepared to do that in a patch release?

thomasfire:
> For the first, I don't know how to do that given the way I can get the plumbing to work with pickle. If you have a design, I'm happy to hear it.

Is it possible to integrate these pieces of IR from tests to some function and see if LLVM compiles it?

apmasell:
That's what the current test does.

esc:
OK, so just to double check, the way this works now:

a) We have code to add the types to the Parameter Attributes
b) The "canned" tests have been updated, but they only test for the `sret` parameter attribute
c) We have a generated test, where we test `byval` -- such that we generate some IR and then ensure that the given version of LLVM can indeed parse this IR

The question I have, and I have no idea if this even needs to be asked: so it seems that we only check for `sret` and `byval`, but these are not the only parameter attributes? Do we need to -- or should we -- test the other attributes too?

The second question that I have is if we can ask anyone from the `artiq` project to check this PR with their code? This is where these issues were originally detected, and perhaps this is a good test case for this PR: https://github.com/m-labs/artiq/issues/2050 @thomasfire would you be able to facilitate this?

apmasell:
Added.

thomasfire:
I've tested our code against this branch, and would confirm that this fixes the issue

esc:
> I've tested our code against this branch, and would confirm that this fixes the issue

Awesome, thank you for your feedback! 

@thomasfire would you like to be added to our RC-Testers group on our discourse: 

https://numba.discourse.group/

it's a group of folks who are interested in new Numba/llvmlite release candidates, perhaps this could be of interest to you? It would perhaps help with future llvmlite testing?

thomasfire:
> would you like to be added to our RC-Testers group on our discourse:

Yes, you can add me, username is same as github's

llvmlite/ir/values.py
```diff
@@ -863,14 +863,16 @@ def __init__(self, args=()):
...
     def add(self, name):
         if name not in self._known:
             raise ValueError('unknown attr {!r} for {}'.format(name, self))
         return super(AttributeSet, self).add(name)
 
-    def __iter__(self):
-        # In sorted order
-        return iter(sorted(super(AttributeSet, self).__iter__()))
+    def _to_list(self, typ):
+        return list(map(lambda a: self._expand(a, typ), iter(sorted(self))))
```
esc, line 875: Could this be converted to?

```suggestion
        return [self.expand(i, typ) for i in sorted(self)]
```

esc (changes requested):
I have a suggestion to make about using slightly more Pythonic syntax.

esc:
> > would you like to be added to our RC-Testers group on our discourse:
> 
> Yes, you can add me, username is same as github's

You should have received an invite now.

esc:
Since this also supports LLVM 11 -- I was wanting to build this PR and test against that. However, I am having trouble getting a development environment setup. Mostly I am troubled with conda environment conflicts.

apmasell:
I've added a commit that should make it work properly on LLVM 11. Some of the attributes being tested weren't supported on LLVM 11.

esc:
> I've added a commit that should make it work properly on LLVM 11. Some of the attributes being tested weren't supported on LLVM 11.

Thank you, I can confirm that with the commit efb856a the llvmlite test suite passes with LLVM 11.

apmasell:
That's exciting and I desperately hope not related to these changes.

esc:
> That's exciting and I desperately hope not related to these changes.

No, it wasn't, it was a side effect of not having pyyaml installed.

esc:
I am now hitting:

```
.Assertion failed: (isInt<33>(Addend) && "Invalid page reloc value."), function encodeAddend, file /Users/ci/miniconda3-arm64/conda-bld/llvmdev_1643905487494/work/lib/ExecutionEngine/RuntimeDyld/Targets/RuntimeDyldMachOAArch64.h, line 210.
Fatal Python error: Aborted

Current thread 0x0000000104d34580 (most recent call first):
  File "/Users/vhaenel/git/llvmlite/llvmlite/binding/ffi.py", line 152 in __call__
  File "/Users/vhaenel/git/llvmlite/llvmlite/binding/executionengine.py", line 92 in finalize_object
  File "/Users/vhaenel/git/numba/numba/core/codegen.py", line 1060 in wrapper
  File "/Users/vhaenel/git/numba/numba/core/codegen.py", line 999 in _finalize_specific
  File "/Users/vhaenel/git/numba/numba/core/codegen.py", line 797 in _finalize_final_module
```

Which I believe is known issue on Mac ARM64?

esc:
> I am now hitting:
> 
> ```
> .Assertion failed: (isInt<33>(Addend) && "Invalid page reloc value."), function encodeAddend, file /Users/ci/miniconda3-arm64/conda-bld/llvmdev_1643905487494/work/lib/ExecutionEngine/RuntimeDyld/Targets/RuntimeDyldMachOAArch64.h, line 210.
> Fatal Python error: Aborted
> 
> Current thread 0x0000000104d34580 (most recent call first):
>   File "/Users/vhaenel/git/llvmlite/llvmlite/binding/ffi.py", line 152 in __call__
>   File "/Users/vhaenel/git/llvmlite/llvmlite/binding/executionengine.py", line 92 in finalize_object
>   File "/Users/vhaenel/git/numba/numba/core/codegen.py", line 1060 in wrapper
>   File "/Users/vhaenel/git/numba/numba/core/codegen.py", line 999 in _finalize_specific
>   File "/Users/vhaenel/git/numba/numba/core/codegen.py", line 797 in _finalize_final_module
> ```
> 
> Which I believe is known issue on Mac ARM64?

It is this one here: https://github.com/numba/numba/issues/8567

OK, I ran it again and it seems like the issue is transient, so I would say this did indeed pass a spot check, with LLVM 11.

llvmlite/tests/test_ir.py
```diff
@@ -1779,6 +1786,75 @@ def test_fma_mixedtypes(self):
...
+            'noundef',
+            'preallocated',
+            'returned',
+            'signext',
+            'swifterror',
+            'swiftself',
+            'zeroext',
+        ) + supplemental:
+            # If this parses, we emitted the right byval attribute format
+            llvm.parse_assembly(gen_code(attr_name))
```
esc, line 1855: Would it be possible to try to compile this too? Or is the test-code not self-sufficient?

esc (changes requested):
Question: can we compile?

llvmlite/tests/test_ir.py
esc: ok, that makes sense, thank you.

esc:
This is now going in, thank you for the patch!

PR #952, opened 2023-05-19
Fix CI failing due to unsupported target triple on non-x86 platforms.

sklam:
Fix test by only running on x86 platforms.


sklam:
BFID `llvmlite_210`

sklam:
> BFID `llvmlite_210`

passed

gmarkall (approved):

Issue #675, opened 2020-12-18
String caching causes incorrect results with mutable instructions


trbabb:
Certain instructions (like Phi) can be mutated. If they are stringified at any point (including for debugging), subsequent mutations won't be reflected in the SSA, giving incorrect results and very confusing bugs!

It's probably strictly better for those instructions to invalidate their string caches immediately upon mutation.

I have a PR for this: https://github.com/numba/llvmlite/pull/661

esc:
@trbabb thank you for raising this issue.

siboehm:
I just ran into this, here's a self-contained example (modified from `examples/`): 
```python
from llvmlite import ir
from llvmlite.ir.transforms import Visitor

# Create some useful types
double = ir.DoubleType()
fnty = ir.FunctionType(double, (double, double))

# Create an empty module...
module = ir.Module(name=__file__)
# and declare a function named "fpadd" inside it
func = ir.Function(module, fnty, name="fpadd")

# Now implement the function
block = func.append_basic_block(name="entry")
builder = ir.IRBuilder(block)
a, b = func.args
c = ir.Constant(ir.IntType(bits=1), False)
result = builder.select(c, a, b, name="res", flags=["nsz"])
builder.ret(result)


class VisitSelect(Visitor):
    def visit_Instruction(self, instr):
        if instr.opname != "select":
            return

        print("Before:", instr)
        instr.flags = ["fast"]
        print("After: ", instr)


# Print the module IR
before = str(module)
VisitSelect().visit(module)
after = str(module)


print("\nBefore:", before)
print("\nAfter: ", after)
```

This prints
```
Before: %"res" = select nsz i1 false, double %".1", double %".2"
After:  %"res" = select nsz i1 false, double %".1", double %".2"
```
But should print
```
Before: %"res" = select nsz i1 false, double %".1", double %".2"
After:  %"res" = select fast i1 false, double %".1", double %".2"
```

This is also due to the string caching and makes for really hard to find bugs (took me 2h of head scratching). Numba's fastmath pass seems to work correctly, presumably because there's no printing before the pass, but that seems really brittle. I'll give the #661 PR a shot, maybe I can get it updated & merged.

siboehm:
(running note) I'm trying to come up with a way to fix the fast-math flag problem. It seems hard to do without breaking changes.

Current ideas:
## Make `flags` a tuple (non-modifable).
- Add a setter, which clears the string cache. 
[...]

PR #954, opened 2023-05-30
Invalidate string cache for mutable instructions

siboehm:
This is a re-do of #661 and adds testing of all modifications. It addresses: Phi nodes, call, cbranch, indirect branch, switch, landing pad.

As mentioned in #675, there is also the issue of fastmath flags on floating point instructions. I'm thinking about how to do those next and will open a separate PR.


siboehm:
@kc611 I fixed the flake8 errors, should be good for review now

siboehm:
@kc611 can you mark this as ready for review?

Issue #956, opened 2023-06-05
docs: README indicates 0.39 supports LLVM 14, but `ffi/build.py` requires LLVM 11


wdconinc:
The README includes in the [compatibility table](https://github.com/numba/llvmlite#compatibility) the following:

llvmlite versions | compatible LLVM versions
-- | --
0.41.0 - ... | 14.x.x
0.37.0 - 0.40.0 | 11.x.x-14.x.x

But `ffi/build.py` [requires LLVM 11](https://github.com/numba/llvmlite/commit/7b977a6ff830356dba253ba9b18a070f79df75b7#diff-ec77745671d702188f8a6e143c2f15e641d1d3681d377241d28250412ea135d5L168) through 0.39. Only with 0.40 does it seem that building with LLVM 14 became possible.

Should the table indicate the following?

llvmlite versions | compatible LLVM versions
-- | --
0.41.0 - ... | 14.x.x
0.40.0 | 11.x.x-14.x.x
0.37.0 - 0.39.1 | 11.x.x

esc:
@wdconinc thank you for the report. We will discuss this today in the developer meeting.

esc:
@wdconinc can you take a look at: https://github.com/numba/llvmlite/pull/958 -- should fix the issue.

PR #958, opened 2023-06-06
fixup LLVM versions in version compat table

esc:
As title

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,7 +68,8 @@ Historical compatibility table:
 llvmlite versions  compatible LLVM versions
 =================  ========================
 0.41.0 - ...       14.x.x
-0.37.0 - 0.40.0    11.x.x-14.x.x
+0.40.0 - 0.40.1    11.x.x and 14.x.x (12.x.x and 13.x.x untested but may work)
+0.37.0 - 0.39.1    11.x.x
 0.34.0 - 0.36.0    10.0.x (9.0.x for  ``aarch64`` only)
 0.33.0             9.0.x
 0.29.0 - 0.32.0    7.0.x, 7.1.x, 8.0.x


esc:
Fixes: https://github.com/numba/llvmlite/issues/956

esc:
> Thanks for the quick fix! I think this clarifies it well (including the gap for LLVM 12 and 13).

I am still a bit unsure about 0.41.0 -- let's wait for an answer on that.

esc:
So, the state is: 0.41.0 does technically support LLVM 11 -- but this is not intended. This means, before this can be merged, we need to remove the LLVM 11 specific code-paths before we can finish this PR and get it merged.

apmasell:
Removal is in #959

apmasell (approved):

YoushaaMurhij:
I think the table should be updated. v0.41.0 does not support LLVM 14.0.1 on aarch64 (but maybe later 14.x.x releases work!)

sklam:
@YoushaaMurhij, is there a particular issue that you are referring to that cause llvm 14.0.1 to not work?

YoushaaMurhij:
> @YoushaaMurhij, is there a particular issue that you are referring to that cause llvm 14.0.1 to not work?

Sorry for not providing enough details. I referred to the table to install `llvmlite v0.41.0` on aarch64 by building `llvm 14.0.1` first.  but, `v0.41.0` was not found using `pip install llvmlite==0.41.0` and I downgraded to `v0.40.1`.

sklam:
ah, i see. `v0.41` is the in-development version, files on the `main` branch reflects information for that. While we don't publish `dev` version of the package, you can get it via conda with `conda install numba/label/dev::llvmlite`.

YoushaaMurhij:
I understand. Thanks for your response.

PR #961, opened 2023-06-15
Add a "bridge" to run separately compiled passes

apmasell:
Add C++ support to invoke a pass manager using a different LLVM instance (either linked into the same binary or separate module). This allows adding passes not compiled at runtime against an llvmlite that has been statically linked against LLVM and stripped.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,5 +2,5 @@ include *.rst setup.py runtests.py run_coverage.py LICENSE
 include versioneer.py
 include llvmlite/_version.py
 recursive-include examples *.py *.ll
-recursive-include ffi *.txt *.h *.cpp Makefile.* *.py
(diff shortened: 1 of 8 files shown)


DrTodd13:
Thanks!  I'll take a look.

DrTodd13:
@apmasell I think I've changed my mind a bit on this and given that we are wanting an interface with llvm IR in and llvm IR out but no requirement that the process to do the transformation is LLVM code at all (obviously the code has to understand LLVM IR but doesn't strictly have to link to LLVM code) that perhaps this should be done on the Numba side as part of the Numba compiler pass machinery.  Our additional pass happens at the very beginning so should be a little cleaner for us to add that to the Numba pipeline than to use your mechanism in llvmlite.  I still think this is good work and could be useful to someone but just not perhaps in our use case.

apmasell:
Sounds good to me.

PR #966, opened 2023-06-24
build: support building on GNU/Hurd

pinotree:
LLVM is ported on the Hurd, and the Hurd has the same GNU toolchain that
is also generally used on Linux. Since build & the tests work fine,
apply/use some Linux build tweaks also on any GNU platform (including
the Hurd):
- the Linux `Makefile`
- `-fPIC` in `CXXFLAGS`

### DIFF
--- a/ffi/build.py
+++ b/ffi/build.py
@@ -213,7 +213,8 @@ def main_posix(kind, library_ext):
 def main():
     if sys.platform == 'win32':
         main_windows()
-    elif sys.platform.startswith('linux'):
+    elif sys.platform.startswith(('linux', 'gnu')):
+        # Linux and GNU-based OSes (e.g. GNU/Hurd), using the same Makefile
         main_posix('linux', '.so')
     elif sys.platform.startswith(('freebsd','openbsd')):
         main_posix('freebsd', '.so')
--- a/setup.py
+++ b/setup.py
@@ -64,9 +64,9 @@ build_ext = cmdclass.get('build_ext', build_ext)
 
 def build_library_files(dry_run):
     cmd = [sys.executable, os.path.join(here_dir, 'ffi', 'build.py')]
-    # Turn on -fPIC for building on Linux, BSD, and OS X
+    # Turn on -fPIC for building on Linux, BSD, OS X, and GNU platforms
     plt = sys.platform
-    if 'linux' in plt or 'bsd' in plt or 'darwin' in plt:
+    if 'linux' in plt or 'bsd' in plt or 'darwin' in plt or 'gnu' in plt:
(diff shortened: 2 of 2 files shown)


ffi/build.py
```diff
@@ -213,7 +213,7 @@ def main_posix(kind, library_ext):
 def main():
     if sys.platform == 'win32':
         main_windows()
-    elif sys.platform.startswith('linux'):
+    elif sys.platform.startswith(('linux', 'gnu')):
```
gmarkall, line 216: If I encountered this whilst reading the code of `build.py` outside of the context of this PR, I'd not know what `'gnu'` means here (apologies for the harsh-sounding comment, but unfortunately it is the case that GNU Hurd has a relatively small userbase and there isn't broad awareness of the platform). Could you add a comment that notes what's going on here, e.g.:

```suggestion
    elif sys.platform.startswith(('linux', 'gnu')):
        # Linux and GNU Hurd both use the Linux Makefile
```

gmarkall (changes requested):
Many thanks for the PR! I have added a small note on the diff.

Generally in the past our policy has been to accept changes that help llvmlite build on other platforms and maintain them on a best-effort basis, even if we don't support them (as you will have seen FreeBSD and OpenBSD support below your changes). My expectation is that this GNU Hurd build support should be acceptable in a similar vein. However, I'll bring this up at the triage meeting tomorrow, and let you know if there's any further feedback or a broad consensus on the idea of GNU Hurd build support.

ffi/build.py
pinotree: makes sense, done

pinotree:
> Generally in the past our policy has been to accept changes that help llvmlite build on other platforms and maintain them on a best-effort basis, even if we don't support them (as you will have seen FreeBSD and OpenBSD support below your changes). My expectation is that this GNU Hurd build support should be acceptable in a similar vein.

Sounds good to me. llvmlite is available in Debian [1], and one of the ports architectures is hurd-i386 (i.e. GNU/Hurd on x86); you can see here [2] the build status, including the logs for old versions of each architecture (see the "old" links).

I did not see a lot of OS-specific code in llvmlite, so I think it will not break on the Hurd that often.

[1] https://tracker.debian.org/pkg/llvmlite
[2] https://buildd.debian.org/status/package.php?p=llvmlite&suite=sid

gmarkall (approved):
Many thanks for the additional info - I agree with your assessment that llvmlite is likely to generally be OK on Hurd. Thanks also for updating the PR, and the additional change in `setup.py`.

From my perspective this looks good, but I'm going to bring it up in the triage meeting today before marking as "Ready to merge" to allow other maintainers the opportunity to comment.

esc:
@pinotree thank you for that patch! @gmarkall thank you for the review!

PR #973, opened 2023-08-05
Bindings: Query incoming blocks of a phi instruction

tbennun:
When iterating over the operands of a phi instruction, the incoming basic blocks are missing and cannot be queried by any means. This PR adds them to the bindings.


gmarkall:
Many thanks for the PR - I'll mark as ready for review.

sklam (approved):
LGTM

PR #985, opened 2023-08-22
Update release checklist post 0.41.0rc1

esc:
As title


kc611 (approved):
LGTM! The noticeable change in release process structure suggested in this PR is the change in timing of change-log backporting to `main` which has been shifted from doing it at every RCs to doing it once at Final release.

PR #996, opened 2023-10-10
fix typos found by codespell

esc:
As title

### DIFF
--- a/docs/source/user-guide/ir/ir-builder.rst
+++ b/docs/source/user-guide/ir/ir-builder.rst
@@ -530,7 +530,7 @@ Function call
        optimized, but may be ignored.
      * ``"musttail"``: indicates that the call must be tail-call optimized for
        program correctness.
-     * ``"notail"``: indicate thats the call should never be tail-call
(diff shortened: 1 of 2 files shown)


docs/source/user-guide/ir/ir-builder.rst
```diff
@@ -530,7 +530,7 @@ Function call
        optimized, but may be ignored.
      * ``"musttail"``: indicates that the call must be tail-call optimized for
        program correctness.
-     * ``"notail"``: indicate thats the call should never be tail-call
+     * ``"notail"``: indicate that the call should never be tail-call
```
gmarkall, line 533: The grammar is still incorrect:

```suggestion
     * ``"notail"``: indicates that the call should never be tail-call
```

gmarkall (changes requested):
The spellings are now correct but the sentences still don't read well - suggestions on the diff.

esc:
> The spellings are now correct but the sentences still don't read well - suggestions on the diff.

Thank you! I accepted the suggestions.

PR #1005, opened 2023-10-19
Remove suggestion that add_global_mapping() is unused

gmarkall:
It is used in various places in Numba. The comment here forces the reader to do a little pointless investigation, so remove it.


guilhermeleobas (approved):

stuartarchibald (approved):
Thanks for the patch. Indeed, it is used in Numba!

esc:
@stuartarchibald I think this is in the wrong milestone. 0.41.1 was already released on 17th Oct, so punting this to 0.42.0.

stuartarchibald:
Ah yes, thanks @esc.

PR #1016, opened 2023-11-29
Add convergent as a supported FunctionAttribute and CallInstrAttribute.

diptorupd:
The PR adds a new attribute string to the set of known `FunctionAttributes` and `CallInstrAttributes`.

The `convergent` attribute is needed for certain function declarations and call instructions to generate correct SPMD code for OpenCL. Refer: https://llvm.org/docs/ConvergentOperations.html

### DIFF
--- a/llvmlite/tests/test_ir.py
+++ b/llvmlite/tests/test_ir.py
@@ -145,6 +145,22 @@ class TestFunction(TestBase):
         # Check pickling
         self.assert_pickle_correctly(func)
 
+    def test_declare_convergent_attribute(self):
+        # Now with function attributes
+        func = self.function()
+        func.attributes.add("convergent")
+        func.attributes.alignstack = 16
+        tp_pers = ir.FunctionType(int8, (), var_arg=True)
+        pers = ir.Function(self.module(), tp_pers, '__gxx_personality_v0')
+        func.attributes.personality = pers
+        asm = self.descr(func).strip()
+        self.assertEqual(asm,
+                         ("declare %s convergent alignstack(16) "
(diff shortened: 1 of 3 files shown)


stuartarchibald (commented):
Thanks for the patch. This seems like an appropriate addition to support IR generation for convergent operation semantics. It would be good to add a check for IR generation into a suitable existing tests in `llvmlite/tests/test_ir.py`, thank you.

diptorupd:
> Thanks for the patch. This seems like an appropriate addition to support IR generation for convergent operation semantics. It would be good to add a check for IR generation into a suitable existing tests in `llvmlite/tests/test_ir.py`, thank you.

@stuartarchibald I have added some basic unit tests. Please let me know if these are adequate.

llvmlite/tests/test_ir.py
```diff
@@ -145,6 +145,22 @@ def test_declare_attributes(self):
         # Check pickling
         self.assert_pickle_correctly(func)
 
+    def test_declare_convergent_attribute(self):
+        # Now with function attributes
+        func = self.function()
+        func.attributes.add("convergent")
```
stuartarchibald, line 151: To save repeating test code, could this line perhaps just be added into the above test? I think all that needs checking is that `convergent` wires through as an attribute, regardless of whether it makes sense for a `__gxx_personality` ?

stuartarchibald (commented):
Thanks for adding the tests @diptorupd, one minor comment to resolve else looks good.

diptorupd:
Closing this PR in favour of #1019

diptorupd:
Closing this PR in favour of https://github.com/numba/llvmlite/pull/1019

PR #1021, opened 2023-12-13
Add release notes for 0.42.0rc1

gmarkall:
Generated with:

```
python ../numba/maint/gitlog2changelog.py --token="<omitted>" --beginning="v0.41.0" --repo=numba/llvmlite --digits=4
```

and

```
python ../numba/maint/gitlog2changelog.py --token="<omitted>" --beginning="v0.41.0" --repo=numba/llvmlite --digits=3
```

run in the root of the llvmlite repo, checked out next to the Numba repo, and then merging the results, and eliminating duplicate entries that already appeared in the notes for 0.41.0 and 0.41.1

### DIFF
--- a/CHANGE_LOG
+++ b/CHANGE_LOG
@@ -1,3 +1,41 @@
+v0.42.0rc1 (Dec 13, 2023)
+-------------------------
+
+Highlights of this release include:
+
+- Support for Python 3.12.
+- A fix for relocation overflows on AArch64 systems.
+- Binding layer: new queries for incoming blocks of phi instructions, type
+  kinds, and elements. Addition of the Instruction Namer pass.
+- IR layer: Support `convergent` a an attribute of function calls and call
+  instructions.
+
+Pull-Requests:
+
+* PR `#973 <https://github.com/numba/llvmlite/pull/973>`_: Bindings: Query incoming blocks of a phi instruction (`tbennun <https://github.com/tbennun>`_)
+* PR `#978 <https://github.com/numba/llvmlite/pull/978>`_: Bindings: Query type kinds, derived types, and elements (`tbennun <https://github.com/tbennun>`_ `sklam <https://github.com/sklam>`_)
+* PR `#981 <https://github.com/numba/llvmlite/pull/981>`_: Add Instruction Namer pass to PassManager (`tbennun <https://github.com/tbennun>`_)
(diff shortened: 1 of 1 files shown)


CHANGE_LOG
```diff
@@ -1,3 +1,41 @@
+v0.42.0rc1 (Dec 13, 2023)
+-------------------------
+
+Highlights of this release include:
+
+- Support for Python 3.12.
+- A fix for relocation overflows on AArch64 systems.
+- Binding layer: new queries for incoming blocks of phi instructions, type
+  kinds, and elements. Addition of the Instruction Namer pass.
+- IR layer: Support `convergent` a an attribute of function calls and call
```
stuartarchibald, line 10:
```suggestion
- IR layer: Support `convergent` as an attribute of function calls and call
```

CHANGE_LOG
stuartarchibald, line 1: I think a mistake was made in v0.41.1, #985 targets main and not the release branch, so 0.42.x will be the release series containing this change?

stuartarchibald (commented):
Thanks for making the release notes @gmarkall. There's a typo to fix, and a query about which branch #985 was merged into. It's really hard to manually validate these lists, it'd be good to do some work on the generation script in the future to make it so that it de-duplicates etc and "just works", will open an issue in Numba tracker as that's where the script resides.

CHANGE_LOG
gmarkall: It looks like that was between 0.41.0rc and 0.41.0 - so it appears here: https://github.com/numba/llvmlite/pull/1021/files#diff-0f9dd0b75ca4dd0210c8be84066de570e59c0713d22688ac5b26422af8b05021R84

stuartarchibald: hmmm, but it's not actually on the `release0.41` branch ?

gmarkall: Do you want a change making? If so, what should be changed?

stuartarchibald: I think #985 might need moving from 0.41.x to 0.42, if my observation about where it was merged is correct. However, this is a) quite minor and b) has no impact on the software as the patch updates release checklists, perhaps leaving it would also be fine. The take away is that this process is hard to verify, as noted above!

gmarkall: ok - if you're OK leaving it, let's leave it - it seems completely irrelevant to the actual source and binaries that get shipped, in addition to being quite minor.

stuartarchibald: I agree. Let's leave it.

sklam (approved):

PR #1025, opened 2023-12-23
skip `raise` basic blocks in `verifyFanoutBackward`

### LINKED ISSUE #1023: `fanout_raise` in `refpruning` algorithm doesn't work on this subtle case
Continue from https://github.com/numba/numba/issues/9213
Tested with 0.39.1. I feel latest version still has this issue.

I found a subtle case for refcounting, which our refpruning algorithm under `ffi` can't prune it.
However, it should be in theory, since I think any path reaching `raise` block can be ignored?

image

Just want to make sure that you also think this's prunable. 
I do have a real case having this kind of pattern and have a workaround for this. Removing these refcounting calls can boost 5%~10% performance for my benchmark.

cc @sklam

dlee992:
I provide a potential fix for #1023 .
But I can't prove it in theory, maybe you can give a counterexample against this PR.
Will provide a test case for this.

fix #1023

### DIFF
--- a/ffi/custom_passes.cpp
+++ b/ffi/custom_passes.cpp
@@ -400,7 +400,7 @@ struct RefPrunePass : public FunctionPass {
                     SmallBBSet tail_nodes;
                     tail_nodes.insert(decref->getParent());
                     if (!verifyFanoutBackward(incref, incref->getParent(),
-                                              &tail_nodes))
+                                              &tail_nodes, false))
                         continue;
 
                     // scan the CFG between the incref and decref BBs, if
@@ -671,10 +671,12 @@ struct RefPrunePass : public FunctionPass {
                 for (BasicBlock *bb : *decref_blocks) {
                     raising_blocks.insert(bb);
                 }
-                if (verifyFanoutBackward(incref, head_node, p_raising_blocks))
+                if (verifyFanoutBackward(incref, head_node, p_raising_blocks,
+                                         prune_raise_exit))
                     return true;
 
-            } else if (verifyFanoutBackward(incref, head_node, decref_blocks)) {
+            } else if (verifyFanoutBackward(incref, head_node, decref_blocks,
+                                            prune_raise_exit)) {
                 return true;
             }
         }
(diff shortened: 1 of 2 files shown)


llvmlite/tests/test_refprune.py
```diff
@@ -521,6 +521,37 @@ def test_fanout_raise_5(self):
         mod, stats = self.check(self.fanout_raise_5)
         self.assertEqual(stats.fanout_raise, 2)
 
+    # test case 6 is from https://github.com/numba/llvmlite/issues/1023
```
gmarkall, line 524: There's no IR or even Python source anywhere in https://github.com/numba/llvmlite/issues/1023 - perhaps this link should be pointing somewhere else?

gmarkall (commented):
I rebuilt llvmlite locally with this patch, and the Numba test suite passed for me.

llvmlite/tests/test_refprune.py
dlee992: Emm, I mean this test case is from the graph in #1023.

dlee992:
> I rebuilt llvmlite locally with this patch, and the Numba test suite passed for me.

Thanks, graham! This can be marked as `ready for review` I think.

ffi/custom_passes.cpp
```diff
@@ -864,6 +887,10 @@ struct RefPrunePass : public FunctionPass {
             while (workstack.size() > 0) {
                 // Get a basic block
                 BasicBlock *cur_node = workstack.pop_back_val();
+                // If cur_node is a raising block, then skip it
+                if (isRaising(cur_node)) {
+                    continue;
+                }
```
sklam, line 893: This needs to respect the `prune_raise_exit` flag from `findFanout()`.

llvmlite/tests/test_refprune.py
sklam: Is this pattern generated by the reproducer in https://github.com/numba/numba/issues/9213#issue-1911942120? I don't see this patch to have any effect to that.

sklam (commented):
I will need a second pass to think through this change.

dlee992:
RE: https://github.com/numba/llvmlite/pull/1025#discussion_r1458090545
No, this pattern comes from #1023. I didn't share the real code for this, since it's too big, and don't how to get an example from it.
But I think this patch is kind of general.

ffi/custom_passes.cpp
dlee992: addressed this suggestion. Now check `prune_raise_exit` as well.

dlee992:
Has a format issue. Will install pre-commit to solve it

dlee992:
fixed the C++ format issue.

dlee992:
Gentle cc @sklam , ready for review again.

dlee992:
And how about put this small fix into next release?

dlee992:
Can this be marked with next release milestone?

sklam (approved):
Passed the numba testsuite.

PR #1035, opened 2024-03-13
 Support building against llvm15

yashssh:
Add support for llvmlite to be built against llvm15 while keeping support for llvm14 as well.

    1) Drop 2 passes no longer supported by old PM
    2) Update domTree passes to their wrapper passes
    3) Update getAddress() api
    4) Disable opaque pointer mode
    5) Give warning when build with llvm15
    
All the changes are put inside llvm version check guard so that it builds with both llvm14 and llvm15

### DIFF
--- a/ffi/Makefile.linux
+++ b/ffi/Makefile.linux
@@ -2,7 +2,7 @@
 CXX ?= g++
 
 # -flto and --exclude-libs allow us to remove those parts of LLVM we don't use
-CXX_FLTO_FLAGS ?= -flto
+#CXX_FLTO_FLAGS ?= -flto
 LD_FLTO_FLAGS ?= -flto -Wl,--exclude-libs=ALL
 # -fPIC is required when compiling objects for a shared library
 CXX_FPIC_FLAGS ?= -fPIC
--- a/ffi/Makefile.linux
+++ b/ffi/Makefile.linux
@@ -2,7 +2,7 @@
 CXX ?= g++
 
 # -flto and --exclude-libs allow us to remove those parts of LLVM we don't use
-#CXX_FLTO_FLAGS ?= -flto
+CXX_FLTO_FLAGS ?= -flto
 LD_FLTO_FLAGS ?= -flto -Wl,--exclude-libs=ALL
 # -fPIC is required when compiling objects for a shared library
 CXX_FPIC_FLAGS ?= -fPIC
--- a/ffi/core.cpp
+++ b/ffi/core.cpp
@@ -21,10 +21,23 @@ API_EXPORT(void)
 LLVMPY_DisposeString(const char *msg) { free(const_cast<char *>(msg)); }
 
 API_EXPORT(LLVMContextRef)
-LLVMPY_GetGlobalContext() { return LLVMGetGlobalContext(); }
+LLVMPY_GetGlobalContext() {
+  auto context = LLVMGetGlobalContext();
+#if LLVM_VERSION_MAJOR > 14
+  LLVMContextSetOpaquePointers(context, false);
+#endif
+  return context;
+}
+
 
 API_EXPORT(LLVMContextRef)
-LLVMPY_ContextCreate() { return LLVMContextCreate(); }
+LLVMPY_ContextCreate() { 
+    LLVMContextRef context = LLVMContextCreate();
+#if LLVM_VERSION_MAJOR > 14
+    LLVMContextSetOpaquePointers(context, false);
+#endif
+    return context;
+}
 
 API_EXPORT(void)
 LLVMPY_ContextDispose(LLVMContextRef context) {
--- a/ffi/core.cpp
+++ b/ffi/core.cpp
@@ -22,16 +22,15 @@ LLVMPY_DisposeString(const char *msg) { free(const_cast<char *>(msg)); }
 
 API_EXPORT(LLVMContextRef)
 LLVMPY_GetGlobalContext() {
-  auto context = LLVMGetGlobalContext();
+    auto context = LLVMGetGlobalContext();
 #if LLVM_VERSION_MAJOR > 14
-  LLVMContextSetOpaquePointers(context, false);
+    LLVMContextSetOpaquePointers(context, false);
 #endif
-  return context;
+    return context;
 }
 
-
 API_EXPORT(LLVMContextRef)
-LLVMPY_ContextCreate() { 
+LLVMPY_ContextCreate() {
     LLVMContextRef context = LLVMContextCreate();
 #if LLVM_VERSION_MAJOR > 14
     LLVMContextSetOpaquePointers(context, false);
--- a/ffi/core.h
+++ b/ffi/core.h
@@ -2,6 +2,11 @@
 #define LLVMPY_CORE_H_
 
 #include "llvm-c/Core.h"
+
+// Needed for macros that control version-specific behaviour - included here so
+// that they are available in all ffi translation units
+#include "llvm/Config/llvm-config.h"
+
 #include <cstdlib>
 #include <cstring>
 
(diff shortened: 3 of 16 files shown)


yashssh:
Removing `draft` tag as it won't allow me to add reviewers. Adding "draft" to the title instead.

ffi/Makefile.linux
```diff
@@ -2,7 +2,7 @@
 CXX ?= g++
 
 # -flto and --exclude-libs allow us to remove those parts of LLVM we don't use
-CXX_FLTO_FLAGS ?= -flto
+#CXX_FLTO_FLAGS ?= -flto
```
yashssh, line 5: Kindly ignore this. I was facing some difficulties building locally so had to turn LTO off until I sort that.

gmarkall:
```suggestion
CXX_FLTO_FLAGS ?= -flto
```

ffi/core.cpp
```diff
@@ -21,10 +21,23 @@ API_EXPORT(void)
 LLVMPY_DisposeString(const char *msg) { free(const_cast<char *>(msg)); }
 
 API_EXPORT(LLVMContextRef)
-LLVMPY_GetGlobalContext() { return LLVMGetGlobalContext(); }
+LLVMPY_GetGlobalContext() {
+  auto context = LLVMGetGlobalContext();
+#if LLVM_VERSION_MAJOR > 14
+  LLVMContextSetOpaquePointers(context, false);
```
gmarkall, line 27: Note to other reviewers - this does result in opaque pointers being disabled for the global context multiple times, but this operation is idempotent.

ffi/core.h
```diff
@@ -2,6 +2,11 @@
 #define LLVMPY_CORE_H_
 
 #include "llvm-c/Core.h"
+
+// Needed for macros that control version-specific behaviour - included here so
+// that they are available in all ffi translation units
```
gmarkall, line 7: In particular, this defines the `LLVM_VERSION_MAJOR` macro referenced in various places in this PR.

llvmlite/tests/test_binding.py
```diff
@@ -653,7 +654,7 @@ def break_up_asm(self, asm):
     def test_rv32d_ilp32(self):
         self.check_riscv_target()
         llmod = self.fpadd_ll_module()
-        target = self.riscv_target_machine(features="+f,+d")
+        target = self.riscv_target_machine(features="+f,+d", abiname="ilp32")
```
gmarkall, line 657: LLVM 14 defaults to the soft-float ABI, LLVM 15 defaults to the hard-float ABI. We explicitly specify the soft-float ABI for consistency with the expected output. Note that hard float and hard double ABIs are tested in the subsequent two tests below.

llvmlite/tests/test_binding.py
```diff
@@ -783,12 +784,14 @@ def test_set_option(self):
             """
         subprocess.check_call([sys.executable, "-c", code])
 
+    @unittest.skip('Edit this test to test for both llvm14 and llvm15')
```
gmarkall, line 787: I think this is a leftover from prior to when you had the test passing.
```suggestion
    @unittest.skip('Edit this test to test for both llvm14 and llvm15')
```

llvmlite/tests/test_binding.py
```diff
@@ -786,9 +787,10 @@ def test_set_option(self):
     def test_version(self):
         major, minor, patch = llvm.llvm_version_info
         # one of these can be valid
-        valid = [(14, )]
-        self.assertIn((major,), valid)
-        self.assertIn(patch, range(10))
+        valid = (14, 15)
+        self.assertIn(major, valid)
+        self.assertIn(patch, range(8))
```
gmarkall, line 792: (Max patch versions are 14.0.6 and 15.0.7)

gmarkall (approved):
The code changes here have my approval - I've already done a couple of rounds of review internally on this PR, and I made a few edits for a couple of left over changes that needed removing (adjustments to the Makefile, and a test that was accidentally skipped).

I added the CI config so we can test with LLVM 15 from conda-forge on each platform - someone else might want to review those changes.

I ran the Numba test suite with this PR locally, and didn't encounter any issues - the two passes that aren't available with the legacy pass manager anymore (arg promotion and loop unswitch) don't seem to be used by Numba.

As I see it, this PR provides us a way to give experimental support for LLVM 15 once merged, with no further blockers. Following that, I think we can:

- Build an LLVM 15 llvmdev package and move testing to LLVM 15 exclusively, with the aim of officially moving to 15.
- Move to the new pass manager, and re-enable the passes that have to be disabled with LLVM 15 at present.
- Switch to opaque pointers (at least in the binding layer)
- Move to LLVM 16+

Some discussion of the ordering / dependencies of these is warranted though - perhaps at the next office hours.

ffi/passmanagers.cpp
```diff
@@ -161,8 +161,13 @@ LLVMPY_AddCallGraphDOTPrinterPass(LLVMPassManagerRef PM) {
 
 API_EXPORT(void)
 LLVMPY_AddDotDomPrinterPass(LLVMPassManagerRef PM, bool showBody) {
+#if LLVM_VERSION_MAJOR <= 14
```
modiking, line 164: Suggest being consistent with the #ifs, either all "<= 14" or all "< 15"

ffi/passmanagers.cpp
```diff
@@ -333,12 +345,14 @@ LLVMPY_AddLoopUnrollAndJamPass(LLVMPassManagerRef PM) {
     LLVMAddLoopUnrollAndJamPass(PM);
 }
 
+#if LLVM_VERSION_MAJOR < 15
 API_EXPORT(void)
 LLVMPY_AddLoopUnswitchPass(LLVMPassManagerRef PM, bool optimizeForSize,
```
modiking, line 350: For reference the loop unswitch pass has been renamed to `SimpleLoopUnswitchPass` in the new pass manager (https://reviews.llvm.org/D124376). Naturally to enable this we need to move to NPM so it makes sense that this is disabled for now but it can come back with NPM.

ffi/passmanagers.cpp
```diff
@@ -244,10 +254,12 @@ LLVMPY_AddAlwaysInlinerPass(LLVMPassManagerRef PM, bool insertLifetime) {
     unwrap(PM)->add(llvm::createAlwaysInlinerLegacyPass(insertLifetime));
 }
 
+#if LLVM_VERSION_MAJOR < 15
 API_EXPORT(void)
 LLVMPY_AddArgPromotionPass(LLVMPassManagerRef PM, unsigned int maxElements) {
```
modiking, line 259: Ditto with `ArgumentPromotionPass` in NPM as below

ffi/passmanagers.cpp
modiking: Actually `createSimpleLoopUnswitchLegacyPass` still exists in LLVM-15 so it can be dropped in here without issue.

modiking (commented):
Thanks for putting this up!

ffi/passmanagers.cpp
modiking:
```
API_EXPORT(void)
LLVMPY_AddLoopUnswitchPass(LLVMPassManagerRef PM, bool optimizeForSize,
                           bool hasBranchDivergence) {
#if LLVM_VERSION_MAJOR < 15
    unwrap(PM)->add(
        createLoopUnswitchPass(optimizeForSize, hasBranchDivergence));
#else
    unwrap(PM)->add(
        createSimpleLoopUnswitchLegacyPass(optimizeForSize));
#endif
}
```

yashssh: Thanks for finding this, I have updated the code.

ffi/passmanagers.cpp
yashssh: Did you mean legacy version of this pass is available as well? I couldn't find it.

gmarkall (approved):
Re-approving following latest changes.

modiking:
LGTM

gmarkall:
Added "Pending Buildfarm" to make sure this is all OK on all other platforms.

sklam:
BFID: `llvmlite_275`

sklam:
`llvmlite_275` passed

sklam:
Also passed buildfarm test against numba main (`numba_yaml_448`)

sklam (approved):

PR #1036, opened 2024-04-01
LLVM 15 conda recipe

sklam:


sklam:
f0def581a930e11f730073c82b03284b370aef72 builds successfully on all platform in the buildfarm

sklam:
passed numba testsuite when running manually on linux-64 machine

gmarkall:
We need to merge this and build the llvmdev package, as I understand it. (This requires the setup internal to Anaconda on their build farm).

After that, I think we can switch the llvmlite CI over to use LLVM 15 by default, and start to move it from being experimental to being the default supported version.

sklam:
BFID `llvmdev_llvm15_4`

sklam:
> BFID `llvmdev_llvm15_4`

build passed

sklam:
Now testing llvmdev build against llvmlite test in `llvmlite_llvm15_2`

sklam:
`llvmlite_llvm15_3` also passed.

gmarkall (approved):
I had previously reviewed this and tested it locally and my understanding is that we were only waiting on a build farm run, so I think this is OK to move to RTM and build new LLVMdev 15 packages for the main channel so we can switch llvmlite over to more general testing with LLVM 15.

gmarkall:
Marked RTM as per discussion in triage meeting.

PR #986, opened 2023-08-24
Statically link compiler-rt

oliverhu:
This PR addesses issues in https://github.com/numba/llvmlite/issues/834 and https://github.com/tinygrad/tinygrad/issues/1367, in which libllvmlite.so fails to resolve symbols in libclang_rt.builtin.a since compiler rt is not linked. This PR statically links the compiler_rt library.

(address the comments from https://github.com/numba/llvmlite/pull/976)

PR #1038, opened 2024-04-15
Test #986 on CI with LLVM 15

gmarkall:
I can't push to #986 so this PR is for testing #986 on CI. Initially this consists of the `ohu/crt` branch with `main` from https://github.com/numba/llvmlite merged in.

CC @oliverhu


gmarkall:
I think this is superfluous now, I'll re-visit #986 following #1067's merge.

PR #1045, opened 2024-04-29
remove all refops if a basic block is a raising block

### LINKED ISSUE #1044: A concrete case for potential ref-pruning chances
code of `typed.List`'s `get_item`:
```python
@overload(operator.getitem)
def impl_getitem(l, index):
    ...
    if index in index_types:
        if IS_NOT_NONE:
            def integer_non_none_impl(l, index):
                castedindex = _cast(index, indexty)
                handledindex = handle_index(l, castedindex)
                status, item = _list_getitem(l, handledindex)
                if status == ListStatus.LIST_OK:
                    return _nonoptional(item)
                else:
                    raise AssertionError("internal list error during getitem")
            return integer_non_none_impl
```
IR of `integer_non_none_impl`:
```llvm
entry:
  tail call void @NRT_incref(i8* %arg.l.0)
  tail call void @NRT_incref(i8* %arg.l.0), !noalias !22
  ...
  br i1 %.27.i, label %B0.if, label %B20.endif.i

B20.endif.i:                                      ; preds = %entry
  ...
  tail call void @NRT_decref(i8* %arg.l.0), !noalias !22
  ...
  br i1 %.60.not.i, label %B46.endif, label %B0.if

common.ret:                                       ; preds = %B46.endif, %B0.if
  %common.ret.op = phi i32 [ 0, %B46.endif ], [ 1, %B0.if ]
  ret i32 %common.ret.op

B0.if:                                            ; preds = %B20.endif.i, %entry
  %l.sroa.0.0.i = phi i8* [ %arg.l.0, %entry ], [ null, %B20.endif.i ]
  tail call void @NRT_decref(i8* %l.sroa.0.0.i), !noalias !22
  store ..., ... %excinfo, align 8, !numba_exception_output !13
  br label %common.ret

B46.endif:                                        ; preds = %B20.endif.i
  ...
  tail call void @NRT_decref(i8* %arg.l.0)
  ...
  br label %common.ret
```
Ideally, we should remove all refcount calls of `%arg.l.0` and `%l.sroa.0.0.i`, since `%l.sroa.0.0.i` is `null` or `%arg.l.0`.
However, current ref-pruning algorithm can't remove any of them.

dlee992:
A possible way to 
fix #1044 

Notes for reviewers:
- remove all refops if a basic block is a raising block, since we allow memory leak in a raising block
- put this pruning operation in the per_bb pattern
- after per_bb pattern removes this case, fanout_raise pattern can do the rest as before

### DIFF
--- a/ffi/custom_passes.cpp
+++ b/ffi/custom_passes.cpp
@@ -181,8 +181,7 @@ struct RefPrunePass : public FunctionPass {
         Diamond = 0b0010,
         Fanout = 0b0100,
         FanoutRaise = 0b1000,
-        RefInRaise = 0b10000,
-        All = PerBasicBlock | Diamond | Fanout | FanoutRaise | RefInRaise
+        All = PerBasicBlock | Diamond | Fanout | FanoutRaise
     } flags;
 
     RefPrunePass(Subpasses flags = Subpasses::All, size_t subgraph_limit = -1)
@@ -211,8 +210,6 @@ struct RefPrunePass : public FunctionPass {
                 local_mutated |= runFanoutPrune(F, /*prune_raise*/ false);
             if (isSubpassEnabledFor(Subpasses::FanoutRaise))
                 local_mutated |= runFanoutPrune(F, /*prune_raise*/ true);
-            if (isSubpassEnabledFor(Subpasses::RefInRaise))
-                local_mutated |= runRefInRaise(F);
             mutated |= local_mutated;
         } while (local_mutated);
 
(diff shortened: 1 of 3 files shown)


dlee992:
Hi, @sklam. This is another small improvement for ref-pruning. It can enjoy your review! Thanks

dlee992:
Hi, any comment for this PR and the related issue? cc @esc

esc:
@dlee992 I suppose this ought to make `typed.List` faster? If so, do you have any benchmarks to share?

dlee992:
Yeah, I will make up one test case for it tmr. 
In theory, it will definitely help the performance since we prune more atomic refcounts.

dlee992:
> @dlee992 I suppose this ought to make typed.List faster? If so, do you have any benchmarks to share?

Actually, this one could be quite tricky to prove, and I don't have an existing reproducer/benchmark to show the performance benefits. So close it for now.

Issue #1043, opened 2024-04-25
Upload wheels and sdist to PyPI. (upload from `ci_artifacts`).


sklam:

sklam:
I think i accidentally clicked something to open this

PR #1046, opened 2024-05-02
Add basic infra required to move Numba to NewPassManager

yashssh:
Add separate classes and API endpoints for new pass manager and gradually plan on retiring/deleting legacy pass manger code.

This can be treated as a self contained starter patch to move Numba to NewPassManager. In following patches things like hooks for all other passes, refprune pass, more tests, TODOs and FIXMEs can be taken care of.

### DIFF
--- a/.github/ISSUE_TEMPLATE/first_rc_checklist.md
+++ b/.github/ISSUE_TEMPLATE/first_rc_checklist.md
@@ -18,6 +18,7 @@ labels: task
 * [ ] Test `HEAD` of release branch on buildfarm (pre-tag testing):
     * [ ] conda package build and test.
     * [ ] wheel build.
+* [ ] Test `HEAD` of release branch on conda-forge
 * [ ] Annotated tag `vX.Y.Zrc1` on release branch (`llvmlite` tags DO have a `v` prefix).
 * [ ] Build and upload conda packages on buildfarm (check "upload").
 * [ ] Build wheels and sdist on the buildfarm (check "upload").
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -67,11 +67,6 @@ jobs:
         CONDA_ENV: cienv
         WHEEL: 'yes'
 
-      py39_docs:
-        PYTHON: '3.9'
-        CONDA_ENV: cienv
-        BUILD_DOCS: 'yes'
-
       llvm15:
         PYTHON: '3.12'
         CONDA_ENV: cienv
--- a/buildscripts/azure/azure-linux-macos.yml
+++ b/buildscripts/azure/azure-linux-macos.yml
@@ -42,17 +42,6 @@ jobs:
     - script: |
         set -e
         export PATH=$HOME/miniconda3/bin:$PATH
-        conda install -y sphinx sphinx_rtd_theme pygments
-        echo "Building docs"
-        cd docs
-        make SPHINXOPTS=-Wn clean html
-        cd ..
-      displayName: 'Build Docs'
-      condition: eq(variables['BUILD_DOCS'], 'yes')
-
-    - script: |
-        set -e
-        export PATH=$HOME/miniconda3/bin:$PATH
         buildscripts/incremental/test.sh
       displayName: 'Test'
 # This is last because we need to install a bunch of LLVM stuff that will pollute the build environment
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -304,3 +304,5 @@ intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
     'llvm': ('http://llvm.org/releases/14.0.0/docs', None),
     }
+
+nitpicky = True


ffi/Makefile.linux
```diff
@@ -2,7 +2,7 @@
 CXX ?= g++
 
 # -flto and --exclude-libs allow us to remove those parts of LLVM we don't use
-CXX_FLTO_FLAGS ?= -flto
+#CXX_FLTO_FLAGS ?= -flto
```
yashssh, line 5: Please ignore this for now, for some reasons I can't build llvmlite locally with lto enabled.

ffi/CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ endif()
 add_library(llvmlite SHARED assembly.cpp bitcode.cpp core.cpp initfini.cpp
             module.cpp value.cpp executionengine.cpp transforms.cpp type.cpp
             passmanagers.cpp targets.cpp dylib.cpp linker.cpp object_file.cpp
-            custom_passes.cpp orcjit.cpp memorymanager.cpp)
+            custom_passes.cpp orcjit.cpp memorymanager.cpp newpassmanagers.cpp)
```
gmarkall, line 47: We'll soon (relatively) be in a position where there are only the new pass managers, so I think it would be better to rename the existing `passmanagers.cpp` to `legacypassmanagers.cpp`, and make the new file `passmanagers.cpp`.

ffi/Makefile.linux
gmarkall: IIRC from my prior experience, the reason is that older versions of GCC will ICE when linking LLVM with LTO on AArch64, and I think this can happen if you're using e.g. the system compilers on Ubuntu. Moving to a newer GCC ought to resolve this.

ffi/newpassmanagers.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+/*
+ * Exposed API
+ */
+
+namespace llvm {
+
+struct OpaqueModulePassManager;
+typedef OpaqueModulePassManager *LLVMModulePassManagerRef;
+DEFINE_SIMPLE_CONVERSION_FUNCTIONS(ModulePassManager, LLVMModulePassManagerRef)
```
gmarkall, line 23: We should probably be using this macro whenever we want `wrap()` and `unwrap()` functions elsewhere in llvmlite - I am not sure why we're not doing that already.

gmarkall:
Just starting to look at this, but a general point - although the new pass manager is referred to as such in documentation, discussions, etc., to distinguish it from the legacy pass manager, can we avoid using "new" to refer to it in the code? There will be a time when the pass manager is not new, and then the name will be obsolete - instead can we just use the names like `ModulePassManager`, `FunctionPassManager`, `PassBuilder`, etc. where appropriate?

If we need to disambiguate between them in the names, I'd prefer if we renamed existing code to use the term "legacy" instead, where we can do so without exposing the change externally - for example, renaming `passmanagers.cpp` to `legacypassmanagers.cpp`, c.f. keeping the name `PassManager` for the Python legacy pass manager class.

ffi/newpassmanagers.cpp
```diff
@@ -0,0 +1,317 @@
...
+}
+
+} // namespace llvm
+
+extern "C" {
+
+// MPM
+
+API_EXPORT(LLVMModulePassManagerRef)
+LLVMPY_CreateNewModulePassManager() {
```
gmarkall, line 50: As an example of the avoidance of "new":
```suggestion
LLVMPY_CreateModulePassManager() {
```

ffi/newpassmanagers.cpp
```diff
@@ -0,0 +1,317 @@
...
+    return llvm::wrap(new PassManager<Module>());
+}
+
+API_EXPORT(void)
+LLVMPY_NMPRun_module(LLVMModulePassManagerRef MPMRef, LLVMPassBuilderRef PBRef,
```
gmarkall, line 55: I presume `NMP` is a typo and was intended to stand for "new pass manager", but I think perhaps this would be better named:
```suggestion
LLVMPY_RunModulePassManager(LLVMModulePassManagerRef MPMRef, LLVMPassBuilderRef PBRef,
```

or

```suggestion
LLVMPY_ModulePassManagerRun(LLVMModulePassManagerRef MPMRef, LLVMPassBuilderRef PBRef,
```

(probably the first option makes more sense, but I'm not 100% sure yet)

ffi/newpassmanagers.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+    LoopAnalysisManager LAM;
+    FunctionAnalysisManager FAM;
+    CGSCCAnalysisManager CGAM;
+    ModuleAnalysisManager MAM;
+    PB->registerLoopAnalyses(LAM);
+    PB->registerFunctionAnalyses(FAM);
+    PB->registerCGSCCAnalyses(CGAM);
+    PB->registerModuleAnalyses(MAM);
+    PB->crossRegisterProxies(LAM, FAM, CGAM, MAM);
```
gmarkall, line 70: Can you run a module pass manager multiple times? If so, what are the semantics WRT these analysis managers? They appear to go out of scope at the end of this function, but does the `PassBuilder` consume them? Does it expect them to outlive it? What happens when analysis managers are registered a second time?

yashssh: >Can you run a module pass manager multiple times? If so, what are the semantics WRT these analysis managers?

Yes we can, we will invoke this function again. Re-register all the analyses passes and then run the pipeline.
> They appear to go out of scope at the end of this function, but does the PassBuilder consume them? Does it expect them to outlive it?

They indeed go out of scope and PassBuilder doesn't consume/store them. The reason why we are creating all these analyses managers is various analyses passes can be registered only under their own analyses managers. For eg there will be some loop analyses registered under LoopAnalysesManager, but while optimizing a module we need the ability to run these loop analyses on loops inside the module. So we register analyses passes in under their respective managers and cross register them with each other to solve the problem I described above.

>What happens when analysis managers are registered a second time?

You are free to create as many managers as you want and they are independent of each other. All you need is populated + cross registered corresponding analysis manager whenever invoking `run` on the pass manager (FAM for FPM, MAM for MPM).

ffi/newpassmanagers.cpp
yashssh: yes I thought the same, will take a look.

yashssh:
I agree with the naming convention comments. We should prefix the current pass manager files and classes with `legacy` prefix and remove `new` from the NewPassManager classes and files. I will have to check no current API breaks as a result of this change though.

gmarkall:
Another note: With the changes in this PR and https://github.com/gmarkall/numba/tree/npm, I was able to test Numba with the new module pass manager (but not function pass manager). This resulted in: 

```
Ran 11969 tests in 1314.257s

FAILED (failures=5, errors=3, skipped=607, expected failures=33)
```

from the testsuite, which is not too bad. I had to disable the refcount pruning pass, so some of the failures are about that. There are a few other failures I haven't looked into yet, but given the number of passing test cases I'm not too worried.

In #1043 I asked about the porting of the refcount pruning pass to the new pass manager on top of this PR, which may enable me to test more thoroughly.

ffi/CMakeLists.txt
yashssh: done in latest revision

llvmlite/binding/transforms.py
yashssh, line 1: Code moved to legacypassmanagers.py

llvmlite/binding/passmanagers.py
yashssh, line 1: Renamed to legacypassmanagers.py

ffi/transforms.cpp
yashssh, line 1: code moved to legacypassmanagers.cpp

ffi/passmanagers.cpp
yashssh, line 1: Renamed to legacypassmanagers.cpp

ffi/newpassmanagers.cpp
gmarkall: Ah, sorry, I didn't mean to imply you should sort that out in this PR. It could/should be orthogonally done in another PR rather than piling more changes in here.

gmarkall:
A quick summary of changes / work needed towards getting this over the line, from OOB discussion:

- [ ] The example `npm-usage.py` should be in the `examples` dir, and tidied up / split up into multiple examples each demonstrating a bit less functionality than the entire example (does the example code need to be so large?)
- [ ] The documentation at https://llvmlite.readthedocs.io/en/latest/user-guide/binding/optimization-passes.html needs updating to reflect the new pass managers. Would you like me to work on this documentation? Let me know if so.
- [ ] Regarding the name `newpassmanagers.cpp` - lets leave it as it is, and minimise hassle. We can always rename it from `newpassmanagers.cpp` when we delete `passmanagers.cpp`.
- [ ] `ffi/Makefile.linux` - delete the removal of LTO flags, just keep this change locally.
- [ ] I need to review `newpassmanagers.cpp` and properly comprehend your answer to my earlier query.
- [ ] We need to find the minimal change that doesn't bake the word "new" into the public Python API for the new pass managers, nor change the existing API.

yashssh:
> The example npm-usage.py should be in the examples dir, and tidied up / split up into multiple examples each demonstrating a bit less functionality than the entire example (does the example code need to be so large?)

Done 

> The documentation at https://llvmlite.readthedocs.io/en/latest/user-guide/binding/optimization-passes.html needs updating to reflect the new pass managers. Would you like me to work on this documentation? Let me know if so.

I tried adding it but please do add more/modify it if it isn't sufficient.

>  Regarding the name newpassmanagers.cpp - lets leave it as it is, and minimise hassle. We can always rename it from newpassmanagers.cpp when we delete passmanagers.cpp.

Sure

> ffi/Makefile.linux - delete the removal of LTO flags, just keep this change locally.

Fixed

>We need to find the minimal change that doesn't bake the word "new" into the public Python API for the new pass managers, nor change the existing API.

PassBuilder.getNewModulePassManager can be renamed to PassBuilder.getModulePassManager without consequences, same for the FPM api. But I don't see how we can get around removing `new` from the new classes as the names would clash with old Pass manager classes. Can we prefix the old PM classes with `legacy` prefix?

gmarkall:
Note to any other reviewers / interested parties: whilst reviewing this, I came across this blog post, which is helpful for understanding the new pass manager and the design behind it: https://blog.llvm.org/posts/2021-03-26-the-new-pass-manager/

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
+from ctypes import (c_bool, c_int)
```
gmarkall, line 1: No need for the brackets:

```suggestion
from ctypes import c_bool, c_int
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+from llvmlite.binding import ffi
+from llvmlite.binding.initfini import llvm_version_info
+
+llvm_version_major = llvm_version_info[0]
```
gmarkall, line 5: This appears to be unused?

```suggestion
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+
+def create_new_function_pass_manager():
+    return NewFunctionPassManger()
+
+
+def create_pass_builder(tm, pto):
+    return PassBuilder(tm, pto)
+
+
+def create_pipeline_options():
```
gmarkall, line 20: For consistency:
```suggestion
def create_pipeline_tuning_options():
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+    return PipelineTuningOptions()
+
+
+class NewModulePassManager(ffi.ObjectRef):
+
+    def __init__(self, ptr=None):
+        if ptr is None:
+            ptr = ffi.lib.LLVMPY_CreateNewModulePassManager()
+        ffi.ObjectRef.__init__(self, ptr)
```
gmarkall, line 29: I think it would be more idiomatic to do:

```suggestion
        super().__init__(ptr)
```

(the old pass manager is this way because it was written when Python 2 was still supported, where you couldn't do this).

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def run(self, module, pb):
+        ffi.lib.LLVMPY_NMPRun_module(self, pb, module)
```
gmarkall: Should be changed in alignment with https://github.com/numba/llvmlite/pull/1046/files/445ebb46764ccea4726640dd2e0970b9213d70de#r1595274582:

```suggestion
        ffi.lib.LLVMPY_RunModulePassManager(self, pb, module)
```

(for example)

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def addVerifier(self):
+        ffi.lib.LLVMPY_AddVeriferPass(self)
```
gmarkall, line 35: Just noticed this is mis-spelled:
```suggestion
        ffi.lib.LLVMPY_AddVerifierPass(self)
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+        ffi.lib.LLVMPY_AddSimplifyCFGPass_module(self)
+
+    def add_loop_unroll_pass(self):
+        ffi.lib.LLVMPY_AddLoopUnrollPass_module(self)
+
+    def add_loop_rotate_pass(self):
+        ffi.lib.LLVMPY_LLVMAddLoopRotatePass_module(self)
+
+    def add_instruction_combine_pass(self):
+        ffi.lib.LLVMPY_LLVMAddInstructionCombinePass_module(self)
```
gmarkall, line 50: Similar removal of "LLVM" to above:

```suggestion
        ffi.lib.LLVMPY_AddInstructionCombinePass_module(self)
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def add_jump_threading_pass(self, threshold=-1):
```
gmarkall, line 52: Note the default threshold matches that in LLVM's C++: https://llvm.org/doxygen/classllvm_1_1JumpThreadingPass.html#a3cd21bd06a36557dd863c554b7b10205

gmarkall: (for the benefit of other reviewers, not a comment towards @yashssh )

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+        ffi.lib.LLVMPY_AddJumpThreadingPass_module(self, threshold)
+
+    # What will happen to the pointer to objects created with run() method
```
gmarkall, line 55: What pointer is that? I can't see a pointer to objects created with the `run()` method.

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+    def _dispose(self):
+        ffi.lib.LLVMPY_DisposeNewModulePassManger(self)
+
+
+class NewFunctionPassManger(ffi.ObjectRef):
+
+    def __init__(self, ptr=None):
+        if ptr is None:
+            ptr = ffi.lib.LLVMPY_CreateNewFunctionPassManager()
+        ffi.ObjectRef.__init__(self, ptr)
```
gmarkall, line 65: `super()` as above:
```suggestion
        super().__init__(ptr)
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def add_aa_eval_pass(self):
+        ffi.lib.LLVMPY_AddAAEvalPass_function(self)
+
+    def add_simplify_cfg_pass(self):
+        ffi.lib.LLVMPY_AddSimplifyCFGPass_function(self)
+
+    def run(self, fun, pb):
+        ffi.lib.LLVMPY_NMPRun_function(self, pb, fun)
```
gmarkall: Should probably be something like:
```suggestion
        ffi.lib.LLVMPY_RunFunctionPassManager(self, pb, fun)
```

(similar to above for the module pass manager's `run()` function)

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+        ffi.lib.LLVMPY_LLVMAddInstructionCombinePass_function(self)
+
+    def add_jump_threading_pass(self, threshold=-1):
+        ffi.lib.LLVMPY_AddJumpThreadingPass_function(self, threshold)
+
+    def _dispose(self):
+        ffi.lib.LLVMPY_DisposeNewFunctionPassManger(self)
+
+
+# Make this nested class of PassBuilder?
```
gmarkall, line 92: It doesn't seem to be nested in LLVM C++, so I don't think we should nest it in the Python without a compelling reason.

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+class PipelineTuningOptions(ffi.ObjectRef):
+
+    def __init__(self, ptr=None):
+        if ptr is None:
+            ptr = ffi.lib.LLVMPY_CreatePipelineTuningOptions()
+            self._opt_level = 2
+            self._size_level = 0
```
gmarkall, line 99: Would it make sense to have default keyword arguments to the `__init__()` method for these? Otherwise if you want something non-default you have to create a `PipelineTuningOptions` then mutate it.

Rather than having getters and setters for the property like below, we could just write:

```suggestion
            self.opt_level = opt_level
            self.size_level = size_level
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+        ffi.lib.LLVMPY_AddJumpThreadingPass_function(self, threshold)
+
+    def _dispose(self):
+        ffi.lib.LLVMPY_DisposeNewFunctionPassManger(self)
+
+
+# Make this nested class of PassBuilder?
+class PipelineTuningOptions(ffi.ObjectRef):
+
+    def __init__(self, ptr=None):
```
gmarkall, line 95: Suggested kwargs (see comment below):

```suggestion
    def __init__(self, ptr=None, opt_level=2, size_level=0):
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+        if ptr is None:
+            ptr = ffi.lib.LLVMPY_CreatePipelineTuningOptions()
+            self._opt_level = 2
+            self._size_level = 0
+        ffi.ObjectRef.__init__(self, ptr)
```
gmarkall, line 100: `super()` as above:

```suggestion
        super().__init__(ptr)
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+        """
+        Whether and how much to optimize for size.
+        An integer between 0 and 2.
+        """
+        return self._size_level
+
+    @size_level.setter
+    def size_level(self, size):
+        self._size_level = size
+
```
gmarkall, line 166: We can get rid of these and just have public attributes set in `__init__()` as above:
```suggestion
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+        ffi.lib.LLVMPY_DisposePipelineTuningOptions(self)
+
+
+class PassBuilder(ffi.ObjectRef):
+
+    def __init__(self, tm, pto, ptr=None):
+        if ptr is None:
+            self._pto = pto
+            self._tm = tm
+            ptr = ffi.lib.LLVMPY_CreatePassBuilder(tm, pto)
```
gmarkall, line 177: `pto` and `tm` will be ignored if a pointer is passed in - is that what we'd want? Or should it be:

```suggestion
            ptr = ffi.lib.LLVMPY_CreatePassBuilder(tm, pto)
        self._pto = pto
        self._tm = tm
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+        ffi.ObjectRef.__init__(self, ptr)
+
+    def getNewModulePassManager(self):
```
gmarkall, line 180: I think we can avoid the `New` here:

```suggestion
    def getModulePassManager(self):
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+        return NewModulePassManager(
+            ffi.lib.LLVMPY_buildPerModuleDefaultPipeline(
+                self, self._pto.opt_level, self._pto.size_level)
+        )
+
+    def getNewFunctionPassManager(self):
```
gmarkall, line 186: As above:
```suggestion
    def getFunctionPassManager(self):
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+        return NewFunctionPassManger(
+            ffi.lib.LLVMPY_buildFunctionSimplificationPipeline(
+                self, self._pto.opt_level, self._pto.size_level)
+        )
+
+    def _dispose(self):
+        # Should I explicitly delete pointer to pto and tm?
```
gmarkall, line 193: No, they will be disposed when they are collected by the GC after they're out of scope everywhere.

gmarkall:
Thanks for all the changes! I just finished another round of review and pushed the small tidy-up / reorganization we discussed.

> > The documentation at https://llvmlite.readthedocs.io/en/latest/user-guide/binding/optimization-passes.html needs updating to reflect the new pass managers. Would you like me to work on this documentation? Let me know if so.
> 
> I tried adding it but please do add more/modify it if it isn't sufficient.
> 

Thanks, I'll have a look over this in the next iteration of review, as I'm running out of time for today.

> PassBuilder.getNewModulePassManager can be renamed to PassBuilder.getModulePassManager without consequences, same for the FPM api. But I don't see how we can get around removing `new` from the new classes as the names would clash with old Pass manager classes. Can we prefix the old PM classes with `legacy` prefix?

The class names won't class because they're in different modules - `llvmlite.binding.passmanagers.ModulePassManager` and `llvmlite.binding.newpassmanagers.ModulePassManager` would be distinct. That said, I'm still thinking about this because it still bakes the word "new" into a public API.

llvmlite/binding/newpassmangers.py
yashssh: My bad, I was referring to the Analysis manager objects created in `LLVMPY_RunNew[Function|Module]PassManager` , but it's clear now they are just local objects with automatic storage, which should be managed implicitly.

llvmlite/binding/newpassmangers.py
yashssh: Agree looks cleaner, made them default arguments.

yashssh:
Thank you for the review @gmarkall! I have updated the code to reflect the feedback.

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,283 @@
...
+    return PipelineTuningOptions(opt_level=opt_level, size_level=size_level)
+
+
+class NewModulePassManager(ffi.ObjectRef):
+
+    def __init__(self, ptr=None):
+        if ptr is None:
+            ptr = ffi.lib.LLVMPY_CreateNewModulePassManager()
+        super().__init__(ptr)
+
```
gmarkall, line 27: As far as I can tell there isn't a use case for passing in a pointer to an existing LLVM object, which I think is inherited from the style in `passmanagers.py`. Let's just get rid of that capability:

```suggestion
    def __init__(self):
        super().__init__(ffi.lib.LLVMPY_CreateNewModulePassManager())
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,283 @@
...
+    def _dispose(self):
+        ffi.lib.LLVMPY_DisposeNewModulePassManger(self)
+
+
+class NewFunctionPassManger(ffi.ObjectRef):
+
+    def __init__(self, ptr=None):
+        if ptr is None:
+            ptr = ffi.lib.LLVMPY_CreateNewFunctionPassManager()
+        super().__init__(ptr)
```
gmarkall, line 61: Always creating a new underlying object, as above:

```suggestion
    def __init__(self):
        super().__init__(ffi.lib.LLVMPY_CreateNewFunctionPassManager())
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+class PipelineTuningOptions(ffi.ObjectRef):
+
+    def __init__(self, ptr=None, opt_level=2, size_level=0):
+        if ptr is None:
+            ptr = ffi.lib.LLVMPY_CreatePipelineTuningOptions()
+        self.opt_level = opt_level
+        self.size_level = size_level
+        super().__init__(ptr)
```
gmarkall, line 95: As above:
```suggestion
    def __init__(self, opt_level=2, size_level=0):
        super().__init__(ffi.lib.LLVMPY_CreatePipelineTuningOptions())
        self.opt_level = opt_level
        self.size_level = size_level
```

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+class PassBuilder(ffi.ObjectRef):
+
+    def __init__(self, tm, pto, ptr=None):
+        if ptr is None:
+            ptr = ffi.lib.LLVMPY_CreatePassBuilder(tm, pto)
+        self._pto = pto
+        self._tm = tm
+        super().__init__(ptr)
```
gmarkall, line 149: As above:
```suggestion
    def __init__(self, tm, pto):
        super().__init__(ffi.lib.LLVMPY_CreatePassBuilder(tm, pto))
        self._pto = pto
        self._tm = tm
```

gmarkall (changes requested):
Just had another pass over passmanagers.py - I think there's some simplifications I think we can make... Let me know if I've missed something and there is a use case for passing in pointers to existing objects that you know of though.

llvmlite/binding/newpassmangers.py
yashssh: > As far as I can tell there isn't a use case for passing in a pointer to an existing LLVM object, which I think is inherited from the style in `passmanagers.py`. Let's just get rid of that capability:

Yes that was why I added them here, I can't think of a use for this feature from python user's pov

yashssh: But it's being used internally. Just checked `PassBuilder` relies on `ptr` based initialization of `MPM` and `FPM` objects, when we call `PB.getModulePassManager()` and `PB.getFunctionPassManager()` APIs respectively.

We can get rid of it at other 2 places though (PassBuilder and PipelineTuningOptions's constructors).

gmarkall: Ah, OK - well spotted!

ffi/newpassmanagers.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+    PassBuilder *PB = llvm::unwrap(PBref);
+    OptimizationLevel OL = mapLevel(speed_level, size_level);
+    if (OL == OptimizationLevel::O0) {
+        return llvm::wrap(
+            new ModulePassManager(PB->buildO0DefaultPipeline(OL)));
+    }
+
+    return llvm::wrap(
+        new ModulePassManager(PB->buildPerModuleDefaultPipeline(OL)));
```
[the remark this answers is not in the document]
yashssh: Yes, exactly. Yeah a FIXME here will be good.

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,279 @@
...
+    def add_jump_threading_pass(self, threshold=-1):
+        ffi.lib.LLVMPY_AddJumpThreadingPass_function(self, threshold)
+
+    def _dispose(self):
+        ffi.lib.LLVMPY_DisposeNewFunctionPassManger(self)
+
+
+class PipelineTuningOptions(ffi.ObjectRef):
+
+    def __init__(self, opt_level=2, size_level=0):
```
gmarkall, line 90: The `opt_level` and `size_level` arguments need to be validated - presently I can pass invalid values like `speed_level=4, size_level=-1` - with a release build this will result in junk values ending up somewhere being passed to the LLVM APIs, and I imagine the `llvm_unreachable`s in `mapLevel` above would end up killing the interpreter with a debug build. I think we need to check for validity using the same rules as in the `OptimizationLevel` constructor in LLVM's C++ implementation (speed level 0-3, size level 0-2, optimization for size requires speed level 2).

ffi/newpassmanagers.cpp
```diff
@@ -0,0 +1,319 @@
...
+}
+
+API_EXPORT(LLVMFunctionPassManagerRef)
+LLVMPY_buildFunctionSimplificationPipeline(LLVMPassBuilderRef PBref,
```
aseyboldt, line 306: I hope it's ok if I chime in a little here...
I think it might be helpful if the ModuleSimplificationPipeline was also exposed? (https://llvm.org/doxygen/classllvm_1_1PassBuilder.html#ad6f258d31ffa2d2e4dfaf990ba596d0d). That could be used to simplify whole modules before they are linked into multiple different modules?

I also wonder how hard it would be to expose some of the extension points for the pipelines (https://llvm.org/docs/NewPassManager.html#inserting-passes-into-default-pipelines) so for instance the `registerPipelineEarlySimplificationEPCallback` method on `PassBuilder`? Maybe that would be a nice way of adding the refpune pass to reasonable locations inside a pipeline?

yashssh: > I think it might be helpful if the ModuleSimplificationPipeline was also exposed?

BuildPerModuleDefultPipeline does call this internally, do you explicitly need it?

> I also wonder how hard it would be to expose some of the extension points for the pipelines

Won't be too difficult but it would make the APIs confusing I believe(maybe not, haven't really thought about in it in depth)

If you don't need them urgently, can you create separate tickets for them and I can revisit them again?

aseyboldt: `buildPerModuleDefultPipeline` calls `buildModuleSimplificationPipeline` ([here](https://github.com/llvm/llvm-project/blob/bc946f52870a51b49a0c7c3c508cef1905a11bc0/llvm/lib/Passes/PassBuilderPipelines.cpp#L1579)) but it *also* calls `buildModuleOptimizationPipeline`. And according to the docs that should not be executed repeatedly. So `per_module_default_pipeline(per_module_default_pipeline(module).link_into(module2))` would be incorrect, but `per_module_default_pipeline(module_simplification_pipeline(module1).link_into(module2))` would be fine, right? (That's one of the things I've been experimenting in https://github.com/numba/numba/pull/9566.)

About the register_* functions:
I'm only curious if they can be used to speed up numba compilation, because then fewer total optimization/simplification passes might be needed. Nothing urgent about it, it just might (?) make numba compilation faster (assuming the numba devs agree that that's a good idea anyway...).

gmarkall (changes requested):
I've added a couple of notes on naming I spotted whilst working on the docs.

ffi/CMakeLists.txt
gmarkall: (note that we undid this later - resolving the conversation)

llvmlite/binding/__init__.py
```diff
@@ -7,6 +7,7 @@
 from .linker import *
 from .module import *
 from .options import *
+from .newpassmangers import *
```
gmarkall, line 10: This is mis-spelled:

```suggestion
from .newpassmanagers import *
```

gmarkall:
A new to-do list for myself for the next (and hopefully final, or almost final) iteration:

- [ ] Review / editing / commentary of the examples
- [ ] Notes on / resolve minor edits (e.g. accidental newlines, etc)
- [ ] Review test coverage

dlee992:
Thanks for your efforts! (Allow me to also chime in..)

I suppose after this is merged, numba will have a follow-up PR to use NewPassManager. I found the new one doesn't support 
```
.. FIXME: Available from llvm16
.. * .. attribute:: inlining_threshold
```
And I did tune this threshold internally for a bit better perf. So my concern is which strategy numba will use for this new one? The two coexist for a while, or totally remove the old one. 
Yeah, perhaps discussing this is too soon, since I haven't tested if the NewPassManager can bring other benefits to overcome the temp limitation.

ffi/passmanagers.cpp
```diff
@@ -475,9 +475,7 @@ LLVMPY_AddBasicAliasAnalysisPass(LLVMPassManagerRef PM) {
 }
 
 API_EXPORT(void)
-LLVMPY_LLVMAddLoopRotatePass(LLVMPassManagerRef PM) {
-    LLVMAddLoopRotatePass(PM);
-}
+LLVMPY_AddLoopRotatePass(LLVMPassManagerRef PM) { LLVMAddLoopRotatePass(PM); }
```
gmarkall, line 478: Note to other reviewers - there was a slight inconsistency in the naming here, with the redundant `LLVM` in the name, which was removed as part of a bit of general clean-up (not that it matters greatly, as this code will eventually be removed).

gmarkall:
> So my concern is which strategy numba will use for this new one? The two coexist for a while, or totally remove the old one.

The two will coexist for a bit and be selectable with a configuration variable in Numba. This is motivated by the anticipation that there might be regressions in performance or functionality initially.

llvmlite/binding/__init__.py
yashssh: Thanks for catching this, seems like I mis-spelled in the file name. I don't want to ruin the review history of the file so will rename it when the pr is merge ready. Let me know if you suggest otherwise :)

ffi/newpassmanagers.cpp
```diff
@@ -0,0 +1,319 @@
...
+}
+
+API_EXPORT(LLVMModulePassManagerRef)
+LLVMPY_buildPerModuleDefaultPipeline(LLVMPassBuilderRef PBref, int speed_level,
+                                     int size_level) {
+
+    PassBuilder *PB = llvm::unwrap(PBref);
+    OptimizationLevel OL = mapLevel(speed_level, size_level);
+
+    // FXIME: No need to explicitly take care of O0 from LLVM 17
```
gmarkall, line 295: Whilst I enjoy the irony of mis-spelling FIXME, I will sleep a little better at night knowing that it's not typoed:
```suggestion
    // FIXME: No need to explicitly take care of O0 from LLVM 17
```

ffi/newpassmanagers.cpp
gmarkall: Created #1055 to further the discussion(s)

llvmlite/binding/__init__.py
gmarkall: Thanks - I think the old review comments will still be visible in the "conversation" tab, and I think I'm pretty happy with the resolution of the old comments, so feel free to rename it as soon as is convenient :slightly_smiling_face:

llvmlite/binding/newpassmangers.py
```diff
@@ -0,0 +1,305 @@
...
+    def size_level(self):
+        return self._size_level
+
+    @size_level.setter
+    def size_level(self, value):
+        if not 0 <= value <= 2:
+            raise ValueError("Optimization level for size should be 0, 1, or 2")
+        if value != 0 and self.speed_level != 2:
+            raise ValueError(
+                "Optimize for size should be encoded with speedup level == 2")
```
gmarkall, line 118: Maybe just a slight clarification of the message:
```suggestion
                "Optimization for size should be encoded with speed level == 2")
```

gmarkall (changes requested):
I accidentally started a review when I meant to add comments to the diff - no further comments now other than those on the diff.

gmarkall:
Notes on my most recent testing with Numba, using the branch https://github.com/gmarkall/numba/tree/npm:

* Tested with the Module Pass Manager replaced to prove the concept for this branch (not the Function Pass Manager as well).
* Test results:
   ```
   Ran 11970 tests in 1489.362s

   FAILED (failures=19, errors=3, skipped=607, expected failures=33)
   ```
* Failures are not any cause for concern - reasons for them include:
   * No refop pruning with the new pass manager
   * My quick hack broke some caching
   * Some debuginfo is transformed under the new pass manager in a way that is not anticipated by the tests (regexes not matching, etc.)

In conclusion, I think this provides a sufficient base to develop Numba support for the New Pass Manager.

gmarkall:
A quick further note related to my Numba testing branch - on some test runs there are no caching fails, in which case the results look like:

```
Ran 11970 tests in 1423.368s

FAILED (failures=8, errors=3, skipped=607, expected failures=33)
```

instead (i.e. 11 of the failures are from cache tests).

gmarkall (approved):
Thanks for all the efforts here! I think all review feedback is resolved now, and this provides a good base from which to start implementing NPM support in Numba.

(Note that the docs test is failing - it requires #1053)

gmarkall:
I just merged `main` to get the changes from #1053, which should turn CI green.

sklam:
BFID: `llvmlite_290`

sklam:
build passed

gmarkall:
@sklam Many thanks! Can this go to RTM?

sklam (approved):

PR #1051, opened 2024-05-13
Roundtrip llvmlite.binding.TypeRef back into llvmlite.ir.Type

sklam:
Adds `llvmlite.binding.TypeRef.as_ir()` to extract `llvmlite.ir.Type`. This is useful for building wrappers around functions emitted by other LLVM-IR producing tools. See new tests that uses output from clang on C++ sources.


sklam:
Example use: https://gist.github.com/sklam/a62f184725bc8c57ae039d7d2e66a481

stuartarchibald (approved):
Thanks for the patch and fixes!

PR #1053, opened 2024-05-30
Use RTD for all docs CI / testing (was: Try and get CI to use a newer Sphinx)

gmarkall:
~Some PRs (e.g. #1046)  have problems with the docs because of the old Sphinx version CI is using, the latest available in the defaults channel being 5.0.2. This PR switches to using conda-forge to get later versions, so that features such as `:no-index:` can be used.~

~From the CI logs:~

```
    sphinx-7.3.7               |     pyhd8ed1ab_0         1.3 MB  conda-forge
    sphinx_rtd_theme-2.0.0     |     pyha770c72_0         2.5 MB  conda-forge
```

**Update**: The new approach in this PR is to remove the docs testing on Azure and enable nitpicky mode in the Sphinx config. Now:

- The RTD build is checking the same things as the Azure docs build was
- There is no inconsistency between Sphinx versions in docs testing (since RTD is only running one version of Sphinx)

I purposely broke a reference in adf79ff, and then the RTD build failed with nitpicky mode was enabled, demonstrating that this is testing for broken references. I have subsequently reverted / fixed the broken reference in a later commit, and removed the Azure docs testing.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -67,11 +67,6 @@ jobs:
         CONDA_ENV: cienv
         WHEEL: 'yes'
 
-      py39_docs:
-        PYTHON: '3.9'
-        CONDA_ENV: cienv
-        BUILD_DOCS: 'yes'
-
       llvm15:
         PYTHON: '3.12'
         CONDA_ENV: cienv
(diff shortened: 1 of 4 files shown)


gmarkall:
Notes from triage meeting - we should make sure RTD is building with warnings as errors and then remove the py39_docs build.

esc (approved):

PR #1057, opened 2024-06-07
Port refprune pass to NewPassManager

yashssh:
Based on the changes introduced in #1042 by @modiking

### DIFF
--- a/docs/source/user-guide/binding/optimization-passes.rst
+++ b/docs/source/user-guide/binding/optimization-passes.rst
@@ -169,6 +169,11 @@ The ``add_*`` methods supported by both pass manager classes are:
    Add the `Simplify CFG
    <https://llvm.org/docs/Passes.html#simplifycfg-simplify-the-cfg>`_ pass.
 
+.. method:: add_refprune_pass()
+
+   Add the `Reference pruning
+   <https://github.com/numba/llvmlite/blob/main/ffi/custom_passes.cpp>`_ pass.
+
 .. currentmodule:: llvmlite.binding
 
 Legacy Pass Manager APIs
@@ -300,6 +305,11 @@ create and configure a :class:`PassManagerBuilder`.
    * .. function:: add_instruction_namer_pass()
 
         See `instnamer pass documentation <http://llvm.org/docs/Passes.html#instnamer-assign-names-to-anonymous-instructions>`_.
+
+   * .. function:: add_refprune_pass()
+
+     Add the `Reference pruning
+     <https://github.com/numba/llvmlite/blob/main/ffi/custom_passes.cpp>`_ pass.
 
 .. class:: ModulePassManager()
    :no-index:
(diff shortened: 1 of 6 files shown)


yashssh:
Please just review the last commit as it is rebased on top of #1046

dlee992:
I "checked" the last commit. LGTM. only has some minor format issues, but I feel don't need to bring it up. 
Also learned the difference of how to add a pass for old and new pass managers. Nice.

modiking:
LGTM

yashssh:
Rebased

yashssh:
Clang format does not report anything for me locally. What version of clang-format is used in the checks?

yashssh:
I missed to update `test_refprune.py` and there might be some complications there. Consider this "not yet ready for review" for now. 
Thanks :)

gmarkall:
> Clang format does not report anything for me locally. What version of clang-format is used in the checks?

It is clang-format-14, I think.

yashssh:
Gentle ping

yashssh (commented):
I have added some inline comments to make the thought process behind the the changes more apparent to the reviewers. 

The gist of the changes are:
1) We prefix the old refprune passes with legacy keyword.
2) We create a new refprune pass following the new pass manager syntax.
3) Now that we have multiple instances of the same pass, to prevent duplication of code we move the computation functions to a more global scope(struct) so that both passes can share them.

Found this article online which comes very close to explaining what we did here https://www.neilhenning.dev/posts/llvm-new-pass-manager/

Examples from llvm codebase on porting a pass to NPM 

- https://github.com/yashssh/llvm-project/commit/f5080194fda520cb67334165582aadfb413f3d22
- https://reviews.llvm.org/D22703

gmarkall (changes requested):
Thanks for the PR - the code changes look good, and I was able to follow them well from the info you provided / linked to. I think there are two things that need adding:

- Documentation for the `add_refprune_pass` functions of the new module and function pass managers
- Keep the tests for the legacy pass manager, as well as testing with the new pass manager. Although there's a low probability of issues with the legacy pass manager, the changes do touch their code (and other future changes might potentially as well, we never know) so keeping both paths in use tested is important.

yashssh:
Thanks for the review @gmarkall! I have addressed the comments.

gmarkall (dismissed):
Thanks for the additions - I think this is now good to merge!

esc:
@yashssh @gmarkall code needs a fix, it's become conflicting with `main` due to recent merges.  🙏

yashssh:
Resolved now, thanks!

esc:
> Resolved now, thanks!

@yashssh thank your for the swift update! -- @gmarkall can you check this and approve the fixes?

yashssh:
Ping

gmarkall (approved):
The fixup looks good to me.

esc:
Thank you for the patch!

PR #1063, opened 2024-06-24
Added llvm based process triple partitioning

kc611:
As titiled,

This PR aims to add a `llvm` based binding that uses the internal `Triple` object to partition the process triple string into it's respective parts.

Example:

```python
from llvmlite import binding

triple_string = binding.get_default_triple()
triple_string = "armv6hl-none-linux-gnueabi"

triple_parts = binding.get_triple_parts(triple_string)

print("Triple String:", triple_string)
print("Triple Parts:", triple_parts)
# Triple String: armv6hl-none-linux-gnueabi
# Triple Parts: ('arm', 'v6hl', 'unknown', 'linux', 'gnueabi', 'ELF')

```

### DIFF
--- a/ffi/targets.cpp
+++ b/ffi/targets.cpp
@@ -39,14 +39,22 @@ LLVMPY_GetTripleParts(const char *triple_str, const char **arch_out,
                       const char **subarch_out, const char **vendor_out,
                       const char **os_out, const char **environment_out,
                       const char **object_fmt_out) {
-    auto triple = llvm::Triple(triple_str);
-    *arch_out = LLVMPY_CreateString(triple.getArchName().data());
+    // Normalize the triple string
+    auto triple_str_norm = llvm::Triple::normalize(triple_str);
+    auto triple = llvm::Triple(triple_str_norm);
+
+    *arch_out = LLVMPY_CreateString(
+        llvm::Triple::getArchTypeName(triple.getArch()).data());
+
     // TODO: Change this to something that actually gets the subarch
     // Using temporary value to prevent non empty string issues
     *subarch_out = LLVMPY_CreateString(triple.getArchName().data());
-    *vendor_out = LLVMPY_CreateString(triple.getVendorName().data());
-    *os_out = LLVMPY_CreateString(triple.getOSName().data());
-    *environment_out = LLVMPY_CreateString(triple.getEnvironmentName().data());
+    *vendor_out = LLVMPY_CreateString(
+        llvm::Triple::getVendorTypeName(triple.getVendor()).data());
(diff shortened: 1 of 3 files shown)


gmarkall (changes requested):
The `_object_formats` dict in target.py has got out of sync with LLVM and is also missing some entries. It needs updating to:

```diff
diff --git a/llvmlite/binding/targets.py b/llvmlite/binding/targets.py
index 274eaae..6d63ad0 100644
--- a/llvmlite/binding/targets.py
+++ b/llvmlite/binding/targets.py
@@ -112,9 +112,13 @@ def get_host_cpu_name():
 
 
 _object_formats = {
+    0: "Unknown",
     1: "COFF",
     2: "ELF",
-    3: "MachO",
+    3: "GOFF",
+    4: "MachO",
+    5: "Wasm",
+    6: "XCOFF",
 }
 
 
```

otherwise this fails with certain triples, e.g.:

```python
In [7]: p("arm64_32-apple-ios")
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[7], line 1
----> 1 p("arm64_32-apple-ios")

Cell In[3], line 2, in p(x)
      1 def p(x):
----> 2     return ll.get_triple_parts(x)

File ~/numbadev/llvmlite/llvmlite/binding/targets.py:42, in get_triple_parts(triple)
     39         subarch = _str[len(arch):]
     40         break
     41 return Triple(arch, subarch, str(vendor), str(os),
---> 42               str(env), get_object_format(triple))

File ~/numbadev/llvmlite/llvmlite/binding/targets.py:130, in get_object_format(triple)
    128     triple = get_default_triple()
    129 res = ffi.lib.LLVMPY_GetTripleObjectFormat(_encode_string(triple))
--> 130 return _object_formats[res]

KeyError: 4
```
```

gmarkall:
Following the addition of the above change, I wonder if it's worth also adding tests that check with the XCOFF, Wasm, MachO, and GOFF cases too (there seem to be no tests in the LLVM test suite that result in an unknown object format).

gmarkall:
Source for the above object format enum is: https://github.com/llvm/llvm-project/blob/f28c006a5895fc0e329fe15fead81e37457cb1d1/llvm/include/llvm/ADT/Triple.h#L237-L246

gmarkall (changes requested):
Thanks for spotting that there were additions in LLVM 15 as well (I should have thought of this because I knew DXIL was added in it, but I didn't connect the dots - apologies!)

Just one minor change to resolve on the diff then I think this is good to mark RTM.

PR #1064, opened 2024-06-25
Add support for opaque pointers

rj-jesus:
This patch supports the transition of llvmlite from typed to opaque pointers, making the latter the default.

It also includes the following significant changes:
* Adds the option to fallback to typed pointers; in particular, if LLVMLITE_DISABLE_OPAQUE_POINTERS=1, typed pointers are used.
* Decouples pointee data from llvmlite pointers, moving it to a new class `_TypedPointerType`; objects of this class are created automatically from `PointerType`'s when pointee types are known in a way that's transparent to users.
* Adds explicit types to certain instructions (e.g., loads, gep, etc.) to enable phasing out pointee types on a case-by-case basis.


gmarkall:
I note that this works with LLVM 15 but not 14. Given the imminent move to LLVM 15, I think it's not worth fixing for 14.

rj-jesus:
Hi @gmarkall thanks, sounds good! I'll push a fix for the flake8 tests failing soon as well.

rj-jesus:
@sklam I believe this affects #1051. Do you have any thoughts on the changes in `llvmlite/ir/types.py` to `PointerType` to accommodate`from_llvm` on opaque pointers?

sklam:
> @sklam I believe this affects #1051. Do you have any thoughts on the changes in `llvmlite/ir/types.py` to `PointerType` to accommodate`from_llvm` on opaque pointers?

Looks good. The usecase is solely for getting function type back into a `ir.Type` and all of that is still working as expected.

I happen to notice that on LLVM14 parse_assembly do not like the opaque pointer despite it is enabled. Is that a LLVM bug?  

```bash
% python -m unittest  llvmlite.tests.test_binding.TestTypeRef.test_typeref_as_ir_for_wrappers_of_cpp_vector_struct
<string>:5:55: warning: ptr type is only supported in -opaque-pointers mode
declare void @"_Z3foo8Vector2DPS_"(<2 x float> %".1", ptr %".2")
                                                      ^
E
ERROR: test_typeref_as_ir_for_wrappers_of_cpp_vector_struct (llvmlite.tests.test_binding.TestTypeRef)
Exercise extracting C++ struct types that are passed as vectors.
  File "/path/to/llvmlite/tests/test_binding.py", line 2091, in test_typeref_as_ir_for_wrappers_of_cpp_vector_struct
    self._check_typeref_as_ir_for_wrappers(
  File "/path/to/llvmlite/tests/test_binding.py", line 2062, in _check_typeref_as_ir_for_wrappers
    new_mod = llvm.parse_assembly(str(wrapper_mod))
  File "/path/to/llvmlite/binding/module.py", line 25, in parse_assembly
<string>:5:55: error: expected type
declare void @"_Z3foo8Vector2DPS_"(<2 x float> %".1", ptr %".2")
```

sklam:
Let's wait for LLVM15 becoming default

rj-jesus:
That sounds great, do you have an estimate of when that will be?

llvmlite/__init__.py
```diff
@@ -1,3 +1,10 @@
 from ._version import get_versions
 __version__ = get_versions()['version']
 del get_versions
+
+# FIXME: Remove me once typed pointers are no longer supported.
+def _get_disable_opaque_pointers():
+  import os
+  return os.environ.get('LLVMLITE_DISABLE_OPAQUE_POINTERS', '0') == '1'
+_disable_opaque_pointers = _get_disable_opaque_pointers()
+del _get_disable_opaque_pointers
```
gmarkall, line 10: If this global is to be used from outside the module, then it should not be implied to be private by the leading underscore. Also I think the inversion of logic that this then implies elsewhere in the code could be simplified by making it positively named:
```suggestion
def _opaque_pointers_enabled():
  import os
  return os.environ.get('LLVMLITE_DISABLE_OPAQUE_POINTERS', '0') != '1'
opaque_pointers_enabled = _opaque_pointers_enabled()
del _opaque_pointers_enabled
```

llvmlite/binding/context.py
```diff
@@ -1,12 +1,18 @@
...
 
+# FIXME: Remove me once typed pointers are no longer supported.
+from llvmlite import _disable_opaque_pointers
+from ctypes import c_bool
+
 
 def create_context():
-    return ContextRef(ffi.lib.LLVMPY_ContextCreate())
+    return ContextRef(
+        ffi.lib.LLVMPY_ContextCreate(not _disable_opaque_pointers))
```
gmarkall, line 10: In line with the changed name / semantics mentioned above (needs applying throughout the uses within this file):

```suggestion
        ffi.lib.LLVMPY_ContextCreate(opaque_pointers_enabled))
```

llvmlite/binding/targets.py
```diff
@@ -192,23 +192,11 @@ def get_element_offset(self, ty, position):
                              "type?".format(position, str(ty)))
         return offset
 
-    def get_pointee_abi_size(self, ty):
```
gmarkall, line 195: Is this removal a breaking change? I see Numba doesn't use it, but it is documented: https://llvmlite.readthedocs.io/en/latest/user-guide/binding/target-information.html#llvmlite.binding.TargetData.get_pointee_abi_size

Can / should this be kept, but only usable when typed pointers are enabled, then removed later when typed pointer support is removed?

gmarkall (changes requested):
I've had a look over this and experimented locally with it (not encountering any issues so far in doing that). I've left a couple of comments on the diff, and will have another pass following discussing them.

ffi/value.cpp
```diff
@@ -404,11 +404,6 @@ LLVMPY_SetValueName(LLVMValueRef Val, const char *Name) {
 API_EXPORT(LLVMModuleRef)
 LLVMPY_GetGlobalParent(LLVMValueRef Val) { return LLVMGetGlobalParent(Val); }
 
-API_EXPORT(LLVMTypeRef)
```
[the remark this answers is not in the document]
rj-jesus: I believe this was introduced with #1051, and when I rebased my patch on top of it I thought it would make sense to keep `LLVMPY_GlobalGetValueType` in `type.cpp` next to other similar functions like `LLVMPY_GetElementType` and `LLVMPY_GetReturnType`. But it makes no odds to me, I'm happy to leave it in `value.cpp`.

llvmlite/__init__.py
rj-jesus: > If this global is to be used from outside the module, then it should not be implied to be private by the leading underscore

My intention was to keep the global private to the package (`llvmlite`). In other words, I didn't want to introduce a global symbol that other packages could think they could use to enable/disable opaque pointers, as (at the time at least) I thought this should only be done through an environment flag. But I'm happy to change it if you think otherwise.

> I think the inversion of logic that this then implies elsewhere in the code could be simplified

Sure, that sounds good, I'll work on changing this.

llvmlite/binding/targets.py
rj-jesus: That's a good point, maybe it should be kept. If we keep it, it will definitely only work in typed pointers mode though.

Most use cases for it I can think of are covered with `get_abi_size` and `get_abi_alignment`, so I don't think it's vitally important, but I'll have a look and if we can keep it in typed pointers mode, we might as well.

rj-jesus:
Hi @gmarkall, thanks for having had a look! I'll work on the proposed changes and update the review soon. In the meantime, I've also left a few comments in reply to yours. If you can, please let me know what you think and I'll include any further changes in the next update.

rj-jesus:
Hi @gmarkall, I believe I've addressed most of your comments in the last commit, in particular:
* Change logic `_disable_opaque_pointers` → `opaque_pointers_enabled`
* Remove leading underscore from `opaque_pointers_enabled`
* Move `LLVMPY_GlobalGetValueType` back to `value.cpp`
* Reinstate `LLVMPY_ABISizeOfElementType` and `LLVMPY_ABIAlignmentOfElementType` with error if used in opaque pointer mode

Please let me know if you have any other comments!

llvmlite/binding/targets.py
rj-jesus: I've put this back with a runtime error if opaque pointers are enabled. This will likely have to be removed once we upgrade to the next versions of LLVM though.

ffi/value.cpp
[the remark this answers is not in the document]
rj-jesus: I've moved this back to `value.cpp`, I don't think we need to make this change in this patch.

llvmlite/__init__.py
rj-jesus: I've removed the leading underscore and inverted the logic---please let me know what you think!

gmarkall (changes requested):
Thanks for the updates! I think this is getting there, but I have left a few comments on the diff.

To summarise a few points from our discussions relevant to the continuation of this PR:

1. I will push changes to update the CI setup to test with opaque pointers enabled and disabled.
2. I will push changes to add some documentation (this won't need to be a long section, I think)
3. This PR enables a transitional state where we're supporting a lot of the typed pointer APIs even if we're using opaque pointers in the underlying LLVM binding. This means we can enable opaque pointer usage in Numba without making a large set of changes to it. Once we have confidence in the use of opaque pointers, we can then modify all its API usage to only use the opaque pointer variant of the APIs, allowing the eventual removal of the typed mode in llvmlite. This gives us a simple, gradual, and safe transition path for Numba.

Many thanks for your efforts so far!

llvmlite/ir/types.py
```diff
@@ -109,30 +112,67 @@ def _to_string(self):
...
     is_pointer = True
     null = 'null'
 
-    def __init__(self, pointee, addrspace=0):
-        assert not isinstance(pointee, VoidType)
-        self.pointee = pointee
+    # Factory to create typed or opaque pointers based on `pointee'.
+    def __new__(cls, pointee=None, addrspace=0):
+        if cls is PointerType and pointee is not None:
+            return _TypedPointerType(pointee, addrspace)
```
gmarkall, line 124: I made a mistake in my suggestion below for the double-`__init__` issue - I think it should have been this:
```suggestion
            return super().__new__(_TypedPointerType)
```

llvmlite/ir/types.py
```diff
@@ -109,30 +112,67 @@ def _to_string(self):
...
+    def __hash__(self):
+        return hash(PointerType)
+
+    @classmethod
+    def from_llvm(cls, typeref, ir_ctx):
+        """
+        Create from a llvmlite.binding.TypeRef
+        """
+        if not opaque_pointers_enabled:
+            return _TypedPointerType.from_llvm(typeref, ir_ctx)
```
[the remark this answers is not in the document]
gmarkall: This suggestion was a mistake - see https://github.com/numba/llvmlite/pull/1064/files#r1699988799

llvmlite/ir/instructions.py
```diff
@@ -431,9 +431,18 @@ def descr(self, buf):
...
+        if typ is not None:
+            pass
+        elif isinstance(ptr, AllocaInstr):
+            typ = ptr.allocated_type
+        # For compatibility with typed pointers. Eventually this should
+        # probably be removed (when typed pointers are fully removed).
+        elif not ptr.type.is_opaque:
+            typ = ptr.type.pointee
+        else:
+            raise ValueError("Load lacks type.")
```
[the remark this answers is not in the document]
rj-jesus: The idea was to keep it somewhat consistent with what's done in `GEPInstr`, but you're right, it might be more confusing than it's worth so I've changed it.

llvmlite/ir/types.py
rj-jesus: Thanks very much for noting this - I've updated it!

rj-jesus:
Hi @gmarkall, thanks very much for your comments, I believe I've addressed them in the latest revision, in particular:

* We now default to opaque pointers disabled
* Moved a few imports to the top-level
* Refactored `if` logic in a few constructors
* Fixed double-init issue you pointed out

Please let me know if you have any other comments.

llvmlite/ir/instructions.py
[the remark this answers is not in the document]
gmarkall: Thanks - I hadn't spotted the commonality with `GEPInstr` - I think the simplified form is easier to understand, so thanks for making the change.

gmarkall:
@rj-jesus Thanks for this, I think it looks good - I'll update the CI configuration and docs.

rj-jesus:
@gmarkall Thank you very much for the feedback! Please let me know if you need help with anything.

gmarkall (approved):
I think the code changes here are good. I've marked as "waiting on second reviewer" because I've added:

- CI setup that runs with opaque pointers enabled
- Documentation

that could do with a check from someone else.

rj-jesus (commented):
Hi @gmarkall, Thanks for these commits, I've had a look at the documentation and everything checks out, I've just left a minor comment. Please let me know if you need anything else.

sklam (approved):
Docs and CI looks good

esc:
thanks for the patch

PR #1068, opened 2024-07-18
Early exit from fpm.run() if called on function declarations

yashssh:
This pattern in common in Numba which causes fpm to crash

### DIFF
--- a/ffi/newpassmanagers.cpp
+++ b/ffi/newpassmanagers.cpp
@@ -132,6 +132,10 @@ LLVMPY_RunNewFunctionPassManager(LLVMFunctionPassManagerRef FPMRef,
     PassBuilder *PB = llvm::unwrap(PBRef);
     Function *F = reinterpret_cast<Function *>(FRef);
 
+    // Don't try to optimize function declarations
+    if (F->isDeclaration())
+        return;
+
     LoopAnalysisManager LAM;
     FunctionAnalysisManager FAM;
     CGSCCAnalysisManager CGAM;
--- a/llvmlite/tests/test_binding.py
+++ b/llvmlite/tests/test_binding.py
@@ -452,6 +452,9 @@ declare void @a_readonly_func(i8 *) readonly
 declare i8* @a_arg0_return_func(i8* returned, i32*)
 """
 
+asm_declaration = r"""
+declare void @test_declare(i32* )
+"""
 
 # This produces the following output from objdump:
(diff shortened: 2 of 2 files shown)


esc:
Failed CI restarted. Looking at the log it looks like it failed to setup the testing environment as some packages could not be installed. Assuming it was a network glitch.

gmarkall (commented):
Thanks for the PR! I can observe also that running the NPM on a declaration causes a crash, and that this PR fixes it. 

However, I have a quick question - should we instead raise an exception if one attempts to optimize a declaration, then detect / remove this erroneous behaviour in Numba instead?

yashssh:
Thanks for the review @gmarkall! 

>should we instead raise an exception if one attempts to optimize a declaration, then detect / remove this erroneous behaviour in Numba instead?

I'm not sure on raising an exception as I think it's not entirely incorrect to call optimization on function declarations. The legacy pass manager allows it, module.functions ends up returning these declarations as functions as well.

I agree with the Numba part, it would be good to have an extra check there.

gmarkall:
> I'm not sure on raising an exception as I think it's not entirely incorrect to call optimization on function declarations. 

I'm not following why - if it leads to a segfault, isn't it entirely incorrect?

> The legacy pass manager allows it, module.functions ends up returning these declarations as functions as well.

Is Numba trying to optimize declarations because it iterates over `module.functions` and optimizes each one?

yashssh:
> I'm not following why - if it leads to a segfault, isn't it entirely incorrect?

I meant the program shouldn't crash, for eg https://godbolt.org/z/d6z196hhW

> Is Numba trying to optimize declarations because it iterates over module.functions and optimizes each one?

Yes

yashssh:
Adding more to my reasoning, this is the snippet from `LegacyPassManager.cpp`
```
bool FPPassManager::runOnFunction(Function &F) {
  if (F.isDeclaration())
    return false;
```
Which tells it is the pass managers's job to not try to optimize if it's a declaration rather than relying on user to perform the check. I can't seem to find where this check should-be/should-have-been in the new pass manager code, so I think it makes sense to perform this check before calling the RunNewFunctionPassManager as it is not getting handled automatically like the legacy PM.

gmarkall (approved):
Thanks for the additional info - I see now that this is inline with llvmlite's implementation for the Legacy Pass Manager.

PR #1069, opened 2024-07-18
Add instrumentation callback hook for new pass managers

yashssh:
The new pass manager relies on PIC callbacks for deciding
if a function needs to be skipped for optimization when
compiled with 'optnone'.

### DIFF
--- a/ffi/newpassmanagers.cpp
+++ b/ffi/newpassmanagers.cpp
@@ -80,11 +80,12 @@ LLVMPY_RunNewModulePassManager(LLVMModulePassManagerRef MPMRef,
     StandardInstrumentations SI(DebugLogging, VerifyEach, PrintPassOpts);
     SI.registerCallbacks(PIC, &FAM);
     PassBuilder PB(TM, *PTO, None, &PIC);
-# else
-    StandardInstrumentations SI(M->getContext(), DebugLogging, VerifyEach, PrintPassOpts);
+#else
+    StandardInstrumentations SI(M->getContext(), DebugLogging, VerifyEach,
+                                PrintPassOpts);
     SI.registerCallbacks(PIC, &FAM);
     PassBuilder PB(TM, *PTO, std::nullopt, &PIC);
-# endif
+#endif
 
     PB.registerLoopAnalyses(LAM);
     PB.registerFunctionAnalyses(FAM);
@@ -178,8 +179,9 @@ LLVMPY_RunNewFunctionPassManager(LLVMFunctionPassManagerRef FPMRef,
     StandardInstrumentations SI(DebugLogging, VerifyEach, PrintPassOpts);
     SI.registerCallbacks(PIC, &FAM);
     PassBuilder PB(TM, *PTO, None, &PIC);
-# else
-    StandardInstrumentations SI(F->getContext(), DebugLogging, VerifyEach, PrintPassOpts);
+#else
+    StandardInstrumentations SI(F->getContext(), DebugLogging, VerifyEach,
+                                PrintPassOpts);
     SI.registerCallbacks(PIC, &FAM);
     PassBuilder PB(TM, *PTO, std::nullopt, &PIC);
 #endif
--- a/ffi/newpassmanagers.cpp
+++ b/ffi/newpassmanagers.cpp
@@ -1,7 +1,6 @@
 #include "core.h"
 #include "llvm-c/TargetMachine.h"
 #include "llvm/Analysis/AliasAnalysisEvaluator.h"
-#include "llvm/IR/LLVMContext.h"
 #include "llvm/IR/PassManager.h"
 #include "llvm/IR/Verifier.h"
 #include "llvm/Passes/PassBuilder.h"
--- a/ffi/newpassmanagers.cpp
+++ b/ffi/newpassmanagers.cpp
@@ -38,6 +38,11 @@ typedef OpaquePipelineTuningOptions *LLVMPipelineTuningOptionsRef;
 DEFINE_SIMPLE_CONVERSION_FUNCTIONS(PipelineTuningOptions,
                                    LLVMPipelineTuningOptionsRef)
 
+struct OpaquePassInstrumentationCallbacks;
+typedef OpaquePassInstrumentationCallbacks *LLVMPassInstrumentationCallbacksRef;
+DEFINE_SIMPLE_CONVERSION_FUNCTIONS(PassInstrumentationCallbacks,
+                                   LLVMPassInstrumentationCallbacksRef)
+
 static TargetMachine *unwrap(LLVMTargetMachineRef P) {
     return reinterpret_cast<TargetMachine *>(P);
 }
@@ -56,13 +61,13 @@ LLVMPY_CreateNewModulePassManager() {
 API_EXPORT(void)
 LLVMPY_RunNewModulePassManager(LLVMModulePassManagerRef MPMRef,
                                LLVMModuleRef mod,
-                               LLVMPipelineTuningOptionsRef PTORef,
-                               LLVMTargetMachineRef TMRef) {
+                               LLVMPassBuilderRef PBRef,
+                               LLVMPassInstrumentationCallbacksRef PICRef) {
 
     ModulePassManager *MPM = llvm::unwrap(MPMRef);
     Module *M = llvm::unwrap(mod);
-    PipelineTuningOptions *PTO = llvm::unwrap(PTORef);
-    TargetMachine *TM = llvm::unwrap(TMRef);
+    PassBuilder *PB = llvm::unwrap(PBRef);
+    PassInstrumentationCallbacks *PIC = llvm::unwrap(PICRef);
 
     // TODO: Make these set(able) by user
     bool DebugLogging = false;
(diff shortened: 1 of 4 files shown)


esc:
Looks like the C++ is failing to compile in this PR at present.

yashssh:
It's a llvm16 specific failure, I can see it offline as well. Looking into it

llvmlite/tests/test_binding.py
```diff
@@ -3075,6 +3141,36 @@ def test_run(self):
         self.assertIn("%.4", orig_asm)
         self.assertNotIn("%.4", optimized_asm)
 
+    def test_run_O0(self):
```
gmarkall, line 3144: I think all test names are snake case with no capitals:
```suggestion
    def test_run_o0(self):
```

Also I'd be inclined to factor out the run functionality into a separate function like:

```python
    def run_o_n(self, level):
        pb = self.pb(level)
        mod = self.module()
        fun = mod.get_function("sum")
        orig_asm = str(fun)
        fpm = pb.getFunctionPassManager()
        fpm.run(fun, pb)
        optimized_asm = str(fun)
        return orig_asm, optimized_asm
```

then use it as in the following comment's suggestion...

llvmlite/tests/test_binding.py
```diff
@@ -3075,6 +3141,36 @@ def test_run(self):
...
+        pb = self.pb(0)
+        mod = self.module()
+        fun = mod.get_function("sum")
+        orig_asm = str(fun)
+        fpm = pb.getFunctionPassManager()
+        fpm.run(fun, pb)
+        optimized_asm = str(fun)
```
gmarkall, line 3151: (following from above comment)
```suggestion
        orig_asm, optimized_asm = self.run_o_n(0)
```

... so that the difference between the O3 and O0 runs is a little easier to spot on inspection (either the presence or absence of `"%.4"` in the optimized assembly)

llvmlite/tests/test_binding.py
```diff
@@ -3064,7 +3130,7 @@ def test_close(self):
         fpm = self.pm()
         fpm.close()
 
-    def test_run(self):
+    def test_run_O3(self):
```
gmarkall, line 3133: Changing case as suggestion below:

```suggestion
    def test_run_o3(self):
```

Also this could potentially use the `run_o_n()` function suggested below.

llvmlite/tests/test_binding.py
```diff
@@ -3075,6 +3141,36 @@ def test_run(self):
...
+        fun = mod.get_function("sum")
+        orig_asm = str(fun)
+        fpm = pb.getFunctionPassManager()
+        fpm.run(fun, pb)
+        optimized_asm = str(fun)
+        self.assertIn("%.4", orig_asm)
+        self.assertIn("%.4", optimized_asm)
+
+    def test_optnone(self):
+        # Function shouldn't be optimized if the function as `optnone` attached
```
gmarkall, line 3156:
```suggestion
        # Function shouldn't be optimized if the function has `optnone` attached
```

llvmlite/tests/test_binding.py
```diff
@@ -452,6 +452,43 @@ def no_de_locale():
...
+    store double %j, double* %J
+    %t1 = load i32, i32* %I		; <i32> [#uses=1]
+    %t2 = add i32 %t1, 1		; <i32> [#uses=1]
+    store i32 %t2, i32* %I
+    %t3 = load i32, i32* %I		; <i32> [#uses=1]
+    %t4 = sitofp i32 %t3 to double		; <double> [#uses=1]
+    %t5 = load double, double* %J		; <double> [#uses=1]
+    %t6 = fmul double %t4, %t5		; <double> [#uses=1]
+    ret double %t6
+}
```
gmarkall, line 486: Since these are identical apart from the `optnone` attribute, could you generate `asm_alloca_no_optnone` with something like:

```python
asm_alloca_no_optnone = asm_alloca_optnone.replace("optnone ", "")
```

to save duplicating all the code? (I had to inspect it for a while to make sure that was the only difference)

gmarkall (changes requested):
Thanks for the PR! I can verify locally that there is an issue here and that the implementation resolves it.

However the changes do something odd to the abstractions and design between the FFI and Python layers - we create a PassBuilder in order to get a ModulePassManager, but for running the module pass manager, we end up ignoring the PassBuilder we did have and creating a new one. My gut feeling is that we should be exposing the PassInstrumentationCallbacks and PrintPassOptions as Python classes too, which would allow us to construct the PassBuilder in the way that we require, so it can be passed into `LLVMPY_RunNewModulePassManager` ready-to-use (and potentially therefore not require any actual changes to that function, but instead the changes being new functions in FFI to help expose the PIC and PPO classes and the corresponding Python changes).

Do you agree with that evaluation? If so, would the suggested changes tidy things up?

I've left a few small comments on the tests on the diff.

llvmlite/tests/test_binding.py
yashssh: Ahh right that would be better, sorry for your lost time :)

llvmlite/binding/newpassmanagers.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, ptr=None):
         super().__init__(ptr)
 
     def run(self, module, pb):
-        ffi.lib.LLVMPY_RunNewModulePassManager(self, pb, module)
+        ffi.lib.LLVMPY_RunNewModulePassManager(self, module, pb, pb._pic)
```
gmarkall, line 29: I had previously not spotted that the arguments to `run()` were reversed compared to how they're passed into `LLVMPY_RunNewModulePassManager()` - making them consistent is a nice clean-up.

gmarkall: However, one thing I don't quite get here - can the `PassBuilder._pic` object ever be out-of-sync with the `PIC` object attached to the underlying `PassBuilder` instance? I'm thinking: do we need to pass in `pb._pic` here, or could `LLVMPY_RunNewModulePassManager` just obtain it from `PB.getPassInstrumentationCallbacks()`?

ffi/newpassmanagers.cpp
```diff
@@ -53,16 +59,33 @@ LLVMPY_CreateNewModulePassManager() {
...
+
+    PrintPassOptions PrintPassOpts;
+
+#if LLVM_VERSION_MAJOR < 16
+    StandardInstrumentations SI(DebugLogging, VerifyEach, PrintPassOpts);
+#else
+    StandardInstrumentations SI(M->getContext(), DebugLogging, VerifyEach,
+                                PrintPassOpts);
+#endif
+    SI.registerCallbacks(*PIC, &FAM);
```
gmarkall, line 87: Could this be:
```suggestion
    SI.registerCallbacks(PB.getPassInstrumentationCallbacks(), &FAM);
```

then this function not have to accept and unwrap a `PICRef`, as per comment below?

llvmlite/tests/test_binding.py
```diff
@@ -3020,20 +3040,29 @@ class TestNewModulePassManager(BaseTest, NewPassManagerMixin):
...
-    def test_close(self):
-        mpm = self.pm()
-        mpm.close()
-
-    def test_run(self):
-        pb = self.pb(speed_level=3, size_level=0)
+    def run_o_n(self, level):
         mod = self.module()
-        orig_asm = str(mod)
+        orig_asm = str(self.module())
```
gmarkall, line 3045: Does this not work if it's `str(mod)` for some reason I've missed?

llvmlite/tests/test_binding.py
```diff
@@ -3045,6 +3074,25 @@ def test_instcombine(self):
         self.assertIn("%.3", orig_asm)
         self.assertNotIn("%.3", optimized_asm)
 
+    def test_optnone(self):
+        # Module shouldn't be optimized if the function has `optnone` attached
+        pb = self.pb(speed_level=3, size_level=0)
+        orig_asm_no_optnone = str(asm_alloca_optnone.replace("optnone ", ""))
```
gmarkall, line 3080: A nit, but I'd be inclined to just call this `orig_asm` since it's the same as `orig_asm_optnone` without the `optnone`:

```suggestion
        orig_asm = str(asm_alloca_optnone.replace("optnone ", ""))
```

gmarkall (changes requested):
The revised design looks simpler than before - thanks for looking at making these changes.

I have no problem with PIC being exposed in Python, but I wonder if it wasn't actually necessary because we create the PassBuilder with the PIC, and presumably the PassBuilder can give a reference to that PIC at any time we need it anyway?

I have left a couple of questions on the diff.

yashssh:
> I have no problem with PIC being exposed in Python, but I wonder if it wasn't actually necessary because we create the PassBuilder with the PIC, and presumably the PassBuilder can give a reference to that PIC at any time we need it anyway?

Yes, you are right. I have updated the code.

llvmlite/tests/test_binding.py
yashssh: Fixed it.

gmarkall (dismissed):
Thanks for the updates - this simplifies things!

Looks good to me.

kc611 (approved):

esc (approved):

kc611 (dismissed):

yashssh:
Does this needs rebasing?

esc:
@yashssh I am not sure, we tried to merge this, but it seems like our approvals keep getting dismissed automatically? I've never seen this before. Do you have any idea what might be going on?

esc:
Is this a GitHub bug?

https://github.com/orgs/community/discussions/58535

yashssh:
Closing and reopening seems to have worked for some, I can try that?

esc:
> Closing and reopening seems to have worked for some, I can try that?

by all means, yes please

gmarkall:
This sounds like the potential scenario that triggered it here: https://github.com/orgs/community/discussions/58535#discussioncomment-6495938 - I think this started based on #1068?

esc (approved):

yashssh:
I can see all the commits that were part of the PR in the commit history on main branch, is it common practice?

esc:
> I can see all the commits that were part of the PR in the commit history on main branch, is it common practice?

In `git` there is no such concept as a commit being "on a branch". The commit history is a DAG and branches are pointers at specific nodes in the graph. When a PR is merged to `main` all the commits of the PR are reachable from the `main` branch, that's the whole point. Makes sense?

yashssh:
Ahh yes.. thank you!
________________________________

esc:
> Ahh yes.. thank you!

Here, perhaps this can aid comprehension: https://fossy-cats.github.io/Git-Buch_EN/Git-Book_4.html (it's an english translation of a git book I wrote more than a decade ago -- the concepts are still very relevant today.)

yashssh:
> Here, perhaps this can aid comprehension: https://fossy-cats.github.io/Git-Buch_EN/Git-Book_4.html (it's an english translation of a git book I wrote more than a decade ago -- the concepts are still very relevant today.)

Will give it a read, thanks again!

PR #1073, opened 2024-08-12
bump compiler for linux-aarch64 to 11

esc:
Numba team build system for linux-aarch64 is being updated from Miniforge3 to miniconda3 based docker containers. `llvmlite` fails to compile from `main` with current recipe specification as compiler version 9 is not available for that architecture. We bump to 11 which is the default as referenced by:

https://github.com/AnacondaRecipes/aggregate/blob/master/conda_build_config.yaml

For reference, the error was:

```
The following packages are incompatible
├─ gcc_linux-aarch64 9.*  does not exist (perhaps a typo or a missing channel);
└─ gxx_linux-aarch64 9.*  does not exist (perhaps a typo or a missing channel).
```

### DIFF
--- a/conda-recipes/llvmdev/conda_build_config.yaml
+++ b/conda-recipes/llvmdev/conda_build_config.yaml
@@ -1,15 +1,15 @@
 # Numba/llvmlite stack needs an older compiler for backwards compatability.
 c_compiler_version:         # [linux]
   - 7                       # [linux and (x86_64 or ppc64le)]
-  - 9                       # [linux and aarch64]
+  - 11                      # [linux and aarch64]
 
 cxx_compiler_version:       # [linux]
   - 7                       # [linux and (x86_64 or ppc64le)]
-  - 9                       # [linux and aarch64]
+  - 11                      # [linux and aarch64]
 
 fortran_compiler_version:   # [linux]
   - 7                       # [linux and (x86_64 or ppc64le)]
-  - 9                       # [linux and aarch64]
+  - 11                      # [linux and aarch64]
 
 c_compiler:              # [win]
   - vs2019               # [win]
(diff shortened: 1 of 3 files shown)


gmarkall:
Why do we use 7 for Linux on non-AArch64? I see the comment in `conda_build_config.yaml` says:

```
Numba/llvmlite stack needs an older compiler for backwards compatability.
```

but if we can bump to 11 on AArch64, does that mean that the above comment is no longer relevant?

esc:
> Why do we use 7 for Linux on non-AArch64? I see the comment in `conda_build_config.yaml` says:
> 
> ```
> Numba/llvmlite stack needs an older compiler for backwards compatability.
> ```
> 
> but if we can bump to 11 on AArch64, does that mean that the above comment is no longer relevant?

I did wonder about that too, I looked at the Anaconda feedstock here:

https://github.com/AnacondaRecipes/llvmlite-feedstock/tree/master/recipe

and it seems like they did away withe the `cbc` file altogether...

esc:
Also, we can not compile `llvmlite` with only the change to `cbc` file. It turns out that the `llvmdev` that we use was also built with `Miniforge3` compilers. Running this PR as is with just b09b4cd results in:

```
- The following package could not be installed
- └─ llvmdev 15.*  is not installable because it requires
- └─ libgcc-ng >=12 , which does not exist (perhaps a missing channel).
```

😢

esc:
`llvmdev` has been fresh `build 2` variants on anaconda.org under both `numba/label/main` and `numba/label/ci` (except for `ppc64le` for which no new `continuumio/miniconda3` contains are available because that platform is no longer supported (last release for `ppc64le1 was https://hub.docker.com/layers/continuumio/miniconda3/23.10.0-1/images/sha256-166ff37fba6c25fcad8516aa5481a2a8dfde11370f81b245c1e2e8002e68bcce?context=explore)

Screenshot 2024-08-14 at 11 20 47

esc:
With the `build 2` variant this PR passes all the `llvmlite` tests on the Anaconda build farm!

sklam (approved):

PR #1074, opened 2024-08-26
Add NetBSD support.

0-wiz-0:

PR #1077, opened 2024-09-15
Add NetBSD support.

0-wiz-0:
Second try of #1074.


gmarkall (approved):
I've not tested this on NetBSD, but I'm assuming it works for you :slightly_smiling_face: 

This looks good to me.

esc:
thanks for the patch

PR #1081, opened 2024-09-26
Add convenience func and args

thomaspinckney3:
I found these helpful in a recent project. Not sure if there is a better way to do this. This is my first pull request so hopefully getting it right!

types.py: add get_element_offset to Type
values.py: add packed arg to literal_struct
context: add packed arg to get_identified_type
tests: add test cases

### DIFF
--- a/docs/source/user-guide/ir/types.rst
+++ b/docs/source/user-guide/ir/types.rst
@@ -38,10 +38,16 @@ instantiated, a type should be considered immutable.
         to the *target_data*---an
         :class:`llvmlite.binding.TargetData` instance.
 
-        NOTE: :meth:`get_abi_size` and :meth:`get_abi_alignment`
-        call into the LLVM C++ API to get the requested
-        information.
+   * .. method:: get_element_offset(target_data, position)
 
+        Get the byte offset for the struct element at position,
+        according to the *target_data*---an
+        :class:`llvmlite.binding.TargetData` instance.
+
+        NOTE: :meth:`get_abi_size`, :meth:`get_abi_alignment`,
+        and :meth:`get_element_offset` call into the LLVM C++
+        API to get the requested information.
+        
    * .. method:: __call__(value)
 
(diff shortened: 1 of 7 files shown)


gmarkall:
Thanks for the PR! I need to look over the code more closely, but I think the idea of these is good in general.

docs/source/user-guide/ir/types.rst
```diff
@@ -38,10 +38,16 @@ instantiated, a type should be considered immutable.
         to the *target_data*---an
         :class:`llvmlite.binding.TargetData` instance.
 
-        NOTE: :meth:`get_abi_size` and :meth:`get_abi_alignment`
-        call into the LLVM C++ API to get the requested
-        information.
+   * .. method:: get_element_offset(target_data, position)
 
+        Get the byte offset for the struct element at position,
```
gmarkall, line 43: I think this would be in keeping with the style of these methods:
```suggestion
        Get the byte offset for the struct element at *position*,
```

llvmlite/tests/test_binding.py
```diff
@@ -2701,6 +2701,14 @@ def test_literal_struct(self):
             # Also test constant text repr
             gv.initializer = ir.Constant(typ, [1])
 
+        # Packed layout2
```
gmarkall, line 2704: I think this is the intention of this added part of the test:
```suggestion
        # Packed layout created from Constant.literal_struct
```

gmarkall (changes requested):
Thanks for the additions - I think this all looks good, with a couple of tiny fixups on the diff - I think it's good to go apart from that!

gmarkall:
I've added this to the next milestone since it seems like it should be pretty easy to get in before the next release.

gmarkall (approved):

gmarkall:
Many thanks for your contribution, @thomaspinckney3! This will be included in the next release of llvmlite (v0.44.0).

PR #1082, opened 2024-10-07
Enable opaque pointers

rj-jesus:
This patch turns opaque pointers on by default.


rj-jesus:
I'm closing this as superseded by #1159 and #1160.

PR #1086, opened 2024-10-09
Move Azure to use macos-13

gmarkall:
macos-12 is deprecated and will be removed soon.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -3,7 +3,7 @@ jobs:
 - template: buildscripts/azure/azure-linux-macos.yml
   parameters:
     name: macOS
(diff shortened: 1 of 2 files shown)


stuartarchibald:
I think this might be related to 
https://github.com/numba/numba/pull/7650
and
https://github.com/numba/numba/issues/7653

stuartarchibald:
RE ^ it seems like we previously decided that this wasn't the problem: https://github.com/numba/llvmlite/pull/786#issuecomment-1005912362 ?! Maybe it now is on a new OSX baseline?

gmarkall:
Looks like #1097 will be the way forward here.

PR #1087, opened 2024-10-14
Add missing argtypes description for refprune pass related APIs

yashssh:
These missing descriptions might be causing refprune related windows tests failure after https://github.com/numba/numba/pull/9676


gmarkall (approved):
This looks to me like it correctly adds the missing argtypes for these functions, and I can see how their omission could cause a segfault.

PR #1098, opened 2024-11-10
Fix store to opaque ptr

alexander-shaposhnikov:
If ptr.type.is_opaque is True it doesn't need to have ptr.type.pointee attribute.
Add a unit test.

### DIFF
--- a/llvmlite/ir/builder.py
+++ b/llvmlite/ir/builder.py
@@ -774,7 +774,7 @@ class IRBuilder(object):
         if not isinstance(ptr.type, types.PointerType):
             msg = "cannot store to value of type %s (%r): not a pointer"
             raise TypeError(msg % (ptr.type, str(ptr)))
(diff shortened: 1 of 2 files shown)


rj-jesus:
@sklam Yep, sure.
Thanks for the patch, @alexander-shaposhnikov.
In general it LGTM, my only comment would be that the new test `test_store` could maybe be part of `test_mem_ops`, but I don't think this is critical.

alexander-shaposhnikov:
Many thanks, I've updated the PR.

rj-jesus (commented):
Thanks very much, LGTM!
(@sklam I can't approve it myself.)

alexander-shaposhnikov:
@sklam (would be amazing if you could merge this in, many thanks in advance)

gmarkall (approved):

PR #1103, opened 2024-11-26
Add target attributes to LLVM IR Module getting optimized

yashssh:
LLVM IR that is passed down from Numba frontend to LLVM backend does not contain some basic target specific attributes(`target-cpu`, `target-features`) which is added by Clang frontend (when compiling C++) or either by tools such as `opt` and `llc` when optimizing the LLVM-IR directly with the help of `cli` flags. 

I have mimicked the flow of `opt` optimizer from llvm, which adds these details to the IR module before running the optimization pipeline. 

This change shouldn't affect any functionality and correctness, it's just to add more details to the IR involved and to make it more more verbose and portable across tools.

### DIFF
--- a/ffi/newpassmanagers.cpp
+++ b/ffi/newpassmanagers.cpp
@@ -52,13 +52,51 @@ LLVMPY_CreateNewModulePassManager() {
     return llvm::wrap(new PassManager<Module>());
 }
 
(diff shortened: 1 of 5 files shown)


yashssh:
Ping

PR #1119, opened 2025-01-02
Update ffi.py

xschizo531:


gmarkall:
Why do you want to do this? Was there some issue you ran into?

gmarkall:
If this is aimed at solving https://github.com/StanislavPetrovV/Software_3D_engine/issues/10, then I think this is not the right path.

gmarkall:
Closing for now as I don't think this is the right fix. Please feel to re-open if you'd like to discuss the issue further.

PR #1139, opened 2025-01-30
Add win-64 wheel builder GHA workflow

swap357:
This PR adds a new GitHub Actions workflow for building llvmlite wheels on Windows. Key features:

- Builds wheels for Python 3.10-3.13 on Windows x64
- Includes DLL validation for Python 3.10-3.12:
- Basic validation for Python 3.13 (py-lief not yet supported)
- Artifacts retention: 7 days

PR #1141, opened 2025-02-01
refactor llvmdev win builder

swap357:


swap357:
tested run: https://github.com/swap357/llvmlite/actions/runs/13079158483

.github/workflows/llvmdev_win_builder.yml
```diff
@@ -8,35 +8,33 @@ jobs:
...
-        path: ${{ env.OUTPUT }}
+      - name: Clone repository
+        uses: actions/checkout@v4
+
+      - name: Setup Miniconda
+        uses: conda-incubator/setup-miniconda@v3
+        with:
+          auto-update-conda: true
+          auto-activate-base: true
+          conda-remove-defaults: true
```
esc, line 24:
```suggestion
          activate-environment: ""
```

This change will not activate the default `test` environment. And also, we need `defaults` to get the compilers and `conda-build` and so on.

.github/workflows/llvmdev_win_builder.yml
```diff
@@ -8,35 +8,33 @@ jobs:
...
+
+      - name: Install conda-build
+        run: |
+          conda install conda-build -y
```
esc, line 28:
```suggestion
          conda install conda-build
```

When using `setup-miniconda` action the `-y` and `--yes` are implicit and don't need to be specified.

.github/workflows/llvmdev_win_builder.yml
```diff
@@ -8,35 +8,33 @@ jobs:
...
+
+      - name: Build
```
esc, line 30: This whole section currently outputs the package file-name. For consistency with other workflow, it would be better to generate a directory that can act as a conda channel. You can take inspiration from: 

https://github.com/numba/llvmlite/blob/main/.github/workflows/llvmdev_win-64_conda_builder.yml#L30-L35

.github/workflows/llvmdev_win_builder.yml
```diff
@@ -8,35 +8,33 @@ jobs:
...
+        run: |
+          RECIPE_NAME=./conda-recipes/llvmdev_manylinux
+          conda build $RECIPE_NAME
+          echo "OUTPUT=$(conda build --output $RECIPE_NAME)" >> $GITHUB_ENV
+
+      - name: Upload artifact
```
esc, line 36: This whole section need to refactored to look more like:

https://github.com/numba/llvmlite/blob/main/.github/workflows/llvmdev_win-64_conda_builder.yml#L37-L44

Specifically this will need a better name for the final artifact, retention time etc..

esc (changes requested):
I have left a series of comments on the PR. In addition to those you will also need the following three changes:

1) Rename the file to match the pattern of the current workflows.

2) Rename the workflow itself to match the pattersn of the current workflows.

3)  For this to be reusable, you'll need to echo out the the workflow ID like:

```
      - name: Get Workflow Run ID
        run: |
          echo "Current workflow run ID: ${{ github.run_id }}"
          echo "Use this ID when triggering llvmlite workflow"
```

.github/workflows/llvmdev_win_builder.yml
swap357: updated

.github/workflows/llvmdev_win_builder.yml
swap357: updated

esc:
OK @swap357 this looks good now, I'll have to test this in conjunction with https://github.com/numba/llvmlite/pull/1139 to make sure that the artifact consumed by this "upstream" workflow will work for the "downstream" workflow.

esc (approved):
LGTM

esc:
I am merging this now, to get #1144 rolling 🛼

PR #1142, opened 2025-02-03
point conda build to correct package path

esc:
As title


esc:
Run is here: https://github.com/esc/llvmlite/actions/runs/13117282178

and this actually tests the code...

PR #1144, opened 2025-02-04
Refactor recipe build flows

dbast:
Refactoring the existing llvm build workflow into one:
* Merge llvmdev_win_builder.yml + llvmdev_win-64_conda_builder.yml into one to avoid code duplication
* Add input parameters to be able to select between recipes/platform to build
* Rename flow to llvmdev_build.yml as it contains nothing specific to windows and the code is just a generic conda package builder
* Add ${{ matrix.recipe }}-${{ matrix.platform }} to upload name to better differentiate artifacts
* Remove "x" from `bash -elx {0}` as that pollutes the Action log due to the implicit `source "$CONDA/etc/profile.d/conda.sh` calls in every step. Adding `set -x` explicitly into step code instead.
* Set `run-post: false` to avoid setup-miniconda cleaning up the Github Action VM as that is wiped after the run anyways
* Run the flow if it gets changed in PRs, to see a public verification that the changes are actually working
* Define `concurrency` so that repeated pushes to a PR will cancel all previous builds to save resources and have faster feedback.
* Add support for linux-64, linux-aarch64, osx-64, osx-arm64, win-64 builds via different native runner types
* With simple DAG: fast (5sec) **check** job produces a build matrix based on (event = {label, pr}, changed files, manual input). The **check** job logic is written in Python and can be further extended to influence the DAG.
* Flow can be triggered by
  * Changes to the workflow
  * Adding a label named "build" to random PRs (only maintainers can do that)
  * Manually via workflow dispatch and selecting the **conda-platform** + **recipe** via drop down menu (required merge of PR to be available)
* Aligned the up/download artifact steps

PR #1146, opened 2025-02-05
Extend pre-commit config

dbast:
While working on the Github Action workflow https://github.com/numba/llvmlite/pull/1144 I was relying on a local extended .pre-commit-config.yaml that I now share / propose for addition to the repo:
* Add actionlint hook which does lots of schema checking + shellcheck-ing (!) action step code
* Add check-jsonschema for pure schema checking of workflows and more
* Reducing the clang-format hook to `types_or: [c++, c, c#, cuda, metal]` as it would otherwise also format json files with lots of diff for the repo
* Adding flake8 hook: checks everything that is not excluded... so it includes new files to `./buildscripts/*` :)
* Fixed some .py file lints of files in the repo-root that are now not excluded anymore
* Adding yamlfmt to format the updated .pre-commit-config.yaml (applying to whole repo would be lots of noise)
* Adding the general pre-commit-hooks for json/yaml validation and more
* Adding the meta hooks for pre-commit hygiene
* Updated the renovate.json so that renovate also updates the pre-commit hooks

The config currently cleanly applies (via `pre-commit run -a -v`) to the whole repo (in under 5 seconds!) and was of great help while working on https://github.com/numba/llvmlite/pull/1144

I can also add a Github Workflow that runs pre-commit etc if there is greater interest from the team.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,40 @@
 repos:
-- repo: https://github.com/pre-commit/mirrors-clang-format
-  rev: v13.0.1
-  hooks:
-  - id: clang-format
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v5.0.0
+    hooks:
+      - id: check-added-large-files
+      - id: fix-byte-order-marker
+      - id: check-case-conflict
+      - id: check-json
+      - id: check-yaml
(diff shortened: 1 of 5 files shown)


esc (approved):
LGTM I tried it on current `main` -- works fine. It's certainly better than what was there before.

PR #1147, opened 2025-02-05
fixup path to download artifact from llvmdev_run

esc:
Adapt this workflow such that it can process artifacts from #1144


dbast:
this is also part of https://github.com/numba/llvmlite/pull/1144 :)

esc:
@dbast ah great, didn't see that!

PR #1153, opened 2025-02-06
Make pre-commit super fast via caching + avoiding updating miniconda

dbast:
pre-commit spends most of its time setting up the hooks (=cloning their repos and installing each hook into its own virtual env).

Caching helps to give faster feedback (~30 secs vs 73 secs before) on new PRs. And it protects against rate limit issues in situations of many fast PR updates as every hook is a repo that is cloned. The cache invalidates when the python version, pre-commit version or pre-commit config changes.

Also avoiding the setup-miniconda action as that anyways uses the miniconda already present on the runners to save time.

Also adding black hook to pre-commit and limit to buildscripts.

Note: This uses https://github.com/psf/black-pre-commit-mirror and not https://github.com/psf/black to avoid cloning the entire black repo (=faster).

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -10,19 +10,24 @@ on:
 
 jobs:
   pre-commit:
-    runs-on: ubuntu-latest
-    defaults:
-      run:
-        shell: bash -el {0}
+    runs-on: ubuntu-24.04
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Setup Miniconda
-        uses: conda-incubator/setup-miniconda@v3
-        with:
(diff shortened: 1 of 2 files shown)


esc:
I have looked at this and came to the conclusion that the speed gains here come at the expense of maintainability and readability. If this kind of "golfing" is needed in order to make an action go fast, this is a severe limitation of the GitHub action system in general and I would consider this a design flaw. Especially, since this is a compiler project, I take the stance that: the code should be readable first and foremost by humans and the "compiler" should optimize the code in such a way that it runs as fast as possible. If the human needs to write code in an "unnatural" way, the compiler "can be considered harmful".

PR #1154, opened 2025-02-06
execute dist tests for linux-64 conda pkgs via Azure

esc:
As title

### DIFF
--- a/llvmlite/tests/test_binding.py
+++ b/llvmlite/tests/test_binding.py
@@ -728,7 +728,7 @@ class TestDependencies(BaseTest):
             if not dep.startswith('ld-linux-') and dep not in allowed:
                 fails.append(dep)
         if len(fails) == 1:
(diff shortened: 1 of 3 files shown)


llvmlite/tests/test_binding.py
```diff
@@ -718,13 +718,21 @@ def test_linux(self):
             self.fail("failed parsing dependencies? got %r" % (deps,))
         # Ensure all dependencies are expected
         allowed = set(['librt', 'libdl', 'libpthread', 'libz', 'libm',
-                       'libgcc_s', 'libc', 'ld-linux', 'ld64'])
+                       'libgcc_s', 'libc', 'ld-linux', 'ld64', 'libzstd',
+                       'libstdc++'])
```
sklam, line 722: Why are these changed? Do the new libs require a change in the manylinux target platform for wheel?

esc: FWIW: this is not related to wheel builds, but conda packages.

PR #1159, opened 2025-02-19
Add IR layer typed pointers mode

rj-jesus:
This patch adds preliminary support for a "lazy opaque pointers" option whereby we emit IR with typed pointers if pointee types are known (even if opaque pointers are enabled). This makes numba-cuda work with opaque pointers enabled and should unblock #1082.

I've enabled opaque pointers unconditionally and made lazy opaque pointers on by default to facilitate testing. These changes could (should?) be moved to a separate PR if required.


llvmlite/ir/types.py
```diff
@@ -190,17 +186,6 @@ def gep(self, i):
     def intrinsic_name(self):
         return 'p%d%s' % (self.addrspace, self.pointee.intrinsic_name)
 
-    @classmethod
```
gmarkall, line 193: This was added by @sklam in #1051. I think there's no way this can survive the transition to opaque pointers anyway (and the fact it doesn't seem to have an effect on tests is also somewhat reassuring), but I want to check if @sklam has any thoughts here.

llvmlite/__init__.py
```diff
@@ -3,8 +3,12 @@
 del get_versions
 
 # FIXME: Remove me once typed pointers are no longer supported.
-def _opaque_pointers_enabled():
+# Let's enable opaque pointers unconditionally.
```
gmarkall, line 6: I think we're doing this because we need to use opaque pointers in the binding layer with later LLVMs:

```suggestion
# Opaque pointers unconditionally required for later LLVM versions
```

Is that right?

gmarkall (changes requested):
Thanks for this @rj-jesus !

I think the general idea is good and I'd be keen to press on with it. I'd hope our strategy would be something like:

- Merge this PR
- Get comfortable that nothing Numba-related is broken by it
- Bump the minimum LLVM version to 16 (because it already builds with it anyway, this should be a simple bump)
- Remove the references to typed pointers in the binding layer to clean things up a bit
- Move to later LLVM versions at our convenience

The only idea I'd like to float is WRT naming - we have "opaque pointers" and "lazy opaque pointers" which on the face of it don't have much specific meaning. I'd like to propose that:

- What we're adding here is called "IR layer typed pointers" (or "IR layer opaque pointers", with the required inversion of boolean conditions)
- What is enabled by the `_opaque_pointers_enabled()` function is called "Binding layer opaque pointers"

I think this naming makes it a bit clearer where the effects and differences lie in terms of the abstractions presented by llvmlite. I would be keen to hear the perspective of @sklam and @stuartarchibald here too though.

rj-jesus:
Hi @gmarkall, thanks very much for the feedback!

Your strategy sounds very good to me.

On the naming issue, I'm more than happy to rename "lazy opaque pointers" to "IR layer opaque pointers" or something else entirely, "lazy opaque pointers" was just a work-in-progress name in any case! As far as the normal "opaque pointer mode" goes, while I'm also very happy to rename it, I think maybe we should just remove it entirely. It won't be an _option_ per se going forward as there will be no alternative, and it was only introduced in the first place as a stop-gap solution to enable testing and porting. We've had it for a release cycle, which I believe was the initial plan. What do you think?

llvmlite/ir/types.py
rj-jesus: The opaque pointer version of this shouldn't be affected by this patch (which is why the tests aren't affected either).
I think the typed version can't survive, but my understanding from https://github.com/numba/llvmlite/pull/1064#issuecomment-2204515014 was that this was okay.

llvmlite/__init__.py
rj-jesus: Yeah, in later LLVM versions we don't have the option to disable opaque pointers at all.
As discussed in another reply, I think the whole "opaque pointer option" in llvmlite should be removed eventually (in a separate PR).

gmarkall:
> As far as the normal "opaque pointer mode" goes, while I'm also very happy to rename it, I think maybe we should just remove it entirely. It won't be an option per se going forward as there will be no alternative, and it was only introduced in the first place as a stop-gap solution to enable testing and porting. We've had it for a release cycle, which I believe was the initial plan. What do you think?

I think this is a good idea, and it would save an iteration compared to my original plan - I prefer what you suggest instead.

gmarkall:
> On the naming issue, I'm more than happy to rename "lazy opaque pointers" to "IR layer opaque pointers" or something else entirely, "lazy opaque pointers" was just a work-in-progress name in any case!

Thanks... Will wait and see what others think of the naming.

rj-jesus:
Hi @gmarkall, as discussed I've raised #1160 to drop the old "binding layer" opaque pointers mode (and bump the LLVM release supported to 16), and I've also just committed the rename "lazy opaque pointers" to "IR layer typed pointers" you had suggested. Can you think of anything else we should be doing?

gmarkall:
> Can you think of anything else we should be doing?

Not at this point - thanks for the updates!

gmarkall (dismissed):
The code changes look good to me, so I'm approving this.

However, there are updates needed for the docs, examples, and test scripts, which I'll push after approving for @rj-jesus to review (and CI to test :slightly_smiling_face:).

docs/source/user-guide/ir/types.rst
```diff
@@ -126,23 +127,25 @@ When Typed Pointers are enabled, the pointer type is represented using:
...
 
    import llvmlite
-   llvmlite.opaque_pointers_enabled = True
+   llvmlite.ir_layer_typed_pointers_enabled = False
 
    # ... continue using llvmlite ...
 
-When Opaque Pointers are enabled, the pointer type is represented using:
+When Opaque Pointers are enabled (by disabling Typed Pointers), the pointer
+type is represented using:
```
rj-jesus, line 148: Whilst this is _technically_ true, typed pointers won't be represented as "true" opaque pointers in the sense that they'll still be represented as `_TypedPointerType` regardless of `ir_layer_typed_pointers_enabled`.

This shouldn't matter for users of PointerType as everything should still work transparently (that's the whole point of having the shim). For all intended purposes, all pointer types are always `PointerType`.

docs/source/user-guide/ir/types.rst
```diff
@@ -100,9 +100,10 @@ Atomic types
 Pointer Types
 =============
 
-llvmlite presently supports both *Typed Pointers* and *Opaque Pointers*. Typed
-Pointers are currently the default; Opaque Pointers will become the default in
-future, and support for Typed Pointers will be eventually be removed.
+The IR layer presently supports both *Typed Pointers* and *Opaque Pointers*.
+Typed Pointers are currently the default; Opaque Pointers will become the
```
rj-jesus, line 104: Here, "the default" effectively depends on whether the pointee type is known. It's not so much of an option; it depends on how the user creates the pointer types.

rj-jesus (commented):
Thank you very much for these changes @gmarkall! Everything looks good, I just left a few comments about some of the phrasing, but nothing special.

I can't approve this explicitly, but this generally looks good to me.

docs/source/user-guide/deprecation.rst
```diff
@@ -30,20 +30,38 @@ switching to `Opaque Pointers <https://llvm.org/docs/OpaquePointers.html>`_:
...
+<https://docs.nvidia.com/cuda/nvvm-ir-spec/>`_.
 
 Examples(s) of the impact
 -------------------------
 
-Code that uses llvmlite to work with pointers or to parse assembly that uses
-pointers will break if not modified to use opaque pointers.
+In a future release of llvmlite, code that uses llvmlite to work with pointers
+or to parse assembly that uses pointers will break if not modified to use
+Opaque Pointers.
```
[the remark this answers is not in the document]
gmarkall: What I was aiming to imply here was that we might remove typed pointer support from the IR layer, but without being specific about exactly when. However, I think the "parse assembly that uses pointers" part of the sentence is making this overall statement wrong / misleading... Will see what I can do to improve it.

docs/source/user-guide/ir/types.rst
gmarkall: Actually reading your second comment below, I'm rethinking this. Will push an update that I hope is a bit more representative and helpful.

docs/source/user-guide/deprecation.rst
```diff
@@ -30,20 +30,38 @@ switching to `Opaque Pointers <https://llvm.org/docs/OpaquePointers.html>`_:
   therefore may have bugs that go unfixed).
 - In LLVM 17, support for Typed Pointers is removed.
 
-Although Opaque Pointers are already the default in LLVM 15, llvmlite still uses
-Typed Pointers by default with LLVM 15.
+Although Opaque Pointers are already the default in LLVM 15, llvmlite still used
+Typed Pointers by default with LLVM 15 in llvmlite 0.44.
+
+The binding layer will move to using Opaque Pointers. The IR layer will still
+support both Typed and Opaque Pointers, defaulting to Typed Pointers. This
```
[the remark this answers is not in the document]
gmarkall: Thanks, I've made this change.

docs/source/user-guide/deprecation.rst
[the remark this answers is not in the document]
gmarkall: Yes, that is clearer - I've used that wording.

docs/source/user-guide/deprecation.rst
```diff
@@ -30,20 +30,38 @@ switching to `Opaque Pointers <https://llvm.org/docs/OpaquePointers.html>`_:
...
+will be auto-upgraded to use Opaque Pointers transparently; no action is
+required by the user.
 
 Schedule
 --------
 
-- In llvmlite 0.44, Opaque Pointers will be the default pointer kind.
-- In llvmlite 0.45, support for Typed Pointers will be removed.
+- In llvmlite 0.45, support for Opaque Pointers in the binding layer will be
+  removed. The IR layer will still use Typed Pointers by default.
```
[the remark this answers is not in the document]
gmarkall: I did! Thanks for catching that!

docs/source/user-guide/ir/types.rst
gmarkall: I've tried to decouple the wording from whether the lazy mode is enabled or not, in the hope it simplifies things - let me know what you think.

gmarkall:
Thanks for the review @rj-jesus - I think your comments all make sense, so I've aimed to address / follow them - let me know what you think.

docs/source/user-guide/ir/types.rst
```diff
@@ -100,15 +100,14 @@ Atomic types
 Pointer Types
 =============
 
-llvmlite presently supports both *Typed Pointers* and *Opaque Pointers*. Typed
-Pointers are currently the default; Opaque Pointers will become the default in
-future, and support for Typed Pointers will be eventually be removed.
+The IR layer presently supports both *Typed Pointers* and *Opaque Pointers*.
+Support for Typed Pointers will be eventually be removed.
```
rj-jesus, line 104: I think there's a typo here (an extra "be" around "eventually").

rj-jesus (commented):
Thank you very much for the changes, @gmarkall! I think it's looking really good, I think the distinction between Typed and Opaque Pointers is clear now.

I can't formally approve the review, but please consider it approved.

docs/source/user-guide/ir/types.rst
rj-jesus: Thanks very much, I think the way you've broken it down is spot on.

docs/source/user-guide/ir/types.rst
gmarkall: Thanks, now fixed!

sklam (approved):

PR #1160, opened 2025-02-20
Drop binding layer typed pointer support.

rj-jesus:
At the LLVM level, going forward typed pointers are no longer supported.

This patch removes typed pointer support from the LLVM binding layer.

Initially, it also attempted to bump the LLVM version used to LLVM 16, but due to unrelated reasons with Anaconda, it seems it's easier to do that separately as a follow-up patch.

### DIFF
--- a/ffi/type.cpp
+++ b/ffi/type.cpp
@@ -144,22 +144,7 @@ API_EXPORT(uint64_t)
 LLVMPY_GetTypeBitWidth(LLVMTypeRef type) {
     llvm::Type *unwrapped = llvm::unwrap(type);
     auto size = unwrapped->getPrimitiveSizeInBits();
-    return size.getFixedSize();
-}
-
-// FIXME: Remove me once typed pointers support is removed.
-API_EXPORT(LLVMTypeRef)
-LLVMPY_GetElementType(LLVMTypeRef type) {
-    llvm::Type *unwrapped = llvm::unwrap(type);
-    llvm::PointerType *ty = llvm::dyn_cast<llvm::PointerType>(unwrapped);
-    if (ty != nullptr) {
-#if LLVM_VERSION_MAJOR < 14
-        return llvm::wrap(ty->getElementType());
-#else
-        return llvm::wrap(ty->getPointerElementType());
-#endif
-    }
-    return nullptr;
+    return size.getFixedValue();
 }
 
 API_EXPORT(LLVMTypeRef)
--- a/llvmlite/ir/types.py
+++ b/llvmlite/ir/types.py
@@ -193,7 +193,9 @@ class _TypedPointerType(PointerType):
 
     @property
     def intrinsic_name(self):
-        return 'p%d%s' % (self.addrspace, self.pointee.intrinsic_name)
+        if ir_layer_typed_pointers_enabled:
+            return 'p%d%s' % (self.addrspace, self.pointee.intrinsic_name)
+        return super(_TypedPointerType, self).intrinsic_name
 
 
 class VoidType(Type):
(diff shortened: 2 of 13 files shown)


gmarkall:
Note from triage meeting: items pinned for opaque pointer fixes may also be required here.

ffi/type.cpp
```diff
@@ -144,22 +144,7 @@ API_EXPORT(uint64_t)
...
-    llvm::PointerType *ty = llvm::dyn_cast<llvm::PointerType>(unwrapped);
-    if (ty != nullptr) {
-#if LLVM_VERSION_MAJOR < 14
-        return llvm::wrap(ty->getElementType());
-#else
-        return llvm::wrap(ty->getPointerElementType());
-#endif
-    }
-    return nullptr;
+    return size.getFixedValue();
```
gmarkall, line 147: Note on this change: `getFixedSize()` retured `getFixedValue()` anyway, and `getFixedSize()` is deprecated / removed in later LLVM versions.

llvmlite/ir/types.py
```diff
@@ -193,7 +193,9 @@ def gep(self, i):
 
     @property
     def intrinsic_name(self):
-        return 'p%d%s' % (self.addrspace, self.pointee.intrinsic_name)
+        if ir_layer_typed_pointers_enabled:
```
gmarkall, line 196: I think this is a fix for an error that was previously un-noticed / un-addressed - is that correct?

llvmlite/tests/test_ir.py
```diff
@@ -2451,9 +2451,12 @@ def assert_ne(ptr1, ptr2):
     def test_ptr_intrinsic_name(self):
         self.assertEqual(ir.PointerType().intrinsic_name, 'p0')
         self.assertEqual(ir.PointerType(addrspace=1).intrinsic_name, 'p1')
-        # Note: Should this be adjusted based on the pointer mode?
-        self.assertEqual(ir.PointerType(int1).intrinsic_name, 'p0i1')
-        self.assertEqual(ir.PointerType(int1, 1).intrinsic_name, 'p1i1')
+        if not ir_layer_typed_pointers_enabled:
```
gmarkall, line 2454: I think the changes here pertain to the fix I asked about above. Is that right?

ffi/type.cpp
rj-jesus: Yep, I believe these are just API renames.

gmarkall (commented):
I think this is good, but I have a couple of questions on the diff to complete my understanding of the changes.

llvmlite/ir/types.py
rj-jesus: Yep, it fixes an inconsistent behaviour in how pointer intrinsic names are printed.

Previously, even if we enabled IR layer typed pointers, the pointer's intrinsic name would still reflect its "true" underlying type. This is probably not the desired behaviour, so I've changed this to follow the same logic we have for the "_to_string" method.

I should have done this when we added IR layer typed pointers, but I think this got lost with the opaque pointer fixes (which in practice came _after_ IR layer typed pointers, but I had initially thought them to come before).

llvmlite/tests/test_ir.py
rj-jesus: Exactly, I wasn't sure if we should remove intrinsic names from typed pointers entirely (such that we would always print them in opaque pointer format), but then I thought we should probably keep consistency with `_to_string`, and so I've adjusted this accordingly.

gmarkall (approved):
Many thanks for the PR! I think this is good to approve, noting that:

- I've experimented with this locally and run into no issues, including running the Numba test suite.
- This isn't changing behaviour from the default, which has already had a full buildfarm run.
- It only removes the non-default, deprecated path, which needs to be removed to move forward with LLVM versions.

Issue #1164, opened 2025-03-03
Duplicate debug metadata for DIBasicType


jiel-nv:
The debug metadata cache is not working as expected, look at the following njit reproducer,

```
$ cat simple.py
from numba import njit

@njit("void(int64)", debug=True)
def f(x):
    y = x
    z = x
$ NUMBA_DUMP_LLVM=1 python simple.py | grep DIBasicType
!2 = !DIBasicType(encoding: DW_ATE_unsigned, name: "i8", size: 8)
!5 = !DIBasicType(encoding: DW_ATE_unsigned, name: "i8", size: 8)
!8 = !DIBasicType(encoding: DW_ATE_unsigned, name: "i32", size: 32)
!10 = !DIBasicType(encoding: DW_ATE_unsigned, name: "i8", size: 8)
!13 = !DIBasicType(encoding: DW_ATE_unsigned, name: "i8", size: 8)
!16 = !DIBasicType(encoding: DW_ATE_unsigned, name: "i32", size: 32)
!22 = !DIBasicType(encoding: DW_ATE_signed, name: "int64", size: 64)
!28 = !DIBasicType(encoding: DW_ATE_signed, name: "int64", size: 64)
!32 = !DIBasicType(encoding: DW_ATE_signed, name: "int64", size: 64)
!35 = !DIBasicType(encoding: DW_ATE_signed, name: "int64", size: 64)
```

For the same DWARF encoding and name,, e.g. {DW_ATE_signed, "int64"}, the corresponding DIBasicType are created multiple times.

PR #1165, opened 2025-03-03
Fix the metadata deduplication

jiel-nv:
The llvmlite debug metadata cache does not work as expected due to the encoding value using address of DIToken instance and serves as part of the key.

Solution is to extract the string value from the encoding object.

Since the DIToken class is designed to be a wrapper of enumeration value, e.g. the DW_* enumerations, that should appear bare in the emitted metadata, in this change, stringify DIToken is made to fix the problem.

This solves llvmlite issue https://github.com/numba/llvmlite/issues/1164

### DIFF
--- a/llvmlite/ir/module.py
+++ b/llvmlite/ir/module.py
@@ -42,6 +42,15 @@ class Module(object):
             fixed_ops.append((name, op))
         return fixed_ops
 
+    def str_ditok_operands(self, operands):
+        str_ops = []
+        for name, op in operands:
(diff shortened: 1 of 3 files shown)


jiel-nv:
> Could you add a test please?

Yes, done.

gmarkall (approved):
Thanks for the PR - this looks good to me.

I have tested it with the Numba and Numba-CUDA debuginfo tests locally as well, with no issues.

PR #1167, opened 2025-03-07
fix llvmdev osx64 conda builder

swap357:
This PR fixes osx64 build for llvmdev conda and wheel builder by setting up macOS sdk compatible to work with conda build on GHA runner.


conda-recipes/llvmdev/conda_build_config.yaml
```diff
@@ -16,5 +16,14 @@ c_compiler:              # [win]
 cxx_compiler:            # [win]
   - vs2019               # [win]
 
-MACOSX_SDK_VERSION:    # [osx and x86_64]
-  - 10.12              # [osx and x86_64]
+MACOSX_DEPLOYMENT_TARGET:
+  - 10.10                       # [osx and x86_64]
```
sklam, line 20: hm... how did it used to work? Does it not work without this change?

conda-recipes/llvmdev_for_wheel/conda_build_config.yaml
```diff
@@ -16,5 +16,14 @@ c_compiler:              # [win]
 cxx_compiler:            # [win]
   - vs2019               # [win]
 
-MACOSX_SDK_VERSION:    # [osx and x86_64]
-  - 10.12              # [osx and x86_64]
+MACOSX_DEPLOYMENT_TARGET:
+  - 10.10                       # [osx and x86_64]
```
sklam, line 20: same question here. Do we need these changes?

sklam (commented):
Can we get the osx wheel tester going with this? It's the best way to test whether the build config changes are ok.

conda-recipes/llvmdev/conda_build_config.yaml
swap357: I had that as was used several places on llvmlite build script. does not seem to be used on llvmdev though, I'll check once if that's implicitly by conda build

swap357: not required, removed redundant vars

conda-recipes/llvmdev_for_wheel/conda_build_config.yaml
swap357: removed

.github/workflows/llvmdev_build.yml
```diff
@@ -78,6 +78,11 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@v4
 
+      - name: Setup platform-specific requirements
+        run: |
+          chmod +x ./buildscripts/github/setup_platform.sh
```
esc, line 83: I don't think the file needs to have the execute bit set if you are going to use `source` below.

swap357: yes, removed

swap357:
> Can we get the osx wheel tester going with this? It's the best way to test whether the build config changes are ok.

the condo package as well as wheel are tested with GHA on this change -
https://github.com/numba/llvmlite/actions/runs/14068772517

esc:
@swap357 thank you for this. I think in order to test this well, we will need the osx builders for llvmlite and numba.

conda-recipes/llvmdev/conda_build_config.yaml
```diff
@@ -16,5 +16,5 @@ c_compiler:              # [win]
 cxx_compiler:            # [win]
   - vs2019               # [win]
 
-MACOSX_SDK_VERSION:    # [osx and x86_64]
-  - 10.12              # [osx and x86_64]
+CONDA_BUILD_SYSROOT:
```
esc, line 19: can you explain this change?

buildscripts/github/setup_platform.sh
```diff
@@ -0,0 +1,34 @@
...
+        sdk_dir="buildscripts/github"
+        mkdir -p "${sdk_dir}"
+
+        # Download SDK
+        echo "Downloading MacOSX10.10.sdk.tar.xz"
+        wget -q https://github.com/phracker/MacOSX-SDKs/releases/download/11.3/MacOSX10.10.sdk.tar.xz
+
+        # Verify checksum
+        echo "Verifying SDK checksum"
+        shasum -c "${sdk_dir}/MacOSX10.10.sdk.checksum"
```
esc, line 21:
```suggestion
        shasum -c "${sdk_dir}/MacOSX10.10.sdk.checksum" || exit 1
```

Besides all the OOB discussion regarding using `set -e` and inheriting the flags from the parent shell, let's use the above construct in order to make explicit (for the reader) that this command absolutely must terminate execution of the script

esc (changes requested):
One question and one requested change.

esc:
@swap357 thank you again, now waiting on the llvmlite builders to become available for review such that I can test the artifacts produced from this workflow in earnest.

conda-recipes/llvmdev/conda_build_config.yaml
swap357: MACOSX_SDK_VERSION var wasn't being used explicitly or implicitly, and conda-build uses the CONDA_BUILD_SYSROOT env variable to find downloaded SDK directory. This overrides using system SDK. referring conda build doc:
https://docs.conda.io/projects/conda-build/en/stable/resources/compiler-tools.html

buildscripts/github/setup_platform.sh
swap357: sure, will update

PR #1184, opened 2025-03-27
adding black to pre-commit

esc:
Isolating suggested change as per #1153


jakob1379:
consider using [Ruff](https://github.com/astral-sh/ruff) for faster lints, it adheres to black style as default.

esc:
maybe not then: https://github.com/numba/llvmlite/issues/1373

PR #1192, opened 2025-04-11
Add llvmlite usage example python notebook

yashssh:
Python notebook containing examples on llvmlite use cases

### DIFF
--- a/examples/notebooks/llvmlite.ipynb
+++ b/examples/notebooks/llvmlite.ipynb
@@ -1,0 +1,2272 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Check LLVM version"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 33,
+   "metadata": {
+    "colab": {
+     "base_uri": "https://localhost:8080/",
+     "height": 193
+    },
+    "id": "DFnLdhYtWKQA",
+    "outputId": "27f70239-8c1c-4e02-bead-3d2818d84dc9"
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "llvmlite version: 0.44.0dev0+146.g75f6e31\n",
+      "llvm version: 20.0.0\n"
+     ]
+    }
+   ],
+   "source": [
+    "import llvmlite\n",
+    "print(\"llvmlite version:\", llvmlite.__version__)\n",
+    "\n",
+    "from llvmlite.binding.initfini import llvm_version_info\n",
+    "print(\"llvm version:\", '.'.join(map(str, llvm_version_info)))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Example to show the perf benifits of Numba"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 34,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "60.3 μs ± 79.2 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)\n",
+      "5.13 ms ± 27.3 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)\n"
+     ]
+    }
+   ],
+   "source": [
+    "import numpy as np\n",
+    "from numba import jit\n",
+    "\n",
+    "arr = np.random.randn(100000)\n",
+    "\n",
+    "@jit\n",
+    "def get_sum_jit(arr):\n",
+    "    s = 0.0\n",
+    "    for x in arr:\n",
+    "        s += x\n",
+    "    return s\n",
+    "\n",
+    "def get_sum_no_jit(arr):\n",
+    "    s = 0.0\n",
+    "    for x in arr:\n",
+    "        s += x\n",
+    "    return s\n",
+    "\n",
+    "%timeit get_sum_jit(arr) \n",
+    "%timeit get_sum_no_jit(arr)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "QTJVm1JgeZfQ"
+   },
+   "source": [
+    "# Part 1: IR Builder\n",
+    "### How to use llvmlite's IRBuilder APIs to build your own llvm based compiler"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "hBP3iRdml4MY"
+   },
+   "source": [
+    "#### Example 1: Building a simple llvm function to add 2 32bit integers"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 35,
+   "metadata": {
+    "id": "nf2utaGgUQ3W"
+   },
+   "outputs": [],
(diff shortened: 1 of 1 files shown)


esc:
@yashssh can you rebase this? Seems like there has not been a review, so rebase is fine. Looks like the CI fail is stale in comp. to `main`.

yashssh:
> @yashssh can you rebase this? Seems like there has not been a review, so rebase is fine. Looks like the CI fail is stale in comp. to `main`.

Done

gmarkall:
@yashssh Is this still a draft, or ready for a review?

yashssh:
> @yashssh Is this still a draft, or ready for a review?

Some of the examples in the notebook need LLVM19(I created this mainly for my EuroLLVM 2025 presentation). If that's okay I can mark it ready for review.

gmarkall:
I think that's OK - I think we can review it independently of the LLVM 19 support, and merge it with / after the LLVM 19 support is merged so we're consistent.

esc:
I took a look at this and would like to propose the following patch:

```
commit ae825edc939934c1c8154f2c36eea9b6f180693b (HEAD -> yashwants/llvmlite-notebook)
Author: Emergency Self-Construct <esc@users.noreply.github.com>
Date:   Mon Mar 16 16:38:49 2026 +0100

    fix-up Numba example

    As title

diff --git a/examples/notebooks/llvmlite.ipynb b/examples/notebooks/llvmlite.ipynb
index e648fb552e..d6649386f8 100644
--- a/examples/notebooks/llvmlite.ipynb
+++ b/examples/notebooks/llvmlite.ipynb
@@ -45,39 +45,34 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 34,
+   "execution_count": 11,
    "metadata": {},
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "60.3 μs ± 79.2 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)\n",
-      "5.13 ms ± 27.3 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)\n"
+      "3.17 ms ± 39 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)\n",
+      "48.8 μs ± 408 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)\n"
      ]
     }
    ],
    "source": [
     "import numpy as np\n",
-    "from numba import jit\n",
+    "from numba import njit\n",
     "\n",
     "arr = np.random.randn(100000)\n",
     "\n",
-    "@jit\n",
-    "def get_sum_jit(arr):\n",
+    "@njit\n",
+    "def get_sum(arr):\n",
     "    s = 0.0\n",
     "    for x in arr:\n",
     "        s += x\n",
     "    return s\n",
     "\n",
-    "def get_sum_no_jit(arr):\n",
-    "    s = 0.0\n",
-    "    for x in arr:\n",
-    "        s += x\n",
-    "    return s\n",
     "\n",
-    "%timeit get_sum_jit(arr) \n",
-    "%timeit get_sum_no_jit(arr)"
+    "%timeit get_sum.py_func(arr) \n",
+    "%timeit get_sum(arr)"
    ]
   },
   {
@@ -2264,7 +2259,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.12"
+   "version": "3.14.3"
   }
  },
```

This makes two changes to the source code example:

a) use `@njit` over `@jit` -- this is the preferred spelling
b) instead of duplicating the function, use `.py_func` to execute the non-numba variant

And then I reversed the order of running to have the slow variant run first. I believe this is more natural ordering and makes it easier to see the acceleration.

I choose to post a patch here, wasn't quite sure how to use the github interface to propose a diff otherwise -- let me know if I should submit it via any other format.

esc (changes requested):
Added a proposal to modify the introduction.

yashssh:
Thanks for the review @esc! I have applied your suggestion and have also rebased the patch on top of main.

esc:
> Thanks for the review @esc! I have applied your suggestion and have also rebased the patch on top of main.

No worries, I will continue to review this. Unfortunately I may have suggested to use `@njit` over `@jit`. That suggestion was a bit outdated. `@jit` defaults to `npython=True` these days and so, please use `@jit` instead of `@njit`. I think there is only once instance of this in the patch I suggested. Sorry about that.

yashssh:
No worries, have fixed it

esc:
@yashssh thank you again for fixing this, my review continues. I think the Jupyter notebook should start with a short motivation and glossary to make sure all readers are on the same page. The following diff is a suggestion so feel free to edit it as you see fit. I also modified two of the headings to use the word `two` instead of the number `2`. It is commonplace in written text to use the word for number rather than the number itself for small integers.

```
diff --git c/examples/notebooks/llvmlite.ipynb i/examples/notebooks/llvmlite.ipynb
index d6649386f8..2da4afec2c 100644
--- c/examples/notebooks/llvmlite.ipynb
+++ i/examples/notebooks/llvmlite.ipynb
@@ -4,7 +4,32 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Check LLVM version"
+    "# llvmlite example\n",
+    "\n",
+    "In this Jupyter notebook, we explore llvmlite, the lightweight Python interface to LLVM."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Glossary\n",
+    "\n",
+    "* LLVM : A collection of modular, reusable compiler and tool-chain technologies. The acronym used to mean Low Level Virtual Machine -- but the project has grown well beyond that scope over time and is now just known as \"LLVM\".\n",
+    "* JIT: Just-in-Time, a type of compiler that runs as and when needed, commonly as the target program runs. Contrast this with Ahead-of-Time (AOT) where the program is compiled before execution.\n",
+    "* llvmlite : A lightweight Python interface to LLVM exposing only the LLVM functionality required by JIT compilers, originally designed to enable Numba.\n",
+    "* Numba: A Python JIT compiler that uses LLVM via llvmlite.\n",
+    "* IR : Intermediary Representation, an internal representation used by compilers.\n",
+    "* LLVM-IR: The IR used by LLVM."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Check LLVM version\n",
+    "\n",
+    "For reference, we introspect both the LLVM and the llvmlite versions."
    ]
   },
   {
@@ -82,7 +107,7 @@
    },
    "source": [
     "# Part 1: IR Builder\n",
-    "### How to use llvmlite's IRBuilder APIs to build your own llvm based compiler"
+    "### How to use llvmlite's IRBuilder APIs to build your own LLVM based compiler"
    ]
   },
   {
@@ -91,7 +116,7 @@
     "id": "hBP3iRdml4MY"
    },
    "source": [
-    "#### Example 1: Building a simple llvm function to add 2 32bit integers"
+    "#### Example 1: Building a simple LLVM function to add two 32 bit integers"
    ]
   },
   {
@@ -413,7 +438,7 @@
     "id": "_RTBxfhBh3H1"
    },
    "source": [
-    "Add another function called add3 to the moudle which take 3 integers as arguments and returns their sum"
+    "Add another function called `add3` to the module which take three integers as arguments and returns their sum"
    ]
   },
   {
```

Let me know what you think

yashssh:
Thanks for the suggestions @esc! I have updated the patch. 

Sorry for the delay, I'll be able to quickly address the review comments from now on.

PR #1203, opened 2025-04-22
gha/add llvmlite osx-arm64 wheel builder workflow

swap357:
this PR adds llvmlite wheel build and test workflow for osx-arm64


seibert (changes requested):
Only general comment: Did we decide to pin Github Actions to specific hashes instead of tags (like `v3`)?  If so, can you update this PR to use those hashes?

seibert:
@swap357 has informed me that the pinning of the action to a specific hash is generated by renovate, so we don't have to do it in this PR.

seibert (approved):

PR #1205, opened 2025-04-24
gha/add llvmlite_linux-64 wheel builder workflow

swap357:


seibert (changes requested):
Similar to the other PR: Can you pin the actions you use to hashes rather than tag names?

seibert:
@swap357 informs me that renovate generates the pinning of the action to a specific hash, so we don't have to do it in this PR.

seibert (approved):

PR #1216, opened 2025-05-09
gha/refactor llvmlite osx wheel workflows

swap357:
This PR is to refactor osx-64, and osx-arm64 wheel builder workflows. These are the changes:
- on osx-64 workflow, currently the llvmdev brings in libunwind and llvmlite links against it. to avoid, we move the libunwind to temp dir.
- update MACOSX_DEPLOYMENT_TARGET from "13.0" to "10.15", to use older target, matching buildfarm config. "13.0" was configured here initially to match oldest available sdk on the GHA runner.
- move wheel patching and validation logic onto scripts under - ./buildscripts/github/ to be consistent with practice on other workflows. This creates a clean workflow and version control scripts separately from workflow.


.github/workflows/llvmlite_osx-64_wheel_builder.yml
```diff
@@ -67,50 +67,24 @@ jobs:
           fi
           conda install -c "$CHAN" llvmdev=${{ env.FALLBACK_LLVMDEV_VERSION }} cmake libxml2 python-build
 
+          # Hide libunwind to prevent it from being linked against during build
```
esc, line 70: Can you add what happens in the case, when you do link against libuwind? For example:

```
          # Hide libunwind to prevent it from being linked against during build.
          # This prevents ... from becoming an issue.
          # See also: hyperlink
```

esc (changes requested):
Looks good. Only minor comment to update.. and I would suggest to use `osx-64` instead of `osx64` in the file names of the bash scripts, if possible, that will make it coherent with the workflow files.

esc (approved):

PR #1217, opened 2025-05-14
gha/refactor llvmlite win-64 workflows

swap357:
This PR is refactor on existing win-64 GHA workflows. 
- adds concurrency, package naming consistent with other workflows.
- adds upload to anaconda cloud step on wheel builder


.github/workflows/llvmlite_win-64_wheel_builder.yml
```diff
@@ -148,5 +151,54 @@ jobs:
...
+      - name: Setup Miniconda
+        uses: conda-incubator/setup-miniconda@v3
+        with:
+          python-version: ${{ env.VALIDATION_PYTHON_VERSION }}
+          conda-remove-defaults: true
+          auto-update-conda: true
+          auto-activate-base: true
+
+      - name: Install anaconda-client
+        run: conda install -c anaconda anaconda-client
```
esc, line 189: What is the `-c anaconda` for? Don't think I see that being used often.

.github/workflows/llvmlite_win-64_wheel_builder.yml
```diff
@@ -148,5 +151,54 @@ jobs:
...
+        with:
+          name: llvmlite-win-64-py${{ matrix.python-version }}
+          path: dist
+
+      - name: Upload wheel to Anaconda Cloud
+        env:
+          ANACONDA_API_TOKEN: ${{ secrets.NUMBA_CHANNEL_WHEEL_UPLOAD }}
+        run: |
+          cd dist
+          whl=$(stat -f "%m %N" ./*.whl | sort -n | tail -n 1 | cut -d' ' -f2-)
```
esc, line 202:
```suggestion
          whl=$(ls -1 ./*.whl)
```

As used above? Or does that not work in this context?

esc (changes requested):
Just one suggestion, then good for merge!

.github/workflows/llvmlite_win-64_wheel_builder.yml
swap357: I had looked it up over here - https://anaconda.org/search?q=anaconda-client. I think this can use either defaults, or anaconda channel

.github/workflows/llvmlite_win-64_wheel_builder.yml
swap357: yes, this expression was from other workflow. stat command not on win, and ls is simpler.

.github/workflows/llvmlite_win-64_wheel_builder.yml
swap357: keeping this as of now, if needed changing, will make a PR to change all wheel workflows

Issue #1194, opened 2025-04-14
Missing LLVM assertions in Linux build breaks debug output


Olobir:
## Reporting a bug

- [x ] I have tried using the latest released version of llvmlite (most recent is
 visible in the change log (https://github.com/numba/llvmlite/blob/main/CHANGE_LOG).
- [x ] I have included a self contained code sample to reproduce the problem.
  i.e. it's possible to run as 'python bug.py'.

According to the [Numba documentation](https://numba.readthedocs.io/en/stable/user/faq.html#does-numba-vectorize-array-computations-simd), observing LLVM diagnostics requires LLVM to be built with assertions enabled. The docs also mention that the build of llvmlite in the Numba channel should support this.
It seems that support for LLVM assertions was removed from the Linux build script:
In commit [98452e2](https://github.com/numba/llvmlite/commit/98452e2a63b286133b3d712c304e485ca289c666#diff-d9781738ee758e4f6d6c502016f43b7d9b7165c146b32080cf433ca80659dc4dL18-L27), the line `_cmake_config+=(-DLLVM_ENABLE_ASSERTIONS:BOOL=ON)` was removed from conda-recipes/llvmdev/build.sh
Meanwhile, `-DLLVM_ENABLE_ASSERTIONS=ON ^` remains in conda-recipes/llvmdev/bld.bat, meaning Windows builds still have assertions enabled.
This leads to a difference in behavior between Windows and Linux. 

Here is the environment yaml-file
```
name: numbadev

channels:
  - defaults
  - numba

dependencies:
  - python>=3.12
  - numba::numba=0.61.2  # => llvmlite built against LLVM 15 disables debug info
  # - numba::numba=0.60  # => llvmlite built against LLVM 14 enables debug info
  - numba::llvmlite
  - intel-cmplr-lib-rt
```

Here is a script that demonstrates the issue:
```
import llvmlite.binding as llvm
llvm.set_option('', '--debug-only=loop-vectorize')

import numpy as np
from numba import jit

@jit(nopython=True, fastmath=True)
def test_func(a):
    result = 0.0
    for i in range(a.shape[0]):
        result += a[i] * 2.0
    return result

# Trigger compilation.
a = np.arange(1000, dtype=np.float64)
test_func(a)
```
When this script is run on Windows with llvmlite built against LLVM 15, output like `LV:...` is printed as expected. On Linux no debug output is shown, probably because LLVM was not built with assertions.
[...]

PR #1225, opened 2025-05-29
[WIP] LLVM 19 llvmdev recipe

gmarkall:
Steps taken so far:

  - Update recipe versions and checksums
  - Build with no patches to see what happens - which are necessary now?
  - Build libunwind as a runtime - it seems it is no longer built as a project

I'd like to see if this can be run on GHA to see whether it works there (using this PR to try to experiment with that)

To do:

- Update bld.bat for Windows
- Fix the assertions issue in the build: #1194

### DIFF
--- a/conda-recipes/llvmdev/build.sh
+++ b/conda-recipes/llvmdev/build.sh
@@ -13,7 +13,7 @@ cd build
 
 export CPU_COUNT=4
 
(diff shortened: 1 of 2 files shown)


gmarkall:
The GHA run for this succeeded for linux-64, but doesn't show up in the status here (not sure exactly why): https://github.com/numba/llvmlite/actions/runs/15322773203

PR #1238, opened 2025-06-28
gha/restructure llvmdev ci

swap357:
This PR introduces several key fixes to the llvmdev GHA build process.
- builds llvmdev_for_wheel inside manylinux containers for `linux-64` and `linux-arm64` to ensure wheel compatibility.
- pins the `osx-64` compiler to version 14 to avoid conda selecting 17.x compilers.
- fixes `linux-arm64` builds by pinning `py-lief` to a version older than 0.16.
- standardizes the use of `linux-arm64` as the platform identifier.

### DIFF
--- a/.github/workflows/llvmdev_build.yml
+++ b/.github/workflows/llvmdev_build.yml
@@ -11,21 +11,23 @@ on:
     inputs:
       platform:
         description: Conda Platform
-        default: linux-64
+        default: all
         required: true
         type: choice
         options:
+          - all
           - linux-64
-          - linux-aarch64
+          - linux-arm64
           - osx-64
           - osx-arm64
           - win-64
       recipe:
         description: Recipe to build
-        default: llvmdev
+        default: all
         required: true
         type: choice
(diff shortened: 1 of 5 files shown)


esc:
I think the `osx-64` issue that needed a compiler pin has been resolved, can you remove those?

esc:
Can we use `linux-aarch64` instead of `linux-arm64` since the first is the name that conda uses?

swap357:
> I think the `osx-64` issue that needed a compiler pin has been resolved, can you remove those?

okay, will update.

> Can we use `linux-aarch64` instead of `linux-arm64` since the first is the name that conda uses?

used `-arm64` as its what gha uses but can change this. this will need changing downstream llvmlite workflow, will change there.

swap357:
keeping pin to use v14 compiler toolchain on both `osx-64` and `osx-arm64` for llvmdev and llvmdev-for-wheel recipes, since it otherwise gets 17.x that causes errors.

PR #1250, opened 2025-08-07
Fix pass timing logic to stop printing to stdout.

stuartarchibald:
This fixes the "pass timing" logic, that which reports on the time spent in LLVM internal passes, such that it doesn't print to STDOUT.

### DIFF
--- a/ffi/newpassmanagers.cpp
+++ b/ffi/newpassmanagers.cpp
@@ -293,6 +293,31 @@ LLVMPY_RunNewModulePassManager(LLVMModulePassManagerRef MPMRef,
 
     PrintPassOptions PrintPassOpts;
 
+    // NOTE: The following are used to provide an alternative outstream to
+    // STDOUT and need to be declared ahead of instantiating the
+    // StandardInstrumentations instance as they need to have a lifetime that
+    // is longer than the StandardInstrumentations. See following notes for
+    // details:
+    //
+    // The reason for this is that a StandardInstrumentations instance (SI)
+    // contains a TimePassesHandler instance (TP), when the SI goes out of scope
+    // it triggers the TP member destructor, and this is defined so as to call
+    // `TP->print()` to trigger the side effect of draining the pass timing
+    // buffer and resetting the timers. This `print()` call will by default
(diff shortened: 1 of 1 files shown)


stuartarchibald:
@yashssh I think this patch may fix the issue we've been seeing for quite some time whereby enabling the pass timing instrumentation has lead to the logged data appearing on `STDOUT`. Having debugged and attempted this fix, it also occurred to me that I am not sure of the purpose of using the `StandardInstrumentations` and the associated call back registration, given that the pass timing handler is seemingly registered independently in llvmlite's use case. In the source there's a hint that some options given to the `StandardInstrumentations` instance (`DebugLogging` and `VerifyEach`) may be exposed through the public API to the user in future though, so perhaps the reason for its presence? Your thoughts would be appreciated, thanks!

yashssh:
Thank you for the fix @stuartarchibald! I went over the codebase. I believe SI is essential for registering all the call back hooks including the timepasses hooks. However, as of now llvmlite is only making use of time-passes callback which we are anyway doing ourself independently in `EnableTimePasses`. 

I see below use cases of SI as of now:
- It's important for hooking up DebugLogging and VerifyEach if `llvmlite` needs them in future.
- It's responsible for calling `PreservedCFGCheckerInstrumentation::registerCallbacks` which verifies that the analysis passes are correctly invalidated after an IR change.
- All other callback registrations don't seem relevant for llvmlite but might be useful for any future features.

I think we can get rid of the `TimePasses` [callback registration](https://github.com/numba/llvmlite/blob/0cc7741314a4d556b00f79a7c13783f292d30be6/ffi/newpassmanagers.cpp#L462) in `EnableTimePasses`, that's just duplicating work.

yashssh:
As confirmed in downstream it's not safe to remove the Callback registration in EnableTimePasses so please ignore my suggestion. I'm also slightly in favour of leaving SI as is for now.

yashssh:
I have verified locally that the noise goes away with this patch. Thank you for submitting the change!! :) 

LGTM

yashssh (approved):

PR #1251, opened 2025-08-07
Fix various C++ issues and a linkage issue.

stuartarchibald:
This fixes:
* A few cases of invalid initialisation ordering, typically where the stack declaration of objects is in a different order to their use in constructors.
* A case of a function returning a C++ enum but being declared with C linkage.

### DIFF
--- a/ffi/custom_passes.cpp
+++ b/ffi/custom_passes.cpp
@@ -186,17 +186,15 @@ struct RefPrune {
     static const size_t FANOUT_RECURSE_DEPTH = 15;
     typedef SmallSet<BasicBlock *, FANOUT_RECURSE_DEPTH> SmallBBSet;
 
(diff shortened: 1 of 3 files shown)


gmarkall (approved):
The changes look good to me, though I'm struggling to imagine a concrete scenario where this caused an issue - out of interest was there a problem that this caused, or does it just fix items that are picked up as warnings?

stuartarchibald:
Thanks @gmarkall. There wasn't a specific issue, I was just cleaning up stuff as part of the migration to `CMake` builds in #1251 and was running with `-Wall -Werror` or similar set and it picked up these issues. I am wondering if we ought to just set these flags by default on the Azure CI builds. I also note that the MS compilers are complaining of other issues but haven't looked into them as yet.

PR #1269, opened 2025-08-26
add sdist build and upload steps on wheel workflow

swap357:
adds sdist creation and artifact export on osx-arm64 wheel builder workflow.


esc:
@swap357 thanks for the patch!

esc (approved):

Issue #780, opened 2021-10-18
"Segmentation fault (core dumped)" during call to numba and llvmlite


seafolk-cn:
Hi,

I installed a Qiskit 0.30 env on an IBM LInuxONE machine and want to utilize kaleidoscope to get a good presentation.
The installation of kaleidoscope is successful,  but when i attempting to run the sample test code, i immediately get the "Segmentation fault (core dumped)" error, see below.

The maintainer of kaleidoscope confirmed that the issue located somewhere in llvmlite.  
issue link:  https://github.com/QuSTaR/kaleidoscope/issues/63

Can someone help on this?

Thanks!
--------------------------------------------------------------------------------------------------------------------

(qiskit0.30) [root@rhel84 tutorials]# python
Python 3.8.11 (default, Aug 6 2021, 10:33:43)
[GCC 7.5.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.

            import numpy as np
            from qiskit import QuantumCircuit
            from qiskit.quantum_info import DensityMatrix
            import kaleidoscope.qiskit
            Segmentation fault (core dumped)
            (qiskit0.30) [root@rhel84 tutorials]#

##Try again with fault handler

(qiskit0.30) [root@rhel84 tutorials]# python -q -X faulthandler

            import numpy as np
            from qiskit import QuantumCircuit
            from qiskit.quantum_info import DensityMatrix
            import kaleidoscope.qiskit
            Fatal Python error: Segmentation fault

Current thread 0x000003ff9fbf8240 (most recent call first):
File "/opt/anaconda/envs/qiskit0.30/lib/python3.8/ctypes/init.py", line 373 in init
File "/opt/anaconda/envs/qiskit0.30/lib/python3.8/site-packages/llvmlite/binding/ffi.py", line 185 in
File "", line 219 in _call_with_frames_removed
File "", line 843 in exec_module
File "", line 671 in _load_unlocked
File "", line 975 in _find_and_load_unlocked
File "", line 991 in _find_and_load
File "", line 219 in _call_with_frames_removed
File "", line 1042 in _handle_fromlist
File "/opt/anaconda/envs/qiskit0.30/lib/python3.8/site-packages/llvmlite/binding/dylib.py", line 3 in
File "", line 219 in _call_with_frames_removed
File "", line 843 in exec_module
File "", line 671 in _load_unlocked
File "", line 975 in _find_and_load_unlocked
File "", line 991 in _find_and_load
[...]

PR #1284, opened 2025-09-15
Add s390x support to GHA pipeline

Darragh-Drohan:


esc:
Ref: https://github.com/numba/llvmlite/issues/780

PR #1286, opened 2025-09-16
GHA/refactor wheel builders to use system python for build

swap357:
Currently, the wheel builder workflows for `osx-arm64` and `win-64` have been using `python`  from miniconda env.

This PR updates the python used on build step to system python using -`actions/setup-python`. This would also allow getting pre-release versions of python for build and test using `allow-prereleases: true` param.

This PR also cleans up `Setup Miniconda` step from linux wheel builders which are redundant as build is performed under `manylinux` docker containers for linux platforms.


esc (approved):
LGTM

Issue #1294, opened 2025-09-18
llvmlite==0.45.0 and 0.45.1 do not build with Clang


tiran:
## Reporting a bug

llvmlite 0.45.0rc2 does not build with clang 20 from LLVM 20. The code for HAVE_FLTO detection is not compatible with clang. `clang -flto` generates LLVM IR bitcode, which cannot be linked with `clang++`. See #1293

```console
$ podman run -ti --rm fedora:42
# dnf -y install python3-devel llvm-devel clang cmake zlib-devel
# CC=clang CXX=clang++ pip wheel --no-deps --no-binary llvmlite llvmlite==0.45.0rc2
Collecting llvmlite==0.45.0rc2
  Using cached llvmlite-0.45.0rc2.tar.gz (185 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: llvmlite
  Building wheel for llvmlite (pyproject.toml) ... error
  error: subprocess-exited-with-error
   
  × Building wheel for llvmlite (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [51 lines of output]
      running bdist_wheel
      /usr/bin/python3 /tmp/pip-wheel-yqgq8_65/llvmlite_ba615e62d93a4c29baebbcb910f89d55/ffi/build.py
      -- The C compiler identification is Clang 20.1.8
      -- The CXX compiler identification is Clang 20.1.8
      -- Detecting C compiler ABI info
      -- Detecting C compiler ABI info - done
      -- Check for working C compiler: /usr/bin/clang - skipped
      -- Detecting C compile features
      -- Detecting C compile features - done
      -- Detecting CXX compiler ABI info
      -- Detecting CXX compiler ABI info - done
      -- Check for working CXX compiler: /usr/bin/clang++ - skipped
      -- Detecting CXX compile features
      -- Detecting CXX compile features - done
      -- Could NOT find FFI (missing: FFI_LIBRARIES HAVE_FFI_CALL)
      -- Looking for histedit.h
      -- Looking for histedit.h - found
      -- Found LibEdit: /usr/include (found version "2.11")
      -- Found ZLIB: /usr/lib64/libz.so (found version "1.3.1")
      -- Found zstd: /usr/lib64/libzstd.so
      -- Could NOT find LibXml2 (missing: LIBXML2_LIBRARY LIBXML2_INCLUDE_DIR)
[...]

PR #1295, opened 2025-09-18
fix: use CMake CheckIPOSupported

tiran:
Replace the custom `HAVE_FLTO` code with CMake's CheckIPOSupported module. The old code only works on GCC and fails on clang.

See: https://cmake.org/cmake/help/latest/module/CheckIPOSupported.html
Fixes: #1294

### DIFF
--- a/ffi/CMakeLists.txt
+++ b/ffi/CMakeLists.txt
@@ -91,21 +91,6 @@ else()
 '${LLVMLITE_PACKAGE_FORMAT}', expect one of 'conda' or 'wheel'.")
 endif()
 
-# Inherited from Makefile system, not tested on unsupported targets (BSD).
-if(UNIX)
-    set(LLVMLITE_FLTO_DEFAULT ON)
-    option(LLVMLITE_FLTO
-           "Enable LTO"
-           ${LLVMLITE_FLTO_DEFAULT})
-    if (LLVMLITE_FLTO)
-        check_cxx_compiler_flag(-flto HAVE_FLTO)
-        if(NOT HAVE_FLTO)
-            message(FATAL_ERROR "-flto flag is not supported by the compiler")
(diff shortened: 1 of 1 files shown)


tiran:
The `INTERPROCEDURAL_OPTIMIZATION` property adds `-flto=auto -fno-fat-lto-objects` for GCC and `-flto=thin` for Clang:

`/usr/bin/c++ -DLLVMLITE_LLVM_ASSERTIONS_STATE=0 -Dllvmlite_EXPORTS  -fPIC -std=gnu++17 -flto=auto -fno-fat-lto-objects -fPIC   -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -MD -MT CMakeFiles/llvmlite.dir/assembly.cpp.o -MF CMakeFiles/llvmlite.dir/assembly.cpp.o.d -o CMakeFiles/llvmlite.dir/assembly.cpp.o -c /work/ffi/assembly.cpp`

esc:
Fixes #1294

tiran:
Sure! Feel free to take my patch and do whatever you need to do. I filed the PR under the assumption that CMake would do the right thing on all platforms.

stuartarchibald:
> Sure! Feel free to take my patch and do whatever you need to do. I filed the PR under the assumption that CMake would do the right thing on all platforms.

That's great, thanks for confirming @tiran. I also think `CMake` is likely doing the right thing, it's just that it needs to be informed about the tool chains in use as non-standard behaviours are involved. When I've resolved these, I'll ping on the related PR to make sure that the original intent of fixing this for RHEL/Fedora hasn't been lost!

esc:
Now superseded by #1334.

PR #1298, opened 2025-09-19
remove osx-64 from buildscripts and CI

esc:
As title


esc:
Ref: https://github.com/numba/numba/issues/10187

buildscripts/incremental/build.sh
```diff
@@ -4,44 +4,6 @@ set -ex
 
 source activate $CONDA_ENV
```
swap357, line 6: For this change, we need to test it with changing `macos-13` image to `macos-14` on -
https://github.com/numba/llvmlite/blob/main/azure-pipelines.yml

swap357 (approved):
I've reverted changes to remove `osx-64` platform from azure pipelines, as there's no `osx-arm64` available on azure yet.

PR #1310, opened 2025-09-30
Change-log 0.45.1

esc:


esc:
Looks like this:

Screenshot 2025-10-01 at 10 07 04

the month names in the change-log are inconsistent, somtimes abbreviated, sometimes not. Guess we'll have to live with that?

Issue #1281, opened 2025-09-09
llvmlite CI support for Python 3.14


swap357:
This issue is to document and track the changes required to add initial support for Python 3.14 on CI. 
Python 3.14 is currently in pre-release (final release scheduled for October 7, 2025) per [PEP 745](https://peps.python.org/pep-0745/)

At present, we are using workarounds to build against pre-release versions. Once official release builds and packages are available, those will be reverted.

The goal is to:
- Enable CI jobs against Python 3.14 pre-releases.
- Identify and document any failures/required changes.
- Ensure CI is ready for Python 3.14 on the official release date.

swap357:
https://github.com/numba/llvmlite/pull/1279 adds CI support for llvmlite build and test GHA workflows.

Follow-up items - 
- [x]  Remove `ad-testing/label/py314 `dependency once upstream packages are published.
- [x]  Re-enable lief dependency in `3.14` test jobs once it has upstream support.

esc:
The initial part of this was solved with https://github.com/numba/llvmlite/pull/1279

swap357:
opened PR https://github.com/numba/llvmlite/pull/1311 to cleanup after official release and conda support for python 3.14.

nikimagic:
python 3.14t no GIL

(venv314t) PS C:\Users\User_Niki\venv314t> pip install llvmlite==0.46.0b1
Collecting llvmlite==0.46.0b1
Using cached llvmlite-0.46.0b1.tar.gz (185 kB)
Installing build dependencies ... done
Getting requirements to build wheel ... done
Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: llvmlite
Building wheel for llvmlite (pyproject.toml) ... error
error: subprocess-exited-with-error

× Building wheel for llvmlite (pyproject.toml) did not run successfully.
│ exit code: 1
╰─> [46 lines of output]
running bdist_wheel
C:\Users\User_Niki\venv314t\Scripts\python.exe C:\Users\User_Niki\AppData\Local\Temp\7\pip-install-hp1qa3xs\llvmlite_8c11503a66174cef9d9dcdd93c04fe8b\ffi\build.py
Trying generator ('Visual Studio 17 2022', 'x64', 'v143')
[...]

PR #1311, opened 2025-10-01
Python 3.14 cleanup on official support

swap357:
Cleanup after 3.14 GA:
- remove extra channel (`ad-testing/label/py314`) used on conda builder workflow.

https://github.com/numba/llvmlite/issues/1281


esc:
Also confirmed that these were to the only two references to `ad-testing`.

PR #1316, opened 2025-10-03
Prevent creating typed pointers of opaque pointers.

### LINKED ISSUE #1305: Opaque pointer on stack: ptr* is invalid
llvmlite verion: 0.45.0

Description: llvmlite generates type `ptr*`, which is invalid.

Happens when loading a scoring a pointer to a stack variable.

C equivalent code to make it simpler to understand:
```c
int main() {
  int n;
  int* ptr = *n;
  *ptr = 123;
  return n;
}
```

llvmlite code
```python
from llvmlite import ir, binding

i32_type = ir.IntType(32)
ptr_type = ir.PointerType()

module = ir.Module(name="pointer_manipulation")
fnty_main = ir.FunctionType(i32_type, [])
func_main = ir.Function(module, fnty_main, name="main")

block = func_main.append_basic_block(name="entry")
builder = ir.IRBuilder(block)
int_storage = builder.alloca(i32_type, name="int_storage")
ptr_storage = builder.alloca(ptr_type, name="ptr_storage")
# This fixes the issue:
# ptr_storage.type = ptr_type

builder.store(int_storage, ptr_storage)
loaded_int_addr = builder.load(ptr_storage, name="loaded_int_addr")
# This does not help:
# loaded_int_addr = builder.load(ptr_storage, typ=ptr_type, name="loaded_int_addr")
value_to_store = ir.Constant(i32_type, 123)
builder.store(value_to_store, loaded_int_addr)
return_value = builder.load(int_storage, typ=i32_type, name="final_value")
builder.ret(return_value)


mod = binding.parse_assembly(str(module))
mod.verify()
```

```llvm
define i32 @"main"()
{
entry:
  %"int_storage" = alloca i32
  %"ptr_storage" = alloca ptr
  store i32* %"int_storage", ptr* %"ptr_storage"
  %"loaded_int_addr" = load ptr, ptr* %"ptr_storage"
  store i32 123, ptr %"loaded_int_addr"
  %"final_value" = load i32, i32* %"int_storage"
  ret i32 %"final_value"
}
```

Full python error:
```
  File ".../llvm_exp.py", line 29, in <module>
    mod = binding.parse_assembly(str(module))
  File ".../.venv/lib/python3.13/site-packages/llvmlite/binding/module.py", line 25, in parse_assembly
<string>:10:33: error: ptr* is invalid - use ptr instead
  store i32* %"int_storage", ptr* %"ptr_storage"
                                ^
```

Is it intended that you need to do this in order to avoid the problem?
```python
ptr_storage = builder.alloca(ptr_type, name="ptr_storage")
ptr_storage.type = ptr_type
```

Am I doing something wrong?

rj-jesus:
Pointers of opaque pointers are always opaque themselves.

Fixes #1305.

### DIFF
--- a/llvmlite/ir/types.py
+++ b/llvmlite/ir/types.py
@@ -122,11 +122,13 @@ class PointerType(Type):
 
     # Factory to create typed or opaque pointers based on `pointee'.
     def __new__(cls, pointee=None, addrspace=0):
-        if cls is PointerType and pointee is not None:
+        if cls is PointerType and pointee is not None and \
+           type(pointee) is not PointerType:
             return super().__new__(_TypedPointerType)
         return super(PointerType, cls).__new__(cls)
 
-    def __init__(self, addrspace=0):
+    def __init__(self, __pointee=None, addrspace=0):
+        assert __pointee is None or type(__pointee) is PointerType
         self.addrspace = addrspace
 
     def _to_string(self):
@@ -160,8 +162,8 @@ class _TypedPointerType(PointerType):
     """
 
     def __init__(self, pointee, addrspace=0):
-        super(_TypedPointerType, self).__init__(addrspace)
-        assert pointee is not None
+        super(_TypedPointerType, self).__init__(None, addrspace)
+        assert pointee is not None and type(pointee) is not PointerType
         assert not isinstance(pointee, VoidType)
         self.pointee = pointee
         self.is_opaque = False
--- a/llvmlite/tests/test_ir.py
+++ b/llvmlite/tests/test_ir.py
@@ -2513,6 +2513,62 @@ class TestTypes(TestBase):
         assert_ne(ptrs['tp_i1_a1'], ptrs['tp_i8_a1'])
         assert_ne(ptrs['tp_i8_a0'], ptrs['tp_i8_a1'])
 
+    def test_pointers(self):
+        # Basic opaque pointers.
+        ptr = ir.PointerType()
+        ptr2 = ir.PointerType(addrspace=2)
+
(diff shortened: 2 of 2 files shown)


llvmlite/ir/types.py
```diff
@@ -122,11 +122,13 @@ class PointerType(Type):
...
     # Factory to create typed or opaque pointers based on `pointee'.
     def __new__(cls, pointee=None, addrspace=0):
-        if cls is PointerType and pointee is not None:
+        if cls is PointerType and pointee is not None and \
+           type(pointee) is not PointerType:
             return super().__new__(_TypedPointerType)
         return super(PointerType, cls).__new__(cls)
 
-    def __init__(self, addrspace=0):
+    def __init__(self, __pointee=None, addrspace=0):
```
sklam, line 130:
```suggestion
    def __init__(self, pointee=None, /, addrspace=0):
```

The use of `__` prefix in argument seems unusual. 
```python
>>> class A:
...     def __init__(self, __p=0):
...         self.p = __p
...
>>> A(__p=1)
  File "<stdin>", line 1, in <module>
TypeError: A.__init__() got an unexpected keyword argument '__p'

>>> import inspect
>>> inspect.signature(A.__init__)
<Signature (self, _A__p=0)>
```
If we want it to be positional-only, we should add `/` before `addrspace`.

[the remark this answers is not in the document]
rj-jesus: That was the intention, yes. 

Would you rather I drop the leading underscores, as @sklam also suggested?

rj-jesus: Thanks, it should be fixed now. :)

PR #1317, opened 2025-10-06
Correct debug message text

eric-wieser:
`LlvmliteMemoryManager::reserveAllocationSpace` was logging the unaligned size, but actually reserving the aligned size.


gmarkall (approved):
Whoops! Thanks for the the fix.

PR #1321, opened 2025-10-08
Fix OptimizationLevel mapping for Os

### LINKED ISSUE #1306: llvm.create_pipeline_tuning_options(size_level=…) not accessible
`llvm.create_pipeline_tuning_options()` encodes optimisation for size (`-Os`/`-Oz`) as `speed_level=2`:
https://github.com/numba/llvmlite/blob/4b401bfde9fbd706bf89fc49c9d4e9387b8a63ee/llvmlite/binding/newpassmanagers.py#L514-L521

However, on the C++ side, `newpassmangers.cpp`'s `mapLevel()` checks for `speed_level == size_level` for non-zero `size_level`:

https://github.com/numba/llvmlite/blob/4b401bfde9fbd706bf89fc49c9d4e9387b8a63ee/ffi/newpassmanagers.cpp#L264-L291

This is highly unintuitive (in fact, `Oz` is a lower speed optimisation level than `Os`), and in any case breaks the Python interface. To access `Os`, something like this is required:
```py
        pto = llvm.create_pipeline_tuning_options(size_level=1)
        # Set speed_level to result in -Os side-stepping the checks in create_pipeline_tuning_options()
        # which do not match the C++ side.
        pto.speed_level = 1
```

yashssh:
For Os LLVM defined speed level is 2 not 1.
(https://github.com/llvm/llvm-project/blob/main/llvm/lib/Passes/OptimizationLevel.cpp)

fixes #1306

### DIFF
--- a/ffi/newpassmanagers.cpp
+++ b/ffi/newpassmanagers.cpp
@@ -277,7 +277,7 @@ static OptimizationLevel mapLevel(int speed_level, int size_level) {
             llvm_unreachable("Invalid optimization level");
         }
     case 1:
-        if (speed_level == 1)
+        if (speed_level == 2)
             return OptimizationLevel::Os;
         llvm_unreachable("Invalid optimization level for size level 1");
     case 2:


yashssh:
Not sure what the macOS failure is about

stuartarchibald:
> Not sure what the macOS failure is about

I've restarted it, it may have been network related and a download was interrupted but a process returned `0` and so continued to the next step in the CI build.

dnadlinger:
As mentioned in #1306, I wonder about the backwards compatibility angle here, though – as an `llvmlite` user, there now isn't a piece of code I can write to access `Os` that works with both 0.44 as well as the next release.

yashssh:
While I agree there isn't a piece of code that will work with both `0.44` and the next release for `Os`. I think adding backward compatibility for this might not be ideal.

(Edit: Deleted a part of the comment, got confused in the release timeline)

PR #1334, opened 2025-10-23
CMake IPO support.

stuartarchibald:
This continues #1295, rebases the patch 4af342a from there on `main` and then adds additional code to support the `conda` toolchain.


stuartarchibald:
CC @tiran, as discussed in #1295 this PR continues the IPO support work extending it to work with the `conda` toolchain. Hopefully this will still work as intended with RHEL/Fedora.

stuartarchibald:
Note for reviewers, I've tested this on Windows (msvc), OSX (clang) and Linux (gcc), all are now reporting `Enabling interprocedural optimization (LTO)` and appropriate flags seem to be appearing in the commands to the compiler drivers.

PR #1335, opened 2025-10-23
Fix use of deprecated LLVM API.

stuartarchibald:
As title.


stuartarchibald:
NOTE: I am not immediately familiar with the API in use, I made a guess based on the notes surrounding the upstream LLVM patch that introduced the deprecation notice.

yashssh:
Thanks for the patch! This looks good.

Linking the upstream patch for reference if someone lands here:
https://github.com/llvm/llvm-project/commit/79499f010d2bfe809187a9a5f042d4e4ee1f1bcc
https://github.com/llvm/llvm-project/pull/124283/files

PR #1354, opened 2025-11-05
Improve error handling on FFI binding error

stuartarchibald:
This patch aims to improve the error handling when the FFI binding library cannot be loaded or is failing on symbol lookup. By the nature of the lazy loading in the FFI binding, if there is a problem with the library/symbol lookup it occurs on first use. This has manifested historically as a KeyError (because of the cache miss in the lazy load) with a large traceback, however the issue is not a KeyError, it's something more fundamental. This patch changes the exception handling such that the lazy loading part now raises exceptions with improved error messages raised from 'None' so as to give the user a hint as to what has gone wrong and to remove the noise from the KeyError.

### DIFF
--- a/llvmlite/binding/ffi.py
+++ b/llvmlite/binding/ffi.py
@@ -116,31 +116,32 @@ class _lib_wrapper(object):
 
     def _load_lib(self):
         test_sym = "LLVMPY_GetVersionInfo"
-        try:
-            mod_name = __name__.rpartition(".")[0]
-            lib_name = get_library_name()
-            with _suppress_cleanup_errors(_importlib_resources_path(
-                                          mod_name, lib_name)) as lib_path:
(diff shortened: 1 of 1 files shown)


stuartarchibald:
Note to reviewers. Suggested ways of testing this, try the following things and see what happens:
1. Remove the DSO
2. Rename the DSO
3. Make the DSO depend on a library that doesn't exist (see e.g. `patchelf --add-needed`)
4. Edit the name of the symbol `LLVMPY_GetVersionInfo`, either in the Python lookup or in the C++ code.

In all of the above, an exception should be raised with much shorter and clearer message about the problem.

sklam (approved):
Changes looks good. CI is failing is unrelated to changes in the PR.

sklam:
Ready to merge when CI is fixed

PR #1363, opened 2025-11-12
add publish to PyPI step on upload packages workflow

swap357:
as title


sklam (approved):
The `publish-pypi` mirrors the `publish-testpypi` with expected changes (name, repo-url, `needs`)

PR #1386, opened 2026-01-06
enable py-lief test dependency for py314 on llvmlite recipe

swap357:
closes https://github.com/numba/llvmlite/issues/1281


esc (approved):
Looks good. I did a `git grep` and could find no other instances of `py-lief` in a recipe or otherwise.

PR #1424, opened 2026-05-27
CI: remove osx-64 from azure pipelines testing

swap357:
support for `osx-64` was deprecated. This PR removes `osx-64` from azure CI testing.


stuartarchibald (approved):
Thanks for the patch. This makes the CI support consistent with Numba and the Tier 1 releases as defined by the Numba support policy.

PR #1425, opened 2026-05-29
CI: secure github actions workflows

swap357:
I had run [zizmor](https://github.com/zizmorcore/zizmor) - static analysis tool that helps find GitHub actions security issues.
This PR fixes all the flagged issues and makes CI more secure.

These are 3 categories of zizmor findings addressed on this PR:

- excessive-permissions
Added workflow-level `permissions:` with least-privilege defaults (contents: read; actions: read where jobs download artifacts from other workflow runs). Jobs that need more (e.g. PyPI OIDC) keep explicit job-level permissions.

- artipacked
Set `persist-credentials: false` on every `actions/checkout` step so the job token is not left in the git config for later steps or artifacts.

- template-injection
Removed `${{ }}` expansion inside `run:` shell scripts. Values from inputs, matrix, needs, and job env are passed via step/job env and referenced as shell variables


esc:
Any idea why the following were skipped?

esc:
@swap357 needs a rebase to resolve conflicts now.

swap357:
> Any idea why the following were skipped?
> 
> Uploading Screenshot 2026-06-03 at 13.36.51.png…

I can't access the image link, maybe it did not upload correct. 
On the latest commit [56875a9](https://github.com/numba/llvmlite/pull/1425/commits/56875a934d400bbd176b4bc38518a01e1c520826), the run shows - 
`No findings to report. Good job! (45 suppressed)`
The 45 suppressed are from strict checking that does not get flagged on default `zizmor` audit.
There's a stricter version using `zizmor --persona=auditor .github` which show those 45 suppressed findings and breakdown - 
`45 findings (6 unsafe fixes): 2 informational, 37 low, 6 medium, 0 high`
Most of them are around template usage. I could look into them at a later time and make follow-up patches, for now I think this PR would fix critical issues.

swap357:
The failing `win-64` workflow checks are related to vs2026 upgrade on GHA runner. needs https://github.com/numba/llvmlite/pull/1437 to be merged to resolve those.

PR #1427, opened 2026-06-02
Fix invalid fixup errors on Windows ARM64

MugundanMCW:
**PR Description:**

* This PR fixes the UnicodeSets test failures observed in the Numba test suite on Windows ARM64.

* Currently, llvmlite determines the code model for jitdefault based on `sizeof(void*)`. As a result, jitdefault resolves to the **Large** code model on Windows ARM64.

* According to [LLVM's implementation](https://github.com/llvm/llvm-project/blob/ea8f3dfe487e1c0df1fbd31dded17ec77567229e/llvm/lib/Target/AArch64/AArch64TargetMachine.cpp#L342), JIT compilation on Windows ARM64 should always use the **Small** code model. The Large code model generates long branches using four MOV instructions, and the Windows ARM64 relocation model does not support the relocations required for these sequences.

* The llvmlite overrides jitdefault to use the Large code model, JIT compilation fails with "invalid fixup" errors on Windows ARM64.

* To address this issue, the jitdefault code model selection logic has been updated to use the Small code model on Windows ARM64.

* After fixing the code model selection, the affected tests still failed silently at runtime without producing any output. Further investigation showed that LLVM's memory manager (`lmm`), which is intended for ARM64 targets, was not being enabled on Windows ARM64 platforms.

* This PR also updates the use_lmm detection logic to correctly recognize Windows ARM64 and enable the LLVM memory manager, allowing the affected tests to execute successfully.

Reference: https://github.com/numba/numba/issues/10619


gmarkall:
I think that we need to use the large code model - if we use the small code model I expect we're going to find relocations overflowing in real-world usage. To make that work, I think we need to investigate how to implement the fixups for relocations generated in the large code model.

MugundanMCW:
Hi @gmarkall 

 Thanks for the careful review. I dug into this further, and I'd like to make the case that Small is the correct and only viable approach here  - not a downgrade we're settling for.
 
 1. Large code model for JIT is unmaintained upstream, not just blocked on Windows:
    - Lang Hames (ORC/JIT lead) addressed exactly this in llvm/llvm-project#51371 (https://github.com/llvm/llvm-project/issues/51371). His guidance: "the large code model on MachO/arm64 has been unmaintained since we switched to JITLink as the default JIT linker for arm64... I would recommend... switching to the Small code model."
 
    - He closed the issue won't fix for large-code-model support. So upstream LLVM has actively moved JIT away from Large — the recommended design is Small, with the JIT linker keeping JIT'd code close and emitting the necessary indirection (stubs/GOT) for anything outside that region.
 
 2. On Windows ARM64 specifically, Large isn't even encodable:
    - The Large model's 4×MOV address sequence needs a MOVW-type relocation, and COFF/Windows ARM64 defines none (Microsoft never assigned one). The LLVM community already considered and rejected adding a non-standard one due to MSVC interop (discourse thread (https://discourse.llvm.org/t/fixup-aarch64-movw-support-for-coff-aarch64/53243)). This is also why upstream's getEffectiveAArch64CodeModel hard-codes
    - Small for this exact JIT triple. So "implement Large" would require inventing an object-format relocation that upstream has explicitly declined to add.
 
 3. The overflow concern is handled by the memory manager + linker indirection, which is the intended design
 Re: relocations overflowing in real usage - the ±128 MB branch range is handled by veneer/stub insertion in the runtime linker, and the ADRP ±4 GB data range is kept in bounds by the section memory manager allocating code + data contiguously (this is exactly what the use_lmm part of this PR enables — LlvmliteMemoryManager::reserveAllocationSpace packs Code/ROData/RWData into one contiguous reservation). This is the same Small + near-allocation + indirection model that .NET RyuJIT, V8, and LLVM's own ORC/JITLink use on Windows ARM64.
 
 So the realistic options are "Small + robust memory management" or "Windows ARM64 JIT doesn't work at all" - Large is not a third option on this platform, and is being retired for JIT even where it technically could run.
 
 Happy to add verification/a stress test demonstrating the veneer + near-allocation paths hold at scale if that would help build confidence. Would that address the concern?

gmarkall:
I'm still not at the point of getting a deep understanding here, but my initial observations about each of these points are:

1. The linked issue is about AArch64-Darwin, and it is the same issue that we worked around with the `LlvmliteSectionMemoryManager` in llvmlite. Following from that, I can't see how any facts from it can contribute to the reasoning for this issue.
2. The last comment on the thread suggests using ELF for JIT-only scenarios, which is what (ISTR) we had been doing for Windows on x86_64; I wasn't expecting that we would use COFF on WoA.
3. My understanding was that the `LlvmliteSectionMemoryManager` worked around Point 6 described in the ["Code Models" section of the System V ABI for AArch64](https://github.com/ARM-software/abi-aa/blob/main/sysvabi64/sysvabi64.rst#code-models):
   > The distance to the GOT from the text segment is limited to 2GiB by the R_AARCH64_GOTPCREL32 relocations.

My concern is that Numba does nothing to ensure that the combined span of text and data segments is less than 4GB required in the small code model. Concretely this can occur through the following sequence:

1. Numba JITs some code and data.
2. Other code does something using a large amount of memory (e.g. allocates a large array)
3. Numba JITs more code and data, with references to the code and/or data jitted in step 1.

I think we could explore the argument that Numba should have been designed not to do that, and that the correct fix is to change Numba so that it doesn't try to do that and therefore is compatible with the small code model. But I'm not optimistic about that being a straightforward fix either.

> Happy to add verification/a stress test demonstrating the veneer + near-allocation paths hold at scale if that would help build confidence. Would that address the concern?

I think I don't understand how the small code model is the solution, so I expect it would either help me build confidence, or it would help to highlight an issue with the reasoning behind why it would work. If you have time to add that please do it, but I understand if you have a lot of other things that are higher priority for you.

MugundanMCW:
Thanks @gmarkall for your comments -
 

- On ELF vs COFF  - you're correct. The JIT path appends -elf to the triple in the jitdefault branch of targets.py, so on Windows it goes through `RuntimeDyldELFAArch64` rather than the COFF writer. 
- That means the "COFF has no MOVW relocation" argument doesn't apply to this issue at all. I was able to trace the actual error (invalid fixup for add (uimm12) instruction) back to `AArch64ELFObjectWriter.cpp`, where it's hitting a `fixup_aarch64_add_imm12` that the large code model emits in a form the writer has no relocation so for. 
- ELF supports MOVW, and large-code-model codegen for AArch64 is unmaintained. So the COFF reasoning was the wrong explanation on my part.
- The 4GB span question is the one that matters, and I agree the passing tests don't address it, since none of those modules come close to the boundary. 
- I spent some time in `RuntimeDyldELF.cpp` working out what the loader does for cross-module references. Calls look safe - `resolveAArch64ShortBranch` only takes the direct path when the target is in the same section and within ±128MB, and otherwise falls back to a MOVZ/MOVK stub, so far calls always get a veneer. 
- GOT-based references also look safe, since the loader allocates a GOT entry holding the full 64-bit address and only the nearby GOT has to be in range. The case I can't account for is a direct ADR_PREL_PG_HI21 to a symbol more than 4GB away: there's no stub equivalent, and the range check is only an assert, so a release build would silently truncate. That's precisely the scenario you laid out.
- So it seems to come down to a single question: for a cross-module reference, does codegen produce a GOT access or a direct ADRP? That's governed by the relocation model. If external references go through the GOT, your three-step scenario can't overflow. If they don't, it's a real problem, and the fix is likely to switch the JIT to PIC on this target.
- I can try to verify this - I am thinking to first check which relocations are actually emitted for an external reference under the current settings, then write the stress test you suggested - JIT a module, force a large allocation to push the next one well away, then JIT a second module that references the first and call across it. If that reproduces the overflow, I'll try moving external references to the GOT via PIC and confirm whether it resolves it.

This can take some time to do it properly. I'll follow up with the relocation findings and the stress test results.

gmarkall:
Thanks for the follow-up. A couple of inline responses:

> That means the "COFF has no MOVW relocation" argument doesn't apply to this issue at all. I was able to trace the actual error (invalid fixup for add (uimm12) instruction) back to `AArch64ELFObjectWriter.cpp`, where it's hitting a `fixup_aarch64_add_imm12` that the large code model emits in a form the writer has no relocation so for.

I was planning to dig through this to see why that happens - whether there's an appropriate relocation that's not implemented, or if some earlier bug / issue has led to this path.

> GOT-based references also look safe, since the loader allocates a GOT entry holding the full 64-bit address and only the nearby GOT has to be in range. The case I can't account for is a direct ADR_PREL_PG_HI21 to a symbol more than 4GB away: there's no stub equivalent, and the range check is only an assert, so a release build would silently truncate. That's precisely the scenario you laid out.

Thanks for looking at this and sharing the details.

> So it seems to come down to a single question: for a cross-module reference, does codegen produce a GOT access or a direct ADRP? That's governed by the relocation model. If external references go through the GOT, your three-step scenario can't overflow. If they don't, it's a real problem, and the fix is likely to switch the JIT to PIC on this target.

I am now hoping it is the case that the external references go through the GOT. Intuitively I'd hope that they do, but I think we need to try it to find out.

> I can try to verify this - I am thinking to first check which relocations are actually emitted for an external reference under the current settings, then write the stress test you suggested - JIT a module, force a large allocation to push the next one well away, then JIT a second module that references the first and call across it. If that reproduces the overflow, I'll try moving external references to the GOT via PIC and confirm whether it resolves it.
> 
> This can take some time to do it properly. I'll follow up with the relocation findings and the stress test results.

Thank you - that does sound like a great test.


For a bit of background, I switched to focusing on this issue at the beginning of the week, but it's taken me a bit of time to get a working WoA setup so I'm only just at the point where I've nearly completed an LLVM build I can use for stepping through / debugging / trying things out. I should be able to experiment more actively soon.

ffi/targets.cpp
```diff
@@ -180,10 +180,18 @@ LLVMPY_CreateTargetMachine(LLVMTargetRef T, const char *TripleStr,
     else { // catches "jitdefault" and not set, as per LLVM 5, needed for MCJIT
         // fall through, use model based on bitness
         int bits = sizeof(void *);
-        if (bits == 4)
+        if (bits == 4) {
             cm = CodeModel::Small;
-        else
-            cm = CodeModel::Large;
+        } else {
+            llvm::Triple TT(Triple);
```
gmarkall: With the recent LLVM 22 changes, this needs to be:

```suggestion
            llvm::Triple TT(TripleStr);
```

gmarkall:
With this PR (with the frame lowering patch built into LLVM), the test suite is running correctly for me on a Windows on ARM system - hopefully with the suggested patch, CI should pass.

ffi/targets.cpp
MugundanMCW: Thanks for the suggestion @gmarkall, I have applied the fix

gmarkall (dismissed):

gmarkall:
@MugundanMCW I am happy with this change - the only thing preventing merging is that the formatting check fails. Please fix it up if you would like, or I can fix the formatting issue if you prefer - let me know!

MugundanMCW:
Hi @gmarkall,

Thanks for letting me know. I don't have `clang-format-20` in my local setup. If you don't mind, could you please make the relevant formatting changes?

gmarkall (approved):

swap357:
I tested Numba using llvmlite artifacts from this PR build, including the previously skipped UnicodeSet tests, and can confirm that this patch resolves the errors previosuly seen -

https://github.com/numba/numba/actions/runs/28604381479/job/84821069197?pr=10678
```
test_isdisjoint (numba.tests.test_sets.TestUnicodeSets.test_isdisjoint) ... ok
test_issubset (numba.tests.test_sets.TestUnicodeSets.test_issubset) ... ok
test_issuperset (numba.tests.test_sets.TestUnicodeSets.test_issuperset) ... ok

Ran 3 tests in 32.031s

OK
```

swap357:
All looks good. Thank you @MugundanMCW and @gmarkall for the efforts!

PR #1430, opened 2026-06-03
update README to document support for LLVM 22 on v0.48.0

swap357:
as title

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -54,7 +54,7 @@ Key Benefits
 Compatibility
 =============
 
-llvmlite has been tested with Python 3.10 -- 3.13 and is likely to work with
+llvmlite has been tested with Python 3.10 -- 3.14 and is likely to work with
 greater versions.
 
 As of version 0.48.0, llvmlite requires LLVM 22.x.x on all architectures
--- a/README.rst
+++ b/README.rst
@@ -57,14 +57,15 @@ Compatibility
 llvmlite has been tested with Python 3.10 -- 3.13 and is likely to work with
 greater versions.
 
-As of version 0.45.0, llvmlite requires LLVM 20.x.x on all architectures
+As of version 0.48.0, llvmlite requires LLVM 22.x.x on all architectures
 
 Historical compatibility table:
 
 =================  ========================
 llvmlite versions  compatible LLVM versions
 =================  ========================
-0.45.0 - ......    20.x.x
+0.48.0 - ......    22.x.x
+0.45.0 - 0.47.x    20.x.x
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -57,14 +57,15 @@ Compatibility
 llvmlite has been tested with Python 3.10 -- 3.13 and is likely to work with
```
esc:
```suggestion
llvmlite has been tested with Python 3.10 -- 3.14 and is likely to work with
```

esc (changes requested):
Might as well update the Python version while editing this file.

In the long run, it would be good to link this file into the sphinx based docs, so that less information needs to be duplicated.

esc (dismissed):

swap357:
> Might as well update the Python version while editing this file.
> 
> In the long run, it would be good to link this file into the sphinx based docs, so that less information needs to be duplicated.



I've added llvm compatibility table reference on index page with [e631d6a](https://github.com/numba/llvmlite/pull/1430/commits/e631d6ae1da32111811fff725c4f8705facdbad5). is that okay ?
renders and looks like this-

image

esc:
> > Might as well update the Python version while editing this file.
> > In the long run, it would be good to link this file into the sphinx based docs, so that less information needs to be duplicated.
> 
> I've added llvm compatibility table reference on index page with [e631d6a](https://github.com/numba/llvmlite/pull/1430/commits/e631d6ae1da32111811fff725c4f8705facdbad5). is that okay ?

For now, yes, but using line-numbers to fetch the table will break as soon as we update it, so that is a bit fragile. OK for now in the interest of a release but should be refactored soon. I was thinking to merge the `index.rst` and `README.rst` into a single document: `README.rst`. The Sphinx docs can just include the whole thing and the `setup.py` can pull it to render the overview page on PyPi. I.e. one consistent document, but suitable for different use-cases.

For example the `index.rst` currently has the following section about LLVM compatibility:

```
LLVM compatibility
==================

Despite minimizing the API surface with LLVM, llvmlite is
impacted by changes to LLVM's C++ API, which can occur at every
feature release. Therefore, each llvmlite version is targeted to
a specific LLVM feature version and works across all given bugfix
releases of that version.

EXAMPLE: Llvmlite 0.12.0 works with LLVM 3.8.0 and 3.8.1, but
it does not work with LLVM 3.7.0 or 3.9.0.

Numba's requirements determine which LLVM version is supported.
```

Which is:

a) inferior to the overview table and
b) doesn't event reference it!

However, this is not release critical, so please go ahead with the current variant.

esc (approved):

PR #1439, opened 2026-06-17
Add LLVM FrameLowering patch to llvmdev recipes

swap357:
LLVM PR https://github.com/llvm/llvm-project/pull/204347 addressed cause of errors on https://github.com/numba/numba/issues/10652. 

This PR adds adapted patch from LLVM PR on llvmdev recipes. Adapted because current llvmdev recipe uses LLVM 22.1.0, where TargetMachine::getMCAsmInfo() still returns a pointer (the upstream PR targets main, where it returns a reference).
The functional change and the upstream comment are kept verbatim; only the removed line uses `->`.

### DIFF
--- a/conda-recipes/llvm22-aarch64-windows-target-detection.patch
+++ b/conda-recipes/llvm22-aarch64-windows-target-detection.patch
@@ -1,9 +1,12 @@
 [AArch64] Fix Windows target detection in FrameLowering
 
-Adapted from llvm/llvm-project PR #204347 to apply on the LLVM 22.1.0 release
-tree, where TargetMachine::getMCAsmInfo() still returns a pointer (the upstream
-PR targets main, where it returns a reference). The functional change and the
-upstream comment are kept verbatim; only the removed line uses ->.
+Adapted from LLVM upstream commit c327ab359a959de2e4241a5fcda409958f2c0d11 (PR #204347):
+https://github.com/llvm/llvm-project/commit/c327ab359a959de2e4241a5fcda409958f2c0d11
+
+Applied on the LLVM 22.1.0 release tree, where TargetMachine::getMCAsmInfo() still
+returns a pointer (the upstream commit targets main, where it returns a reference).
+The functional change and the upstream comment are kept verbatim; only the removed
+line uses ->.
 
 For aarch64-pc-windows-msvc-elf we would use the Windows callee-save list in
 AArch64RegisterInfo::getCalleeSavedRegs(), but FrameLowering would handle this
--- a/conda-recipes/llvm22-aarch64-windows-target-detection.patch
+++ b/conda-recipes/llvm22-aarch64-windows-target-detection.patch
@@ -1,0 +1,28 @@
+[AArch64] Fix Windows target detection in FrameLowering
+
+Adapted from llvm/llvm-project PR #204347 to apply on the LLVM 22.1.0 release
+tree, where TargetMachine::getMCAsmInfo() still returns a pointer (the upstream
+PR targets main, where it returns a reference). The functional change and the
+upstream comment are kept verbatim; only the removed line uses ->.
+
+For aarch64-pc-windows-msvc-elf we would use the Windows callee-save list in
+AArch64RegisterInfo::getCalleeSavedRegs(), but FrameLowering would handle this
+like Linux, and fail to invalidate the (x29, x28) pairing.
+
+This patch switches back to using AArch64Subtarget::isTargetWindows(), which
+aligns with getCalleeSavedRegs().
+
+diff --git a/llvm/lib/Target/AArch64/AArch64FrameLowering.cpp b/llvm/lib/Target/AArch64/AArch64FrameLowering.cpp
+--- a/llvm/lib/Target/AArch64/AArch64FrameLowering.cpp
++++ b/llvm/lib/Target/AArch64/AArch64FrameLowering.cpp
+@@ -384,5 +384,9 @@
+ }
+
+ static bool isTargetWindows(const MachineFunction &MF) {
+-  return MF.getTarget().getMCAsmInfo()->usesWindowsCFI();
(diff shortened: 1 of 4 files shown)


conda-recipes/llvm22-aarch64-windows-target-detection.patch
```diff
@@ -0,0 +1,28 @@
+[AArch64] Fix Windows target detection in FrameLowering
+
+Adapted from llvm/llvm-project PR #204347 to apply on the LLVM 22.1.0 release
```
stuartarchibald: Can you please add the URL of the patch with a full commit SHA for reference?

conda-recipes/test-frame-record-win-arm64.ll
```diff
@@ -0,0 +1,23 @@
+; From llvm/llvm-project PR #204347. llvmdev is built with assertions,
```
stuartarchibald: As above, please add the URL of the patch with a full commit SHA for reference.

conda-recipes/llvmdev_for_wheel/meta.yaml
```diff
@@ -64,6 +67,8 @@ test:
     - if not exist %LIBRARY_INC%\\llvm\\Pass.h exit 1        # [win]
     - if not exist %LIBRARY_LIB%\\LLVMSupport.lib exit 1     # [win]
 
+    - llc ../test-frame-record-win-arm64.ll                  # [win and arm64]
```
stuartarchibald, line 75: This may not work as a test for those copying our recipe as it requires an assertion enabled build of LLVM (most places don't build with assertions). I'd typically suggest adding a note to the recipe to explain what is expected, but in this case, I'm wondering what happens on a non-assertion build of LLVM where the "fix" isn't present, does the compiler break somewhere after running the frame lowering?

stuartarchibald (commented):
Thanks for the patch. There's a couple of comments to resolve else looks good.

conda-recipes/llvmdev_for_wheel/meta.yaml
swap357: Away from desk at the moment, will check how this test does with non-assertion build. I'd think it would miscompile and this test would be a no-op. I'll put a note once I confirm though.

conda-recipes/llvm22-aarch64-windows-target-detection.patch
swap357: corrected.

conda-recipes/test-frame-record-win-arm64.ll
swap357: corrected.

conda-recipes/llvmdev_for_wheel/meta.yaml
swap357: I built llc with assertions off changing the CMAKE flag- 
```
cmake -DCMAKE_INSTALL_PREFIX="${PREFIX}"
...
-DLLVM_ENABLE_ASSERTIONS=OFF \
${CMAKE_ARGS} \
      -GNinja \
      ../llvm
 
 ninja -j${CPU_COUNT} llc
```
and built just `llc`

on the non-assertion build the compiler doesn't break after frame lowering, it silently miscompiles. (`x29` (FP) and `x30` (LR) are not paired together)
```
(base) llvm-recipe-repro/llvm-project-22.1.0.src/build ❯ ./bin/llc /Users/mac357/dev/swap357/llvmlite/conda-recipes/test-frame-record-win-arm64.ll -o -
        .file   "test-frame-record-win-arm64.ll"
        .text
        .globl  large_stack_requires_frame_record // -- Begin function large_stack_requires_frame_record
        .p2align        2
        .type   large_stack_requires_frame_record,@function
large_stack_requires_frame_record:      // @large_stack_requires_frame_record
// %bb.0:
        str     x30, [sp, #-32]!                // 8-byte Folded Spill
        stp     x29, x28, [sp, #16]             // 16-byte Folded Spill
        mov     x29, sp
        sub     sp, sp, #512
        adrp    x8, :got:baz
        mov     x0, sp
        ldr     x8, [x8, :got_lo12:baz]
        blr     x8
        mov     w0, wzr
        add     sp, sp, #512
        ldp     x29, x28, [sp, #16]             // 16-byte Folded Reload
        ldr     x30, [sp], #32                  // 8-byte Folded Reload
        ret
.Lfunc_end0:
        .size   large_stack_requires_frame_record, .Lfunc_end0-large_stack_requires_frame_record
                                        // -- End function
        .section        ".note.GNU-stack","",@progbits
(base) llvm-recipe-repro/llvm-project-22.1.0.src/build ❯ echo $?
0
```
I've added a note above the test clarifying that the test only works on build with assertions on.

conda-recipes/llvm22-aarch64-windows-target-detection.patch
```diff
@@ -0,0 +1,31 @@
+[AArch64] Fix Windows target detection in FrameLowering
+
+Adapted from LLVM upstream commit c327ab359a959de2e4241a5fcda409958f2c0d11 (PR #204347):
+https://github.com/llvm/llvm-project/commit/c327ab359a959de2e4241a5fcda409958f2c0d11
+
+Applied on the LLVM 22.1.0 release tree, where TargetMachine::getMCAsmInfo() still
+returns a pointer (the upstream commit targets main, where it returns a reference).
+The functional change and the upstream comment are kept verbatim; only the removed
+line uses ->.
```
stuartarchibald, line 9: Thanks for noting this, I checked upstream, change makes sense.

conda-recipes/llvmdev_for_wheel/meta.yaml
stuartarchibald: Thanks for checking and adding the note.

stuartarchibald (approved):
Thanks for the updates, patch looks good.

PR #1442, opened 2026-06-23
FIX: llvmlite win-arm64 wheel builder llvmdev channel

swap357:
as title


stuartarchibald (approved):
Thanks for the patch.

PR #1460, opened 2026-07-25
Fix FloatType/HalfType formatting of special values passed as strings

### LINKED ISSUE #833: string representations of special FP values work with DoubleType, but not FloatType or HalfType
simple reproducer:
```
from llvmlite import ir

fp = ir.DoubleType()
#fp = ir.FloatType()
#fp = ir.HalfType()
fnty = ir.FunctionType(fp, ())

module = ir.Module(name=__file__)
func = ir.Function(module, fnty, name="fpadd")
block = func.append_basic_block(name="entry")
builder = ir.IRBuilder(block)
x = fp("Inf")
builder.ret(x)

print(module)
```

The above only works with DoubleType. Both FloatType and HalfType produce the following error:
```
    return struct.unpack('f', struct.pack('f', value))[0]
struct.error: required argument is not a float
```

Using `float("Inf")` instead of just `"Inf"` works around the problem:
```
target datalayout = ""

define float @"fpadd"() 
{
entry:
  ret float 0x7ff0000000000000
}
```

vsaraikin:
Fixes #833.

`FloatType` and `HalfType` fail to format a special value passed as a string, while `DoubleType` accepts it:

```python
from llvmlite import ir
ir.DoubleType().format_constant("Inf")  # 0x7ff0000000000000
ir.FloatType().format_constant("Inf")   # struct.error: required argument is not a float
ir.HalfType().format_constant("Inf")    # struct.error: required argument is not a float
```

`DoubleType.format_constant` goes through `_format_float_as_hex`, which does `struct.pack(fmt, float(value))`. The `FloatType`/`HalfType` paths go through `_as_float`/`_as_half`, which call `struct.pack('f'/'e', value)` without the `float()` conversion, so a string argument raises `struct.error`.

The fix converts the argument to float in `_as_float` and `_as_half`, matching what `_format_float_as_hex` already does. After it, all three types produce the same representation for `"Inf"`, `"-Inf"` and `"nan"`.

Added a `test_reals_special_values` case in `TestConstant` covering Inf/-Inf/NaN as strings for half, float and double (this replaces the old `# XXX Test NaNs and infs` note in `test_reals`).

### DIFF
--- a/llvmlite/ir/types.py
+++ b/llvmlite/ir/types.py
@@ -336,7 +336,7 @@ def _as_float(value):
     """
     Truncate to single-precision float.
     """
-    return struct.unpack('f', struct.pack('f', value))[0]
+    return struct.unpack('f', struct.pack('f', float(value)))[0]
 
 
 def _as_half(value):
@@ -344,7 +344,7 @@ def _as_half(value):
     Truncate to half-precision float.
     """
     try:
-        return struct.unpack('e', struct.pack('e', value))[0]
+        return struct.unpack('e', struct.pack('e', float(value)))[0]
     except struct.error:
         # 'e' only added in Python 3.6+
         return _as_float(value)
--- a/llvmlite/tests/test_ir.py
+++ b/llvmlite/tests/test_ir.py
@@ -2807,7 +2807,6 @@ class TestConstant(TestBase):
         self.assertEqual(str(c), 'i1 0')
 
     def test_reals(self):
-        # XXX Test NaNs and infs
         c = ir.Constant(flt, 1.5)
         self.assertEqual(str(c), 'float 0x3ff8000000000000')
         c = ir.Constant(flt, -1.5)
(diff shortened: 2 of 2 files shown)


swap357 (approved):
LGTM 
I reproduced #833 against `main` and verified the change at 4a300c75
Converting with `float()` makes `HalfType` and `FloatType` consistent with the existing `DoubleType` now for special inputs -
```
>>> str(ir.Constant(ir.HalfType(), 'nan'))
'half 0x7ff8000000000000'
>>> str(ir.Constant(ir.HalfType(), 'inf'))
'half 0x7ff0000000000000'
>>> str(ir.Constant(ir.HalfType(), '-inf'))
'half 0xfff0000000000000'
>>> str(ir.Constant(ir.FloatType(), 'nan'))
'float 0x7ff8000000000000'
>>> str(ir.Constant(ir.FloatType(), 'inf'))
'float 0x7ff0000000000000'
>>> str(ir.Constant(ir.FloatType(), '-inf'))
'float 0xfff0000000000000'
>>> str(ir.Constant(ir.DoubleType(), 'Inf'))
'double 0x7ff0000000000000'
```

Note: while this behavior was already present for `DoubleType`, this enables using all strings that can be cast to `float` and `ValueError`s otherwise-
```
>>> str(ir.Constant(ir.FloatType(), '1_22'))
'float 0x405e800000000000'
>>> str(ir.Constant(ir.FloatType(), '1_22a'))
...
ValueError: could not convert string to float: '1_22a'
```

Thank you for your contribution, @vsaraikin. much appreciated!

PR #1466, opened 2026-08-04
CI: expand `win-arm64` py-version matrix to cover python 3.11-3.14

swap357:
as title. 
This PR adds CI support to build `win-arm64` llvmlite wheels and conda packages with python versions `3.11-3.14`, previously only had `3.14`.

### DIFF
--- a/.github/workflows/llvmlite_win-arm64_conda_builder.yml
+++ b/.github/workflows/llvmlite_win-arm64_conda_builder.yml
@@ -39,7 +39,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.14"]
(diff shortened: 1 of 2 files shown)


esc (approved):
Looks good, assuming that 3.10 will never be added, because it is EOL soon and that free-threading variants for `WoA` will be added as the support for free-threaded interpreters matures over time?

swap357:
> Looks good, assuming that 3.10 will never be added, because it is EOL soon and that free-threading variants for `WoA` will be added as the support for free-threaded interpreters matures over time?

For py3.10, there's also no numpy support. And yes, free-threading builds will be added when we have reliable source to get python and numpy `*t` build from, currently there's none.

PR #1473, opened 2026-08-19
CI: test win-arm64 on windows-11-vs2026-arm preview

swap357:
The intention of this PR is to get `win-arm64` workflows ready for upcoming change of `windows-11-arm` image to Visual Studio 2026 ([link](https://github.blog/changelog/2026-08-20-windows-11-arm64-vs2026-image-generally-available/)) . To prepare for it, this PR updates image to `windows-11-vs2026-arm` preview image.

As I found, that image needs cmake >= 4.2 for the Visual Studio 18 2026 generator.
bootstrap still has 3.31.2; defaults (pkgs/main) has 4.3.4.
Jobs use `-c defaults` instead of `-c main`. For cmake, 
- for conda: `cmake >=4.2` in now specified in the recipe ( constraint >=4.2 for win and arm64), no strict channel priority
- for wheel: ~`conda config --add pinned_packages 'defaults::cmake'` to pin cmake from defaults when using `conda create` for build environment.~
EDIT: same `cmake >=4.2` constraint on conda build env create command

### DIFF
--- a/.github/workflows/llvmlite_win-arm64_wheel_builder.yml
+++ b/.github/workflows/llvmlite_win-arm64_wheel_builder.yml
@@ -89,6 +89,7 @@ jobs:
           else
             LLVMDEV_SRC="-c ${LLVMDEV_CHANNEL}"
           fi
+          "${CONDA_EXE}" config --add pinned_packages 'defaults::cmake'
           "${CONDA_EXE}" create --prefix "${CONDA_PREFIX}" --yes \
             ${LLVMDEV_SRC} -c "${WIN_ARM64_CONDA_CHANNEL}" ${EXTRA_CHANNEL} \
             -c m2-winarm64 -c defaults \
--- a/.github/workflows/llvmlite_win-arm64_wheel_builder.yml
+++ b/.github/workflows/llvmlite_win-arm64_wheel_builder.yml
@@ -29,10 +29,14 @@ permissions:
 env:
   FALLBACK_LLVMDEV_VERSION: "22"
   ARTIFACT_RETENTION_DAYS: 7
-  LLVMDEV_CHANNEL: numba/label/llvm_wheel
+  LLVMDEV_CHANNEL: https://conda.anaconda.org/numba/label/llvm_wheel
   # win-arm64 has no Miniconda installer, so the base environment is
   # bootstrapped from conda-standalone using the win-arm64 native channel.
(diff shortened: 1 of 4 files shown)


MugundanMCW:
Hi @swap357,

I could see that the Windows ARM64 jobs are failing due to the recent shift from the VS2022 toolset to the VS2026 toolset. According to the CMake documentation, support for detecting the VS2026 toolchain was added only in[ CMake 4.2](https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2018%202026.html). Since the llvmlite project currently uses the CMake 3.31.x conda package, the VS2026 generator is not recognized, causing the Windows ARM64 builds to fail.

swap357:
with [3e0006a](https://github.com/numba/llvmlite/pull/1473/commits/3e0006aafb6943c6c0288088f942b2992252cba0), I've changed the channels to be explicit urls to improve clarity. names like `-c defaults` are a channel group (`pkgs/main` + `pkgs/r` ..), and `-c main` is a different URL (`https://conda.anaconda.org/main/`) so the short names are easy to mix up.

esc:
Ref: https://github.com/conda-incubator/setup-conda-standalone/issues/34

.github/workflows/llvmlite_win-arm64_wheel_builder.yml
```diff
@@ -89,9 +93,10 @@ jobs:
           else
             LLVMDEV_SRC="-c ${LLVMDEV_CHANNEL}"
           fi
+          "${CONDA_EXE}" config --add pinned_packages 'pkgs/main::cmake'
```
esc:
```suggestion
          "${CONDA_EXE}" config --add pinned_packages 'cmake>=4.2'
```

esc (changes requested):
Suggest to depend on the `cmake` version rather than the `cmake` from a specific channel?

.github/workflows/llvmlite_win-arm64_wheel_builder.yml
swap357: updated getting `cmake` using version constraint on `conda create` command itself with https://github.com/numba/llvmlite/pull/1473/changes/aae7eb9f09e219ad7effe1b5d22c292cb9422831

esc: yeah, that is even better!

esc (approved):

PR #1475, opened 2026-08-20
Fixup remaining references to 3.15

esc:
Followup from #1463


swap357 (approved):
Thanks for catching these. I've checked need for references elsewhere and seems like these are only ones that need updating for now. 

Note: there's remaining reference for cp314 and cp315 on here -
https://github.com/numba/llvmlite/blob/e552c0f43f6db8c505c895b1e3c0484d57a93aac/buildscripts/manylinux/README.md?plain=1#L34-L37
but those can be appropriately handled on llvmlite manylinux image upgrade PR.

swap357:
The CI fails are not relevant to changes here, mostly conda index update related.

Issue #1034, opened 2024-02-27
Add `numba` and `llvmlite` `musl` wheels for `alpine` support


mpierrau:
## Feature request

Hi, I was wondering if it is possible to add [MUSL wheels](https://musl.libc.org/) for `numba` and `llvmlite` since lots of people use CICD images based on [Alpine](https://alpinelinux.org/). This distribution uses different a implementation of libc: [musl](https://musl.libc.org/) (see [Alpine Linux has switched to musl libc | Alpine Linux ](https://alpinelinux.org/posts/Alpine-Linux-has-switched-to-musl-libc.html)), and is incompatible with `manylinux` wheels.

I know that for example the `shap` library is [waiting for `llvmlite` and `numba` to release MUSL wheels](https://github.com/shap/shap/pull/3087).

Thanks!

esc:
@mpierrau quick question about this: where are they distributed, can they be uploaded to PyPi too? How many artifacts are we talking about here?

esc:
@mpierrau we just discussed this during our issue triage and it is unlikely that this will happen anytime soon due to a maintainer resource shortage. Would suggest to leave this open anyway, in case that resources become available over the coming months or years.

mpierrau:
I have no experience from distributing Python packages, but from what I can gather, the `musl` wheels are also uploaded to `PyPi`, and it's one wheel artifact per architecture you wish to support. The number of architectures supported for `musl` are in general the same as for `manylinux` (from looking in my lockfiles for other packages which do distribute `musl` wheels to `PyPi`).

It looks like [`cibuildwheel`](https://cibuildwheel.readthedocs.io/en/stable/) supports building `musl` wheels for `CPython` 3.6 - 3.12, for the following architectures `x86_64`, `aarch64`, `i686`, `s390x`, `ppc64le`.

esc:
@mpierrau thank you for digging into this, much appreciated.

Thoughts:

a) We have space issues on PyPi with llvmlite. Our wheels contain LLVM itself and we regularly run out of space on PyPi because of this. Adding another 4 (or 8?) wheels per release may increase the size requirement so we'll probably need additional checks before uploading, if there is enough space left. This isn't something that many or even most Python packages face so tooling here -- for example to check for space -- is very rudimentary and sometimes not reliable. We usually only notice that we run out of space when a release fails to upload and then need to request more space on PyPi which can sometimes take several days to process from the PyPi side.

[...]

PR #1480, opened 2026-09-02
fix: <command-line>: error: token '=' is not valid in preprocessor expressions

strophy:
While working on https://github.com/numba/llvmlite/issues/1034 to add llvmlite to Alpine Linux, our CI threw hundreds of errors like this on 32-bit machines only:
```
<command-line>: error: token '=' is not valid in preprocessor expressions
<command-line>: error: token '=' is not valid in preprocessor expressions
<command-line>: error: token '=' is not valid in preprocessor expressions
<command-line>: error: token '=' is not valid in preprocessor expressions
<command-line>: error: token '=' is not valid in preprocessor expressions
<command-line>: error: token '=' is not valid in preprocessor expressions
<command-line>: error: token '=' is not valid in preprocessor expressions
<command-line>: error: token '=' is not valid in preprocessor expressions
```
The issue is that `LLVM_DEFINITIONS` is a space separated string and not a CMake list, so passing it straight to `add_definitions()` defines a single macro whose value is the remainder of the string. A definition with an `=` in it like `-D_FILE_OFFSET_BITS=64` (added by LLVM on 32 bit targets) ends up inside the value of the first macro, which then fails to parse in any #if that tests it.  It looks like this on armv7, notice the incorrect quoting:
```
CXX_DEFINES = -DHAVE_LLVMLITE_SHARED -DLLVMLITE_LLVM_ASSERTIONS_STATE=2 -D_GLIBCXX_USE_CXX11_ABI="1 -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS" -Dllvmlite_EXPORTS
```

Fixed by splitting the arguments into a Cmake list first, then adding them as definitions.

### DIFF
--- a/ffi/CMakeLists.txt
+++ b/ffi/CMakeLists.txt
@@ -52,7 +52,8 @@ else()
 endif()
 
 include_directories(${LLVM_INCLUDE_DIRS})
-add_definitions(${LLVM_DEFINITIONS})
+separate_arguments(LLVM_DEFINITIONS_LIST NATIVE_COMMAND "${LLVM_DEFINITIONS}")
+add_definitions(${LLVM_DEFINITIONS_LIST})
 
 # Check for presence of the SVML patch in the LLVM build
 set(CMAKE_REQUIRED_INCLUDES ${LLVM_INCLUDE_DIRS})


esc:
@strophy yeah, thanks for this, I saw this issue when looking at the alpine CI:

https://gitlab.alpinelinux.org/strophy/aports/-/pipelines/468834

Do you have a way to test this in your gitlab instance?

strophy:
Sure, I've added the change in GitLab CI as a patch. The build still errors in a different way, but I'll get to that in coming days. For the fix to Cmake mangling I'm proposing in this PR, here are the logs from armv7 with the patch applied as an example:

Before: https://gitlab.alpinelinux.org/strophy/aports/-/jobs/2499428
After: https://gitlab.alpinelinux.org/strophy/aports/-/jobs/2500460

I think the parsing of `LLVM_DEFINITIONS` only worked by luck to date on 64-bit machines, the parsing has always been incorrect and it was only exposed by the 32-bit builds we attempted here.

strophy:
The value for `-D_GLIBCXX_USE_CXX11_ABI` on x86_64 currently evaluates to `1 -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS` between the quotes. This is parsed as `1-0-0-0`, accidentally resulting in `1` as the value passed to `-D_GLIBCXX_USE_CXX11_ABI`. So merging this will have the side-effect of setting `-D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS` which were previously not being set correctly due to the mangling. Not sure how this will affect your builds, but it seems to still pass for x86_64 on Alpine CI.

esc:
@strophy thank you for submitting this. I will take this to the developer meeting next and discuss with the other maintainers on how to proceed. Since this has side effects on our current builds we need to asses the risk of impact this may have on our downstream dependents.

stuartarchibald:
When I saw this I though "oh yes, that is a bug, because I've also hit it" but I think the work I was doing that carried a fix for it (re-enabling the `remarks` interface) got stuck in something I don't quite understand yet (could be a bug), and so it was never resolved. I can't see an issue with those macros being set as a result of fixing this. Further, `llvmlite` needs to correctly adhere to what is in `LLVMConfig.cmake`, if `llvmlite` doesn't work because of that, that issue needs to be addressed opposed to relying on an incorrect parse.

stuartarchibald (approved):
This was discussed at the maintainer meeting today. The change makes sense and I looked in `LLVMConfig.cmake` where the `LLVM_DEFINITIONS` are found and indeed they would need splitting as it is a quoted string. As noted in my previous comment, if doing this breaks something, we need to fix the break as the `LLVM_DEFINITIONS` need to be respected. A spot check of a couple of `LLVMConfig.cmake` suggests that the extra defines are those listed above (`-D__STDC*`) which are unlikely to cause problems. Thanks again for the patch.

PR #1481, opened 2026-09-03
chore: remove TestDependencies due to failure on musl

strophy:
Another potential bug found while working on https://github.com/numba/llvmlite/issues/1034 `llvmlite.tests.test_binding.TestDependencies.test_linux` assumes glibc when checking for `libpthread`, but `libpthread` doesn't exist on musl because pthreads live inside libc itself. It fails at the sanity check `if 'libc' not in deps or 'libpthread' not in deps:` with:
```
FAIL: test_linux (llvmlite.tests.test_binding.TestDependencies.test_linux)
  File "/home/builder/package/community/py3-llvmlite/src/llvmlite-0.49.0/llvmlite/tests/test_binding.py", line 755, in test_linux
    self.fail("failed parsing dependencies? got %r" % (deps,))
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: failed parsing dependencies? got {'libgcc_s', 'libc.musl-x86_64', 'libzstd', 'libstdc++', 'libz'}
```

This PR patches the test logic so it is aware of this architecture in musl libc.

### DIFF
--- a/llvmlite/tests/test_binding.py
+++ b/llvmlite/tests/test_binding.py
@@ -742,6 +742,7 @@ class TestDependencies(BaseTest):
         # Parse library dependencies
         lib_pat = re.compile(r'^([+-_a-zA-Z0-9]+)\.so(?:\.\d+){0,3}$')
         deps = set()
+        musl = False
         for line in out.decode().splitlines():
             parts = line.split()
             if parts and parts[0] == 'NEEDED':
@@ -749,9 +750,15 @@ class TestDependencies(BaseTest):
                 m = lib_pat.match(dep)
                 if len(parts) != 2 or not m:
                     self.fail("invalid NEEDED line: %r" % (line,))
-                deps.add(m.group(1))
+                name = m.group(1)
+                if name.startswith('libc.musl-'):
+                    musl = True
+                    name = 'libc'
+                deps.add(name)
         # Sanity check that our dependencies were parsed ok
-        if 'libc' not in deps or 'libpthread' not in deps:
+        # musl implements the pthread API inside libc itself
+        required = {'libc'} if musl else {'libc', 'libpthread'}
+        if not required.issubset(deps):
             self.fail("failed parsing dependencies? got %r" % (deps,))
         # Ensure all dependencies are expected
         allowed = set(['librt', 'libdl', 'libpthread', 'libz', 'libm',


stuartarchibald (commented):
Thanks for the patch @strophy. Looking at the test, I'm not sure it's actually needed any more given we have much more involved checking of build artifacts now. @swap357 what do you think, this test seems like it is a duplicate/subset of existing tests? I suppose the minor advantage it has is that it doesn't depend on `lief`.

swap357:
> Thanks for the patch @strophy. Looking at the test, I'm not sure it's actually needed any more given we have much more involved checking of build artifacts now. @swap357 what do you think, this test seems like it is a duplicate/subset of existing tests? I suppose the minor advantage it has is that it doesn't depend on `lief`.

Yeah, better to just drop this test, it's redundant with newer `TestBuild`. 

>  I suppose the minor advantage it has is that it doesn't depend on lief.

These tests are gated for packaging CI specifically, and `lief` should be available to test there(at least for one of the Python version we support). It might be that lief won't have support for newer Python version, but if CI can run check with one of previous Python version wheel, it should be good. (as linkage does not change across python versions.)

swap357:
@strophy can we make this a cleanup PR and remove `TestDependencies` altogether ? Does that resolve llvmlite testing on musl ?

strophy:
Sure, that will fix things on my end 👍

esc:
@strophy do you want to take over the baton here and fix this one up?

strophy:
> @strophy do you want to take over the baton here and fix this one up?

OK, done!

swap357:
> > @strophy do you want to take over the baton here and fix this one up?
> 
> OK, done!

Thank you, @strophy. 
I assume there was no other code change required to get llvmlite build and test on `alpine` apart from this?

strophy:
The only other change was https://github.com/numba/llvmlite/pull/1480 which was already merged. You can track the process of getting it added to Alpine at https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/107692 and I'll update that MR if you release a new version before someone gets around to reviewing and merging it.

Thanks as well for your prompt help on these issues!

esc (approved):

esc:
@strophy thank you for the patch.

PR #1487, opened 2026-09-11
Add support for LLVM range parameter attributes

alcides:
I am using llvmlite in http://github.com/alcides/aeon to lower my programming language to llvm. My language supports refinement types, which I want to keep (as much as llvm supports), so I need this feature in llvmlite to avoid ugly workarounds.

Full transparency: the PR was generated by Codex, but I have reviewed it.

-------

LLVM supports the `range` parameter attribute, but llvmlite's `ArgumentAttributes` allowlist currently rejects it. This patch accepts syntactically formed `range(...)` attributes and emits them for function parameters and call-site arguments.

The accepted form follows LLVM IR's parameter attribute syntax. Malformed values remain rejected, preserving the existing validation behavior for unknown attributes.

Tests: `llvmlite/tests/test_ir.py` updated. The full test suite could not run locally because the shallow checkout environment lacks an LLVM development installation required to build llvmlite's native extension.

### DIFF
--- a/llvmlite/ir/values.py
+++ b/llvmlite/ir/values.py
@@ -1052,6 +1052,13 @@ class ArgumentAttributes(AttributeSet):
         'zeroext': False,
     })
 
+    _range_re = re.compile(r'range\([^()]+\)\Z')
+
+    def add(self, name):
+        if name not in self._known and not self._range_re.fullmatch(name):
+            raise ValueError('unknown attr {!r} for {}'.format(name, self))
+        return super(AttributeSet, self).add(name)
+
     def __init__(self, args=()):
         self._align = 0
         self._dereferenceable = 0
--- a/llvmlite/tests/test_ir.py
+++ b/llvmlite/tests/test_ir.py
@@ -162,6 +162,7 @@ class TestFunction(TestBase):
         # Now with parameter attributes
         func = self.function()
         func.args[0].add_attribute("zeroext")
+        func.args[0].add_attribute("range(i32 0, 10)")
         func.args[1].attributes.dereferenceable = 5
         func.args[1].attributes.dereferenceable_or_null = 10
         func.args[3].attributes.align = 4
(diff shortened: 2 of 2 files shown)


llvmlite/ir/values.py
```diff
@@ -1052,6 +1052,13 @@ class ArgumentAttributes(AttributeSet):
         'zeroext': False,
     })
 
+    _range_re = re.compile(r'range\([^()]+\)\Z')
+
+    def add(self, name):
+        if name not in self._known and not self._range_re.fullmatch(name):
+            raise ValueError('unknown attr {!r} for {}'.format(name, self))
+        return super(AttributeSet, self).add(name)
```
swap357: `range(<ty> <a>, <b>)` does not fit `_known` the same way `zeroext` or `captures(none)` do here. Those are exact strings passed to `add_attribute`. `range` also needs the bounds. `align` and `dereferenceable` already work like that. 
Can you use those as reference to add `range` ?

They can also be set on attributes, checked in the setter and printed from `_to_list`:
```
In [40]: from llvmlite import ir

In [41]: m = ir.Module()

In [42]: i32 = ir.IntType(32)

In [43]: f = ir.Function(m, ir.FunctionType(i32, [i32, i32.as_pointer()]), "f")
    ...: 

In [44]: f.args[0].add_attribute("zeroext")

In [45]: f.args[0].add_attribute("noundef")

In [46]: f.args[0].attributes.range = (0, 10)

In [47]: f.args[1].add_attribute("nonnull")

In [48]: f.args[1].attributes.align = 4

In [49]: f.return_value.add_attribute("noalias")

In [50]: f
Out[50]: <ir.Function 'f' of type 'i32 (i32, i32*)*'>

In [51]: str(f)
Out[51]: 'declare noalias i32 @"f"(i32 noundef zeroext %".1", i32* nonnull align 4 %".2")\n'

```
With current implementation, `range` cannot be set like this as apparent from example above.

llvmlite/tests/test_ir.py
```diff
@@ -162,6 +162,7 @@ def test_function_attributes(self):
         # Now with parameter attributes
         func = self.function()
         func.args[0].add_attribute("zeroext")
+        func.args[0].add_attribute("range(i32 0, 10)")
```
swap357: test with `func.args[0].attributes.range = (0, 10)` if above comment is addressed

swap357 (changes requested):
I've reviewed this.  Since `range(<ty> <a>, <b>)` is not an exact `_known` string like `zeroext` / `captures(none)`, it does not fit `add_attribute + _known`. The bounds are extra data and should follow same implementation as `align` and `dereferenceable`. The tests can be updated similarly. For tests, its also important to remember `range` is not valid on pointers.

Review Assisted by: Kilo AI (Grok 4.6)

